PR #337, opened 2015-02-04
removed additional set cmake_minimum, each one resets any cmake policies...

makeclean:
Removes all sub CMAKE_MINIMUMS from the Cmakefiles as each one resets which policies are in effect from the master cmakefile


makeclean:
@julry this is ready to go, I have one more coming, which depends on this one.

PR #349, opened 2015-04-02
renamed the amalgamated pyne library name to avoid name clash

makeclean:
When originally added the the local amalgamated pyne library name accidentally clashes with the libpyne.so library from PyNE itself, thus users that have both installed often get cryptic shared library problems. This fixes this. Tests pass on my machine.


Pshriwise:
Yes!!

gonuke:
While this may solve the problem, I wonder if it is a standard pattern?  I think other tools that build their own versions of third-party dependencies install them in a specific location and then are hardwired to look in that location. That said, I know it is a source of fragility for those tools, as well.

Visit comes to mind as one of those....

PR #350, opened 2015-04-03
Fix exception

makeclean:
Originally the UWUW class determined if there were materials and tallies by trying to read them using the full from_hdf5 methods, these methods raise an exception if they fail, thus the function was wrapped in a try/except block. This had the unintended consequence that when either materials or tallies were not present the exception was tried and caught. However, this still counted as an exception being raised, subsquently MOAB would go on to pick up the error and report that some operation had failed when it acutally hadn't. This bad design decision has been rectified and now materials and tallies have their existing queried for without try/except, and their datalength determined, thus the correct amount is always loaded in. Tests pass on my machine.


julry:
Tests pass after being added to the CI build.

julry:
The  CI build has been merged, the PR build should run fine...

PR #362, opened 2015-08-20
Fixed re-entrant mesh bug in TrackLengthMeshTally.cpp

ljacobson64:
There was a bug in TrackLengthMeshTally.cpp which concerned particle behavior in re-entrant meshes. Occasionally when a particle entered, left, and then reentered a mesh, the track length outside the mesh was counted as being inside the mesh, which led to incorrect results. This PR fixed the bug and added a new test in test_TrackLengthMeshTally.cpp.


makeclean:
Thanks @ljacobson64, this is excellent, extra tests are always good, we also really need to break your geany habit.

PR #363, opened 2015-08-29
bug fixes

elliottbiondo:
This fixes two bugs in ```uwuw_preproc```. The first bug was that graveyard tags were not being properly read in, so the script would never successfully find the graveyard. The second issue is that materials names where not being properly handled for tags in the form ```mat:X/rho:Y```.


elliottbiondo:
@makeclean you should still consider accepting this, just so the repo version is somewhat usable.

PR #364, opened 2015-08-30
Script generating material libraries from MCNP input files.

elliottbiondo:
This PR adds a Python script to read MCNP input files and create UWUW material library files, with a naming scheme compatible with the MCNP2CAD -U flag. A unit test was added for this script. The script was confirmed to work with ```uwuw_preproc```, contingent on the bug fixes [here](https://github.com/svalinn/DAGMC/pull/363).


makeclean:
Awesome, thanks!

PR #367, opened 2015-09-10
fixed reentrant mesh problem in TrackLengthMeshTally without introducing whitespace changes

ljacobson64:
I think this is what we want.


ljacobson64:
whoops

ljacobson64:
whoops again

(I closed it because I mistakenly thought it didn't build correctly, but what actually happened is I didn't have my environment variables set properly. Ignore my close and reopen.)

makeclean:
Thanks @ljacobson64

PR #369, opened 2015-09-11
fixes the bug found when using data files not in the current working …

### LINKED ISSUE #368: gcad path handled incorrectly
When running DAG-MCNP, the `gcad` argument should specify the path to the CAD geometry file. It looks like this path is being appended to `pwd`, hence the output below. Notice that two paths beginning with `/home/` are appended:

```
The file /home/biondo/FNG_ITER/p_trans_analog/t17_test//home/biondo/benchmarks/fng_iter/adapted_from_sinbad/photon/mcnp_g_adapted_with_detectors_zip.h5m does not exist or is read protected```

### PR DESCRIPTION
**Pull request by makeclean**
This PR fixes the bug @elliottbiondo found when running files not found in the current working directory. The bug was that the cwd was always appended to the file irrespective of the path. Now the if no ".." are found and "/" is found then we assume that its an absolute path. Test has been added to address this case. This will close #368

### DIFF
--- a/uwuw/tests/uwuw_test.cpp
+++ b/uwuw/tests/uwuw_test.cpp
@@ -54,6 +54,24 @@ TEST_F(UWUWTest,filepath2) {
 }
 
 /*
+ * Test of absolute path
+ */
+TEST_F(UWUWTest,filepath3) {
+  // get the full current path
+  char current_path[FILENAME_MAX];
+  // get the cwd
+  getcwd(current_path,sizeof(current_path));
+  std::string filepath(current_path);
+  // full path to file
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by makeclean** [comment]
@elliottbiondo could you confirm that this fixes your problem?

**Comment by elliottbiondo** [comment]
Sure, one sec.

**Comment by elliottbiondo** [comment]
Nope:
```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_S_construct NULL not valid
```

elliottbiondo:
Latest update works.

makeclean:
All tests pass and @elliottbiondo's case fixed.

PR #373, opened 2015-10-15
when static linking need to set vars for findhdf5.cmake

makeclean:
When linking DAGMC in static mode we need to include the appropriate static hdf5 libraries, this is done by setting a cmake flag before find_hdf5 is called.


makeclean:
All tests pass. Merging.

PR #374, opened 2015-10-19
added the new version of moab to the test list

makeclean:
Added MOAB 4.9.0 to the test list as a named release


makeclean:
All good.

PR #376, opened 2015-10-23
added the ability to produce an executable that plots, and allows the…

makeclean:
… setting of the MCNP datapath environemnt.

This PR enables plotting at the request of Tim Bohm, additionally the cmake options has been added to set the MCNP datapath variable by CMAKE argument, otherwise if unspecified will look in the DATAPATH enviroment variable.


makeclean:
This isn't quite right, needs some additional work. Will pull out the plot stuff but will keep in the datapath fixes.

makeclean:
Actually take it back, it works perfectly.

PR #377, opened 2015-10-27
Add mcnp improvements

makeclean:
Added missing quotes around string and redefined the fortran flags.


makeclean:
All passed. Merging

PR #379, opened 2015-10-29
fixed the Geant4 error, came from the Fluka neutron group bin bounds …

makeclean:
…having a duplicate entry


makeclean:
This all passes, in preparation for China visit.

PR #381, opened 2015-10-30
removed fdefault compiler flags

ljacobson64:
The flags were causing errors when running DAGMC; tested on multiple inputs


ljacobson64:
Never mind Andy said to close this

makeclean:
@ljacobson64 thanks for this, I will investigate why this is needed.

makeclean:
thanks @ljacobson64

PR #390, opened 2016-01-14
Fluka lsense fix rebase

makeclean:
This PR addresses the issues that FluDAG had with charged particle transport, these additional changes also have better error handling and improved structure. This is a placeholder, and more issues commits will likely arrive


makeclean:
This has been tested internally and performs as expected.

PR #394, opened 2016-05-16
make sure all cells on lcad card have an imp:<particle> card if any of them do

ljacobson64:
This fixes the issue I brought up in the meeting on 5/10. In problems with a mode other than mode n, the LCAD file would not contain all the importances for the other particles, and MCNP would fail. Now, it guarantees that if any cell has an importance for n, p, or e, all cells will have an importance for those particles.

### DIFF
--- a/mcnp5/mcnp_funcs.cpp
+++ b/mcnp5/mcnp_funcs.cpp
@@ -479,13 +479,12 @@ void write_lcad_old(std::ofstream &lcadfile)
   int num_surfs = DAG->num_entities( 2 );
 
   int cmat = 0;
-  double crho;
+  double crho = 0;
+  double cimp_n = 0, cimp_p = 0, cimp_e = 0;
 
   // Detect which importances are used so all cells, including implicit
   // complement and graveyard, have these importances
-  bool imp_n_needed = false;
-  bool imp_p_needed = false;
-  bool imp_e_needed = false;
+  bool imp_n_needed = false, imp_p_needed = false, imp_e_needed = false;
   for( int i = 1; i <= num_cells; ++i ) {
     moab::EntityHandle vol = DAG->entity_by_index( 3, i );
     if( DAG->has_prop( vol, "imp.n" )) imp_n_needed = true;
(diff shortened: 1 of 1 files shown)


ljacobson64:
Don't merge this yet... I found more problems

mcnp5/mcnp_funcs.cpp
```diff
@@ -479,45 +479,68 @@ void write_lcad_old(std::ofstream &lcadfile)
...
     int cellid = DAG->id_by_index( 3, i );
-    // set default importances for p and e to negative, indicating no card should be printed.
-    double imp_n = 1, imp_p = -1, imp_e = -1;
+    // set default importances to negative, indicating no card should be printed.
+    double imp_n = -1, imp_p = -1, imp_e = -1;
 
     if( DAG->has_prop( vol, "imp.n" )) {
       get_real_prop( vol, cellid, "imp.n", imp_n );
+    } else if( imp_n_needed ) {
+      imp_n = 0;
```
gonuke: why default to 0 - seems wrong...

ljacobson64: I agree, I will change that to 1.

What do you think should happen if the .sat file doesn't contain any info on importances at all? Assume it's a neutron problem and give them all imp:n=1?

makeclean: I think that was the default. If there is no graveyard I think there should be a failure if there already isn't.

mcnp5/mcnp_funcs.cpp
```diff
@@ -479,45 +479,68 @@ void write_lcad_old(std::ofstream &lcadfile)
...
     if( graveyard ) {
-      lcadfile << " 0 imp:n=0";
+      lcadfile << " 0";
+      if( imp_n_needed ) lcadfile << " imp:n=0";
+      if( imp_p_needed ) lcadfile << " imp:p=0";
+      if( imp_e_needed ) lcadfile << " imp:e=0";
       if( DAG->has_prop(vol, "comp") ) {
         // material for the implicit complement has been specified.
         get_int_prop( vol, cellid, "mat", cmat );
         get_real_prop( vol, cellid, "rho", crho );
```
ljacobson64: I'm not sure what's going on here... why would the graveyard ever also be the implicit complement?

makeclean: So imagine a model, like an ITER model where the entirity of the 40 degree sector is fully defined by CAD solids, the implicit compliment could then serve as a graveyard.

ljacobson64: It seems like the current logic is that if the graveyard is also the implicit complement, the variables cmat, crho, and cimp are set. But I don't see how they would ever be used, because the implicit complement code is inside of an "else if" statement.

PR #400, opened 2016-05-31
Refactored the build process

ljacobson64:
* Refactored all the cmake
* Altered the cmake for MCNP to search for the Fortran files instead of hard-coding them in
  * throws error if it can't find the source code or if it looks like the patch hasn't been applied
* Regenerated mcnp5 v1.40, v1.51, v1.60 patches for consistency
* New patch application process: cd into mcnp5 directory, use patch -p0 < patch/\[patch_file]
* Fixed FindGeant4 to do a better job of actually finding Geant4
* We can now successfully build static libraries and executables
* We can now build an MCNP executable that can both run transport and use the native MCNP plotter
* Moved the DAG-MCNP C++ to a new folder in anticipation of DAG-MCNP6
  * Both 5 and 6 will use the same C++ code


ljacobson64:
Green checkmark!

ljacobson64:
Closed, use PR #401 instead

PR #401, opened 2016-06-02
Refactored the build process

ljacobson64:
This PR is the same as PR #400 with all the commits squashed into one.


makeclean:
Could you please 

[x] Move mcnp5 into mcnp, such that when mcnp6 comes along it will site nicely in mcnp
[x] Please move tally/tools back to where it was, but have it build and install into main tools dir
[x] Remove the mcnp sample folder
[x] Remove the fluka input directory
[x] Remove the geom_material.h5m and mat.inp files
[x] share folder should be deleted 

Then we should be good to go. Thanks :)

ljacobson64:
tally/tools/boundary_correction doesn't build. This is the error it gives:

```
/home/lucas/mcnp6_build/src/tools/boundary_correction/boundary.cpp: In function ‘int main(int, char**)’:
/home/lucas/mcnp6_build/src/tools/boundary_correction/boundary.cpp:351:46: error: no matching function for call to ‘moab::Skinner::find_skin(moab::Range&, int, moab::Range&)’
   skinner.find_skin(mesh_cells, 0, skin_nodes);
```

makeclean:
Looks like a signature change, https://bitbucket.org/fathomteam/moab/src/6e8617d49d8abf387e4fc0297e706e3deb6a44c6/src/Skinner.cpp?at=master&fileviewer=file-view-default#Skinner.cpp-240

ljacobson64:
I think fixing that could be a separate pull request

makeclean:
Ok, excellent, Thanks @ljacobson64

PR #403, opened 2016-06-15
Make sure all cells on LCAD card have an IMP:<particle> card if any of them do

ljacobson64:
Trying this again

### DIFF
--- a/mcnp/mcnp_funcs.cpp
+++ b/mcnp/mcnp_funcs.cpp
@@ -476,16 +476,38 @@ void write_lcad_old(std::ofstream &lcadfile)
   int num_surfs = DAG->num_entities( 2 );
 
   int cmat = 0;
-  double crho, cimp = 1.0;
+  double crho = 0;
+  double cimp_n = 0, cimp_p = 0, cimp_e = 0;
+
+  // Detect which importances are used so all cells, including implicit
+  // complement and graveyard, have these importances
+  bool imp_n_needed = false, imp_p_needed = false, imp_e_needed = false;
+  for( int i = 1; i <= num_cells; ++i ) {
+    moab::EntityHandle vol = DAG->entity_by_index( 3, i );
+    if( DAG->has_prop( vol, "imp.n" )) {
+      imp_n_needed = true;
+      cimp_n = 1;
+    }
+    if( DAG->has_prop( vol, "imp.p" )) {
+      imp_p_needed = true;
+      cimp_p = 1;
+    }
+    if( DAG->has_prop( vol, "imp.e" )) {
+      imp_e_needed = true;
+      cimp_e = 1;
+    }
+  }
 
   // write the cell cards
   for( int i = 1; i <= num_cells; ++i ) {
 
     moab::EntityHandle vol = DAG->entity_by_index( 3, i );
     int cellid = DAG->id_by_index( 3, i );
-    // set default importances for p and e to negative, indicating no card should be printed.
-    double imp_n = 1, imp_p = -1, imp_e = -1;
 
+    // Set default importances
+    double imp_n = cimp_n, imp_p = cimp_p, imp_e = cimp_e;
+
+    // Get importances from DAGMC
     if( DAG->has_prop( vol, "imp.n" )) {
       get_real_prop( vol, cellid, "imp.n", imp_n );
     }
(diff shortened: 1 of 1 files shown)


makeclean:
@ljacobson64 could you rebase against develop and then push again?

ljacobson64:
Done

ljacobson64:
Wow, this new Docker CI is really fast

makeclean:
That's the idea :)

gonuke:
Do we know why this fails in CI?

ljacobson64:
Based on the fact that the only check that failed was the one with the master branch of MOAB, I'm assuming it's because I hadn't rebased this since the singleton was removed.

mcnp/mcnp_funcs.cpp
```diff
@@ -476,16 +476,38 @@ void write_lcad_old(std::ofstream &lcadfile)
...
-  double crho, cimp = 1.0;
+  double crho = 0;
+  double cimp_n = 0, cimp_p = 0, cimp_e = 0;
+
+  // Detect which importances are used so all cells, including implicit
+  // complement and graveyard, have these importances
+  bool imp_n_needed = false, imp_p_needed = false, imp_e_needed = false;
+  for( int i = 1; i <= num_cells; ++i ) {
+    moab::EntityHandle vol = DAG->entity_by_index( 3, i );
+    if( DAG->has_prop( vol, "imp.n" )) {
```
gonuke: Is there a way to search for the existence of this tag directly in MOAB without looping through the volumes?  (Maybe not)

mcnp/mcnp_funcs.cpp
```diff
@@ -498,37 +520,53 @@ void write_lcad_old(std::ofstream &lcadfile)
       get_real_prop( vol, cellid, "imp.e", imp_e );
     }
 
-    lcadfile << cellid << " ";
+    // If no importances specified, default to neutron mode
+    if ( ! imp_n_needed && ! imp_p_needed && ! imp_e_needed ) {
```
gonuke: makes more sense to move this up to just after you search for the tags and set `imp_*_needed` rather than inside loop for each cell

mcnp/mcnp_funcs.cpp
```diff
@@ -476,16 +476,38 @@ void write_lcad_old(std::ofstream &lcadfile)
...
+  double cimp_n = 0, cimp_p = 0, cimp_e = 0;
+
+  // Detect which importances are used so all cells, including implicit
+  // complement and graveyard, have these importances
+  bool imp_n_needed = false, imp_p_needed = false, imp_e_needed = false;
+  for( int i = 1; i <= num_cells; ++i ) {
+    moab::EntityHandle vol = DAG->entity_by_index( 3, i );
+    if( DAG->has_prop( vol, "imp.n" )) {
+      imp_n_needed = true;
+      cimp_n = 1;
```
gonuke: reuse of the `cimp_*` variables makes it difficult to understand their intent, and thus to review the method.  Here it looks like they serve as default values, but then later they switch to complement values?  After copying the defaults elsewhere?

mcnp/mcnp_funcs.cpp
```diff
@@ -498,37 +520,53 @@ void write_lcad_old(std::ofstream &lcadfile)
...
+      if( imp_e_needed ) lcadfile << " imp:e=0";
+      lcadfile << " $ graveyard";
       if( DAG->has_prop(vol, "comp") ) {
         // material for the implicit complement has been specified.
         get_int_prop( vol, cellid, "mat", cmat );
         get_real_prop( vol, cellid, "rho", crho );
-        std::cout << "Detected material and density specified for implicit complement: " << cmat <<", " << crho << std::endl;
-        cimp = imp_n;
+        std::cout << "Detected material and density specified for implicit complement: " << cmat << ", " << crho << std::endl;
+        if( imp_n_needed ) cimp_n = imp_n;
```
gonuke: Doesn't this change the "default" values that are reused in line 508 in another iteration of the loop?

mcnp/mcnp_funcs.cpp
makeclean: I would like to change the way that metadata is handled by Dagmc in general, for example at loading store it all in some metadata class so we can query it more neartly. But right now, one has use has_prop.

ljacobson64:
There is a newer version of `mcnp_funcs.cpp` in my MCNP6 branch [here](https://github.com/ljacobson64/DAGMC/blob/mcnp6/mcnp/mcnp_funcs.cpp#L469) that addresses some of these issues.

Instead of having separate variables for `n`, `p`, and `e`, this version contains maps so that all the particle types can be handled. The `cimp` map has two purposes here: one is to specify the importances that should be assigned to the implicit complement, and the other is to keep track of which particles exist anywhere in the problem.

Perhaps it would be best to close this PR and open one for the MCNP6 branch.

PR #408, opened 2016-07-08
Add Docker to DAGMC-CI

makeclean:
This add docker support mostly for building Geant4, which takes the lions share of the current build time. Reduces CI build and test time to around 6 minutes, as opposed to the 30 mins we see currently.


makeclean:
Can someone review this please

PR #426, opened 2016-08-08
Directing mesh tally class to use MOAB's matrix3

Pshriwise:
We've been using a duplicate version of their 3x3 matrix template for some time now. This simply removes the duplicate and directs it to use MOAB's.


makeclean:
Looks good. Thanks @Pshriwise

PR #427, opened 2016-08-10
Update amalgamated pyne

makeclean:
This updates the amalgamated PyNE with a minimised PyNE build. This PyNE includes only what is necessary for the minimal functionality we need from PyNE.


makeclean:
All tests pass, merging

PR #428, opened 2016-08-19
fixed compile error on mac

kkiesling:
This will compile on Mac OS using Clang


makeclean:
This is no longer valid due to @Pshriwise's fix.

PR #429, opened 2016-08-19
Changing VLA to a std::vector and moving local struct out of function.

Pshriwise:
Contains alterations for make_watertight compiled with clang. Specifically, it moves from a VLA to std::vector array. 

A local structure definition, `triangles`, was moved outside of a local function into the zip namespace to fix linux compilation errors as well.

### DIFF
--- a/tools/make_watertight/zip.cpp
+++ b/tools/make_watertight/zip.cpp
@@ -4,22 +4,24 @@
 
 namespace zip
 {
+
+struct triangles {
+  moab::EntityHandle before_tri;
+  const moab::EntityHandle *before;
+  moab::CartVect     before_norm;
+  moab::EntityHandle after0[3];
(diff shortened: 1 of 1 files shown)


Pshriwise:
@kkiesling when you have a chance, can you verify that this will still build with clang after I've moved that structure?

kkiesling:
This built for me on Mac with Clang.

kkiesling:
Are we having issues with the Moab master branch? (Wondering if I can merge this since it passed on all the other branches?)

kkiesling:
@makeclean that is more of a question for you. I need to leave ORNL with a link to where to get the DAGMC tools (for Mac). I would rather leave them with a link to the Svalinn repo than to @Pshriwise's  specific branch.

makeclean:
Looks good, merging.

PR #432, opened 2016-08-25
Removed the pyne nuclear data path, it is no longer required

makeclean:
This PR removes all mentions of pyne::NUC_DATA_PATH as it is now redundant, since the atomic data we need access to is stored directly in memory.


makeclean:
Test pass, except singleton_issue, PR for that incoming.

PR #433, opened 2016-08-29
Dagmc singleton support

makeclean:
This PR removes support for the now defunct DAGMC singleton in the MC code wrappers. This change is needed since we now only support the DAGMC constructor. This has the effect of meaning that we only support MOAB versions 4.9.2 or higher, this in part due to the large change in the DAGMC version number from 1->2 and a number of non-backwards compatible changes in MOAB.


fluka/tests/fludag_unit_tests.cpp
```diff
@@ -45,7 +44,7 @@ class FluDAGTest : public ::testing::Test
     // Default h5m file for testing
     std::string infile = "slabs.h5m";
 
-    rloadval = DAG->load_file(infile.c_str(), 0.0 );
```
Pshriwise: Why the function signature change?

makeclean: This was changed in MOAB, since the faceting tolerance argument no longer serves any useful purpose. Originally this passed the ftol onto CGM to get facets.

Pshriwise: Understood. Thank you!

Pshriwise:
This looks good to me. @makeclean is there anyone else you'd like to take a look at this?

Pshriwise:
Thanks @makeclean!

Issue #437, opened 2016-09-02
[make_watertight] When a volume has no surfaces, we should delete the volume


makeclean:
When make_watertight encounters a surface with no curves, it deletes the surface, but when a volume has no surfaces it does not always delete the volume.

Pshriwise:
@makeclean can you try out [this branch](https://www.github.com/pshriwise/DAGMC/tree/fix_mw_empty_voll) and see if it fixes that problem with the model we were working with yesterday?

makeclean:
Also your current head is good for this too:)

Pshriwise:
Yeah I think I'll rename my latest and PR.

Pshriwise:
...as soon as I add that sample problem you provided as a test, @makeclean

Pshriwise:
Addressed in #438

Issue #436, opened 2016-09-01
[make_watertight] Trelis handling of Spheres


makeclean:
Trelis handles spheres differently to older version of cubit, where it would contain a point curve (zero length curve). This in itself isn't a massive problem, but the logic of make_watertight is such that if there are no curves to a surface, the surface is deleted. make_watertight needs to do additional checks when a surface has no curves, for example it could check to see if the surface is closed, maybe by skinning.

Pshriwise:
@makeclean I think I have this issue resolved as well now. Please try [this branch](https://github.com/pshriwise/DAGMC/fix_trelis_sphere)

makeclean:
Yeah, that's perfect :)

makeclean:
PR away!

Pshriwise:
Taken care of in #438

PR #438, opened 2016-09-06
[make_watertight] fix trelis sphere unit test

Pshriwise:
This PR contains a few changes to make_watertight and a new unit test for make_watertight to address issues #436 and #437

Summary of changes:

* upon deleting a surface, make_watertight now looks at all parent volumes of the surface to check if this was the last surface used to define that volume. If so, the volume is deleted along with the surface. (This condition is reached when there are no curves for the surface, so the curve sets should already be taken care of.)

* make_watertight now accepts the existence of a volume without curves so long as the following conditions are met:
    1) one of its parent volumes is composed of a single surface which is the surface w/o curves
    2) the surface of that volume contains a watertight set of triangles (this is verified by confirming that there are no skin edges returned by the find_skin operation in make_watertight)

* an addition of a unit test for make_watertight which is a sphere produced in Trelis containing no geometric curves. The test verifies that make_watertight reports a successful sealing of the model and that the volume is not deleted in the process.

### DIFF
--- a/tools/make_watertight/gen.cpp
+++ b/tools/make_watertight/gen.cpp
@@ -2071,9 +2071,9 @@ moab::ErrorCode delete_vol(moab::EntityHandle volume)
 {
   // Remove the volume set. This also removes parent-child relationships.
   moab::ErrorCode result;
+  std::cout << "  deleting volume " << gen::geom_id_by_handle(volume)  << std::endl;
   result = MBI()->delete_entities(&volume, 1);
   assert(moab::MB_SUCCESS == result);
-  std::cout << "  deleted volume " << gen::geom_id_by_handle(volume)  << std::endl;
   return result;
 }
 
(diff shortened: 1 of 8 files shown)


tools/make_watertight/gen.cpp
```diff
@@ -1730,6 +1730,28 @@ moab::ErrorCode delete_surface( moab::EntityHandle surf, moab::Tag geom_tag, moa
...
+    moab::EntityHandle parent_vol = *i;
+    std::vector<moab::EntityHandle> child_surfs;
+    result = MBI()->get_child_meshsets(parent_vol, child_surfs);
+    if(gen::error(moab::MB_SUCCESS!=result, "could not get the child surfaces of the volume")) return result;
+    assert(moab::MB_SUCCESS == result);
+
+    // delete volume if it only has this surface as its child
+    if( child_surfs.size() == 1 && child_surfs[0] == surf ) {
+      result = gen::delete_vol(parent_vol);
+      if(gen::error(moab::MB_SUCCESS!=result, "could not delet the parent volume")) return result;
```
makeclean: *delete*

makeclean:
Perfect :)

PR #439, opened 2016-09-12
A simple html implementation of a slide show for our example images.

Pshriwise:
Adding a slideshow for example images of analysis projects. These are not currently linked to specific pages for each project currently, but they could be in the future.

### DIFF
--- a/doc/gallery/gallery.rst
+++ b/doc/gallery/gallery.rst
@@ -5,22 +5,24 @@ Gallery of projects that use DAGMC
 .. toctree::
    :hidden:
     
-   |fnsf/gallery|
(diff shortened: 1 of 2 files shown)


gonuke:
Nice!

gonuke:
We could probably autogenerate this from a simple dictionary of `{filename : "title"}` making it easy for people to add new content.  (Not that it's that hard!)  And not that I'm suggesting it as necessary for this PR!

makeclean:
Could you redo the PR against web_update?

Pshriwise:
@makeclean certainly. Closing.

PR #440, opened 2016-09-13
Adding trelis sphere test to travis run script.

Pshriwise:
This was missed when adding the trelis sphere test before in #438 . It will now run as part of travis CI with this change.


makeclean:
Awesome, thanks!

PR #441, opened 2016-09-13
[make_watertight] regression tests

Pshriwise:
This PR adds new test classes for make_watertight regression tests which will only run with DAGMC's nightly builds (so they won't be run as part of this PR). 

The test models are served from UW box and retrieved by travis via `wget` in the docker instance.

### DIFF
--- a/.run_tests.sh
+++ b/.run_tests.sh
@@ -28,9 +28,9 @@ cd tests
 ./make_watertight_cone_tests
 # if this is not a pull request, run regression tests
 if [ ! TRAVIS_PULL_REQUEST ] ; then
-wget $MW_REG_TEST_MODELS_URL -O mw_reg_test_files.tar.gz -o wget.out
-tar xzvf mw_reg_test_files.tar.gz
-./make_watertight_regression_tests
+    wget $MW_REG_TEST_MODELS_URL -O mw_reg_test_files.tar.gz -o wget.out
+    tar xzvf mw_reg_test_files.tar.gz
+    ./make_watertight_regression_tests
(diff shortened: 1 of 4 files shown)


.run_tests.sh
```diff
@@ -26,6 +26,12 @@ cd tests
 ./make_watertight_no_curve_sphere_tests
 ./make_watertight_cylinder_tests
 ./make_watertight_cone_tests
+# if this is not a pull request, run regression tests
+if [ ! TRAVIS_PULL_REQUEST ] ; then
+wget https://uwmadison.box.com/shared/static/jns2dfu47euucwexe74l4qroxneelp9u.gz -O mw_reg_test_files.tar.gz -o wget.out
```
makeclean: indent with tabs

makeclean: I thought we were going to obfusciate this link and keep in a Travis encrypted variable?

Pshriwise: That's correct. Forgot in my excitement that these are going to be part of the nightly's now. Updated that link to be a Travis environment variable.

Pshriwise:
I've now added necessary environment variables to the docker instance needed to make sure the tests only run for non-pull requests and that the regression test file location can be found by wget using the encrypted travis variable. 

*Note:* any travis account we rely upon will need these variables in the future. So if people begin to test on dagmc forks using travis, they'll need these environment variables as well.

Here is also some proof that the regression tests are successful for a branch push on my fork. 
reg_tests

makeclean:
Looks great :)

PR #442, opened 2016-09-15
[make_watertight] Refactor namespaces to classes and add library.

Pshriwise:
This PR:

1) Alters the structure of `make_watertight` s.t. the `gen`, `arc`, `zip`, `cleanup`, `cw_func`, and `mw_func` namespaces are all now classes. The aim is to allow `make_watertight` to be applied on top of other applications by passing a MOAB interface pointer to the class rather than having it create its own or passing the interface pointer down to each function.

2) Adds a library for `make_watertight` functionality and installs all headers associated with that functionality. Right now these headers are installed in the dagmc `include` directory, but I could also see them going in a sub-directory in the includes dir named something like "make_watertight" or "mw" and am open to that change.

3) This also removes some old test programs/files that are non-essential.

Some of the lower classes like `Gen`, `Zip`, etc. are highly dependent on each other and as such it was difficult to immediately imagine a clear hierarchy for inheritance of classes up to the `MakeWatertight` and `CheckWatertight` level. The result is that some classes will contain instances of each other to account for this. Design change suggestions to help with this problem are very welcome!


Pshriwise:
would someone mind giving this a review plz?

makeclean:
Is there a specific reason that the error function is no longer part of the gen class? I know that you never wrote the function, but it seems odd that one of the arguments true/false if an error occured, and then inside the function never gets called if it wasnt an error!

Pshriwise:
I think I did this so that this particular function could be called from anywhere in the multitude of `make_watertight` classes rather than having to create its own instance of the gen class but it appears that seems to have happened anyways... I can definitely make it part of the Gen class again, but what might be even better is to move to MOAB's recommended set of error macros?

makeclean:
I think shifting to the MOAB macro's might be better, the real question is do you have the time?

Pshriwise:
I should. I'll try to get to it tonight.

Pshriwise:
@makeclean I have a new branch using only MOAB error macros. I've pushed to my fork so that the regression tests run on it. If those succeed, I'll include the changes here.

Pshriwise:
@makeclean @gonuke any further comments on this?

makeclean:
Could you change the build, so we build a "make_watertight" library, just in an internal archive will do, that the executables and tests both link against?

Pshriwise:
I certainly can, but out of curiosity, what's the motivation behind this? I was planning on installing this library and associated headers so that other applications (the Trelis plugins in particular) would be able to use them to access make_watertight's functionality.

Pshriwise:
I'm assuming that most PRs are on a bit of a standstill due to the issues with changes in MOAB. But while waiting on that.. are there any other issues with this one I should address in the meantime?

pshriwise:
Hoping to get this in soon so that other make_watertight fixes can follow. Any further comments?

makeclean:
This is great @Pshriwise thanks :)

PR #445, opened 2016-09-28
Update amalgamated pyne

ljacobson64:
Update the amalgamated pyne to include the changes by @gonuke. Also update the documentation to remove ``src/decay.*`` from the list of files to include.


ljacobson64:
Never mind, closing this because we don't need these changes in DAGMC right now per @makeclean 

Will submit a new PR that only updates the documentation

PR #446, opened 2016-09-28
Update amalgamated pyne documentation

ljacobson64:
Removed ``decay.*`` from the instructions and added more instructions about the manual changes you need to make to produce a working ``pyne.cpp``.


makeclean:
Thanks @ljacobson64

PR #447, opened 2016-10-08
Mandated image height in css. Reduces caption text size and moved to …

Pshriwise:
…bottom to avoid interference. Some pics will now require editing if they are landscape.

@ljacobson64 hammer on this a bit and make sure its working ok?

### DIFF
--- a/doc/slideshow.html
+++ b/doc/slideshow.html
@@ -8,8 +8,9 @@
 
 /* Slideshow container */
  .slideshow-container {
-     max-width: 1000px;
-     max-height: 1000px;
+     max-width: 800px;
+     max-height: 800px;
+     height:550px;
      position: relative;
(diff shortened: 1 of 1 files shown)


gonuke:
Thanks @Pshriwise - I'll go ahead and merge this even though it fails CI because CI doesn't really matter for doc changes

gonuke:
I should have looked a little more carefully... this approach distorts images.  We probably need an approach that computes dimensions based on the size of the image using javascript... or to make the images all the right size in at least one dimension?

gonuke:
This will do for now...

Pshriwise:
@gonuke shoot. I'll take a look and see if I can take care of that (in another PR). Can you give me an example of the distortion to look for?

PR #448, opened 2016-10-09
Increase width of website frames

gonuke:


ljacobson64:
It looks like it's stretching the images to fit the frame; i.e. things that should be circles are becoming ovals

ljacobson64:
Fixed this `here <https://github.com/ljacobson64/DAGMC/commit/d66a59360571baee408319f01e204ea0d16181d2>`_

Sorry for all the extra diffs

PR #450, opened 2016-10-28
Fix material metdata

makeclean:
This PR adds and updates tests for UWUW, previous to these commits metadata from the material objects in a material library were not propgated into the material library inserted into a DAGMC geometry.


uwuw/uwuw_preprocessor.cpp
```diff
@@ -676,9 +677,9 @@ std::string name_concatenator::shift_and_increment(std::string name)
       name[6] = int_as_string[0];
       name[7] = int_as_string[1];
     }  else if(count >= 100 && count < 1000 ) {
-      name[6] = int_as_string[0];
-      name[7] = int_as_string[1];
-      name[8] = int_as_string[2];
+      name[5] = int_as_string[0];
```
pshriwise, line 680: Why the change in indices here?

CMakeLists.txt
```diff
@@ -96,9 +96,7 @@ if (BUILD_FLUKA)
 endif (BUILD_FLUKA)
 
 # Include UWUW in dagmciface library
-if (BUILD_ANY)
-  add_subdirectory(uwuw)
-endif (BUILD_ANY)
+add_subdirectory(uwuw)
```
pshriwise, line 99: So the uwuw preprocessor is now part of the default build? As `make_watertight` and other tools are?

uwuw/uwuw_preprocessor.cpp
makeclean: Because they were wrong, only picked up by the -e on the tests :)

CMakeLists.txt
makeclean: Yeah, its useful without wanting to install the physics packages

uwuw/uwuw_preprocessor.cpp
pshriwise: Stared at it for a bit. Got it.

CMakeLists.txt
pshriwise: Sounds good to me.

PR #451, opened 2016-10-29
Update web

makeclean:
The PR adds all the changes that @ljacobson64, @gonuke and I have worked on (mostly @ljacobson64). We have restructured according to the projected layout and added discussion in all of the major sections. Review welcome.


Pshriwise:
I was going to put some notes on fixes/updates here but I'll address them separately in another place and create a PR (into your branch maybe, @makeclean?). One thing I'll say is that we should really provide an example problem we're confident in for interested parties to run.

In my mind, the ideal sample problem has the following qualities (at minimum):
 - is a geometric model which would not be straight forward to generate in MCNP
 - a model which doesn't take too long to facet and doesn't take up too much space 
       + facets in < 5 min
       + doesn't produce so many facets that it can't be run on a common machine with, say 8GB of memory
- 
 - overall, a model which demonstrates the value of being able to run transport on CAD-based models

PR #455, opened 2016-11-14
[make_watertight] alter conditions for ignoring surface without curves

pshriwise:
This PR alters the conditions under which `make_watertight` will keep a surface without curves rather than deleting it.
These conditions are now: 
 - the surface has no child curves
 - the surface has at least 4 triangles
 - the surface has no skin edges (the surface is a closed body)

Removed from this list is: 
 - ~the surface is the only child of a parent volume~

### DIFF
--- a/tools/make_watertight/MakeWatertight.cpp
+++ b/tools/make_watertight/MakeWatertight.cpp
@@ -1046,7 +1046,7 @@ moab::ErrorCode MakeWatertight::prepare_surfaces(moab::Range &surface_sets,
       MB_CHK_SET_ERR(result, "could not skin the triangles");
       // if the surface is closed, continue to next surface
       if(tris.size() != 4 && skin_edges.size() == 0 && curve_sets.empty()) {
-	continue;
+        continue;
       }
 
       result = gen->delete_surface( *i , geom_tag, tris, surf_id, debug, verbose);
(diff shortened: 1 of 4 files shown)


tools/make_watertight/MakeWatertight.cpp
```diff
@@ -1038,41 +1039,19 @@ moab::ErrorCode MakeWatertight::prepare_surfaces(moab::Range &surface_sets,
...
-          //verify that the surface is closed
-          result = gen->find_skin( tris, 1, skin_edges, false);
-          MB_CHK_SET_ERR(result, "could not skin the triangles");
-          // if the surface is closed, change this indicator
-          if( skin_edges.size() == 0 ) {
-            keep_vol = true;
-          }
+      //verify that the surface is closed
+      moab::Range temp_skin_edges;
+      if( tris.size() >= 4 && curve_sets.empty() ) {
```
makeclean, line 1046: Whats the reason regarding needing at least 4 triangles?

makeclean (commented):
Looks good apart from my query

tools/make_watertight/MakeWatertight.cpp
pshriwise: Well initially it was to make sure that the surface has enough triangle to be a closed surface, but I suppose the skinner check will take confirm that either way.

makeclean:
Awesome, thanks

PR #456, opened 2016-11-15
Set travis test script to exit on error.

pshriwise:
Some failing tests were being missed because travis would continue to run this script despite the exit code of individual commands in the script. This should force the script to exit after the first failing test. If we'd rather all the tests run and the number of errors be accumulated, I can look into that as well.

### DIFF
--- a/.run_tests.sh
+++ b/.run_tests.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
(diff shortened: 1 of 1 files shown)


pshriwise:
Updating w/ @makeclean's changes from #450... this should have tests passing.

makeclean:
Looks great :)

PR #462, opened 2016-11-29
Remove LAPACK dependency

ljacobson64:
This PR removes the LAPACK and BLAS dependencies by copying some of the source files into our repo. This seems to be allowed under the [license](http://www.netlib.org/lapack/#_licensing).


ljacobson64:
Any comments?

makeclean:
Thanks for this @ljacobson64, this is great

PR #469, opened 2016-12-07
Parse metadata before writing fcad

ljacobson64:
as requested by @makeclean


ljacobson64:
Think this is ready now

makeclean:
Awesome, thanks @ljacobson64

PR #470, opened 2016-12-09
Fix tetmesh mcnp6 particles

ljacobson64:
This should fix the issue where MCNP6 particles weren't being scored properly in DAGMC mesh tallies.


tally/TallyManager.cpp
```diff
@@ -244,20 +244,121 @@ Tally *TallyManager::createTally(unsigned int tally_id,
   }
 
   switch (particle) {
-  case 1:
+  case  1:
```
gonuke: Maybe this should be a map/vector now instead of a switch/case statement

makeclean: If you make enum a map like 

      particleList[1] = NEUTRON;
      particleList[2] = PHOTON;
      etc

yeah something like 

    if(TallyInput->particleList[ipt].count == 1) {
       input.particle = TallyInput.particles[ipt];
    } else {
      std::cerr << "no particle found" << std::endl;
      return NULL; 
    }

Although if this is called a lot, something more efficient than a map, or at least some profiling may be in order

makeclean: Or maybe, add an mcnp_ipt function to the particle class in pyne, which we use (the class) elsewhere in the codebase.

ljacobson64:
It also seems to work if you get rid of the enum entirely.

makeclean:
Thanks Lucas this is great, now our mesh tallies are particle type agnostic!

PR #472, opened 2017-01-09
Add some tests and fix others

ljacobson64:
This PR adds 2 tests to test_TrackLengthMeshTally which make sure tetmeshes work for neutrons and protons. It also fixes some existing dagmcnp tests and adds them to the travis run script.

### DIFF
--- a/mcnp/test/CMakeLists.txt
+++ b/mcnp/test/CMakeLists.txt
@@ -18,7 +18,7 @@ include_directories(${GTEST_INCLUDE_DIR})
 # Setup unit tests
 set(TEST_NAME dagmcnp_unit_tests)
 set(DRIVER mcnp5_unit_test_driver.cc)
-add_executable(${test_name} R{TEST_NAME}.cpp ../mcnp_funcs.cpp ${DRIVER})
+add_executable(${TEST_NAME} ${TEST_NAME}.cpp ../mcnp_funcs.cpp ${DRIVER})
 target_link_libraries(${TEST_NAME} "${LINK_LIBS}")
 install(TARGETS ${TEST_NAME} DESTINATION tests
                              PERMISSIONS ${PERMS})
--- a/mcnp/test/CMakeLists.txt
+++ b/mcnp/test/CMakeLists.txt
@@ -10,13 +10,19 @@ list(APPEND LINK_LIBS ${DAG_LINK_LIBS})
 
 # Include directories
 include_directories(..)
+include_directories(../../common)
 include_directories(${MOAB_INCLUDE_DIRS})
 include_directories(${HDF5_INCLUDE_DIRS})
 include_directories(${GTEST_INCLUDE_DIR})
 
 # Setup unit tests
+set(TEST_NAME dagmcnp_unit_tests)
 set(DRIVER mcnp5_unit_test_driver.cc)
-setup_test(dagmcnp_unit_tests cpp ${DRIVER} "${LINK_LIBS}")
+add_executable(${test_name} R{TEST_NAME}.cpp ../mcnp_funcs.cpp ${DRIVER})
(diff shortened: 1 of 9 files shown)


ljacobson64:
Just commenting that the tests are expected to fail for this PR until #470 is merged, at which point the tests should pass

makeclean:
#470 is now merged, can you rebase and push?

ljacobson64:
Done, fingers crossed

ljacobson64:
:(

ljacobson64:
This is failing because the dagmcnp_unit_tests aren't being built, because we aren't using `-DBUILD_MCNP5` or `-DBUILD_MCNP6`, which we can't use because that would require us to build MCNP. The way around this would be to build the MCNP tests every time regardless of whether MCNP is being built.

makeclean:
Yeah we should always build those in testing

ljacobson64:
Perhaps a `-DBUILD_TESTS` flag which we would only use for the CI

makeclean:
Good to go?

ljacobson64:
Yes

mcnp/test/CMakeLists.txt
```diff
@@ -5,17 +5,24 @@ set(LINK_LIBS)
 list(APPEND LINK_LIBS gtest)
 list(APPEND LINK_LIBS pthread)
 list(APPEND LINK_LIBS dagmciface)
+list(APPEND LINK_LIBS dagtally)
 list(APPEND LINK_LIBS ${DAG_LINK_LIBS})
 
 # Include directories
 include_directories(..)
+include_directories(../../common)
```
makeclean: could you change this to ${CMAKE_SOURCE_DIR}/common

mcnp/test/CMakeLists.txt
```diff
@@ -5,17 +5,24 @@ set(LINK_LIBS)
...
 include_directories(${MOAB_INCLUDE_DIRS})
 include_directories(${HDF5_INCLUDE_DIRS})
 include_directories(${GTEST_INCLUDE_DIR})
 
 # Setup unit tests
+set(TEST_NAME dagmcnp_unit_tests)
 set(DRIVER mcnp5_unit_test_driver.cc)
-setup_test(dagmcnp_unit_tests cpp ${DRIVER} "${LINK_LIBS}")
```
makeclean, line 18: Are you changing this for all the tests? What was the reason for this change?

makeclean (changes requested):

mcnp/test/CMakeLists.txt
ljacobson64: mcnp_funcs.cpp needs to be in there for this one, and the setup_test macro doesn't allow for adding additional source files

ljacobson64: To elaborate, mcnp_funcs.cpp needs to be in there when you don't build DAG-MCNP for some reason

ljacobson64:
```
CMakeFiles/dagmcnp_unit_tests.dir/dagmcnp_unit_tests.cpp.o: In function `DAGMCNP5Test_dagmcinit_test_Test::TestBody()':
dagmcnp_unit_tests.cpp:(.text+0x1cf): undefined reference to `dagmcwritemcnp_'
dagmcnp_unit_tests.cpp:(.text+0x3a6): undefined reference to `dagmc_teardown_'
```

mcnp/test/CMakeLists.txt
ljacobson64: Whoops, didn't see this one. Fixed now.

makeclean (commented):
Awesome

makeclean (approved):

PR #474, opened 2017-01-09
Merge master into develop

ljacobson64:
This PR merges master into develop so that develop contains the latest documentation.

It has been suggested that when we want to make changes to the documentation in the future, we PR it into develop instead of master. This would require us to change ``GH_SOURCE_BRANCH`` in ``gh-project.mk`` to be ``develop`` instead of ``master``.

This might also mean that we would stop using the ``update_web`` branch.


gonuke:
Thanks @ljacobson64

PR #475, opened 2017-01-13
Fixed error where graveyard was not assigned a material

makeclean:
In DagGeant4 the graveyard is given the material property vacuum, this is since all the DagSolids are placed within the worldvolume and particles leaving the world volume define the graveyard in Geant4.


ljacobson64:
Looks like it's failing the code style check

PR #476, opened 2017-01-25
Obb builder

makeclean:
This adds the command line tool OBB builder which allows a user to build the tree before running a DAGMC calculation.

### DIFF
--- a/tools/build_obb/CMakeLists.txt
+++ b/tools/build_obb/CMakeLists.txt
@@ -1,4 +1,4 @@
-message(STATUS "Building make_watertight")
+message(STATUS "Building build_obb")
 
 # Source files
 set(SRC_FILES)
--- a/tools/build_obb/CMakeLists.txt
+++ b/tools/build_obb/CMakeLists.txt
@@ -1,0 +1,22 @@
+message(STATUS "Building make_watertight")
+
+# Source files
+set(SRC_FILES)
+list(APPEND SRC_FILES build_obb.cpp)
+
+# Libraries
+set(LINK_LIBS)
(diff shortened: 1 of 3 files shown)


tools/build_obb/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+message(STATUS "Building make_watertight")
```
ljacobson64: Should this be build_obb?

ljacobson64:
Tests are failing because of the style guide check

pshriwise:
Every time

ljacobson64:
It appears that if you don't specify an output file, it does all the work to build the OBB tree but then doesn't give you the fcad file. I think there should be a default filename just like there is for things like mcnp2cad and dagme_preproc... perhaps just call it "fcad"?

ljacobson64:
Otherwise the tool looks like it's working great! I ran a problem using `g=01.h5m`, then with `g=fcad` with the generated fcad from the first run, then with `build_obb 01.h5m -o 01.fcad` and `g=01.fcad` and all three produced identical output.

pshriwise:
What happens if OBBs are already present in the file?

We may want to check for the appropriate tags, remove the trees, and rebuild?

makeclean:
@Pshriwise right now it does nothing, leaves the original tree alone since it just uses the same logic as DAGMC, perhaps that change could be an enhancement more a later date, its likely that functionality should be exposed by DAGMC itself.

ljacobson64:
Looks like that latest merge commit reverted some of the changes you made to address comments

makeclean:
@ljacobson64 ping

ljacobson64:
Thanks @makeclean!

PR #481, opened 2017-02-10
Compile any .cpp files in the mcnp source as well.

pshriwise:


ljacobson64:
Looks good; I'm curious about where we will need this though

pshriwise:
Its needed when DAGMC is being built with the r2s [mesh sampling source](http://pyne.io/usersguide/source_sampling.html#source-sampling-in-mcnp5)

PR #482, opened 2017-02-13
Fix osx prereb

makeclean:
This PR updated DAGMC to allow full compilation of MCNP5/6 & DagGeant4 on OS/X using either the standard OSX clang or package manager install gcc toolchain.


geant4/DagSolid.cc
```diff
@@ -306,7 +309,7 @@ G4double DagSolid::DistanceToIn (const G4ThreeVector &p,
 
 G4double DagSolid::DistanceToIn (const G4ThreeVector &p) const
 {
-
+  //  return 0.0;
```
pshriwise: Can this be removed?

geant4/DagSolid.cc
```diff
@@ -386,6 +389,7 @@ G4double DagSolid::DistanceToOut (const G4ThreeVector &p,
 
 G4double DagSolid::DistanceToOut (const G4ThreeVector &p) const
 {
+  // return 0.0;
```
pshriwise: Same here. remove to reduce clutter?

ljacobson64:
Is this reintroducing the LAPACK dependency?

makeclean:
It shouldn't be, where are you seeing the lapack & blas changes?

tally/CMakeLists.txt
```diff
@@ -14,6 +14,10 @@ file(GLOB PUB_HEADERS ${PUB_HEADERS} "*.hpp")
 # Libraries
 set(LINK_LIBS)
 list(APPEND LINK_LIBS ${DAG_LINK_LIBS})
+list(APPEND LINK_LIBS lapack)
```
ljacobson64: Right here

pshriwise (changes requested):

tally/CMakeLists.txt
```diff
@@ -14,6 +14,8 @@ file(GLOB PUB_HEADERS ${PUB_HEADERS} "*.hpp")
 # Libraries
 set(LINK_LIBS)
 list(APPEND LINK_LIBS ${DAG_LINK_LIBS})
+list(APPEND LINK_LIBS ${GFORTRAN_LIBRARY})
+list(APPEND LINK_LIBS ${QUADMATH_LIBRARY})
```
ljacobson64: This may be a dumb question, but what are libgfortran and libquadmath used for? When I removed the lapack dependency in #462, I was also able to remove the quadmath and gfortran dependencies as well.

makeclean: It was to do with how OS/X finds the libraries, but I had been working on a branch that didnt include your changes. Ill try without

makeclean: No, Fluka needs gfortran, but otherwise the tally library now doesnt.

mcnp/meshtal_funcs.h
```diff
@@ -22,27 +22,29 @@ extern "C" {
...
    */
 #ifndef FORT_FUNC
-
-  /* intel fortran: name mangling is '<module>_mp_<function>_' */
-#ifdef __INTEL_COMPILER
-#define FORT_FUNC( mod, func ) mod##_mp_##func##_
-
+#if defined(__clang__)
+  /* clang/gfortran naming scheme" ___<module>_mod_MOD_<function> */
+#define FORT_FUNC( mod, func ) __##mod##_MOD_##func
```
ljacobson64, line 27: Is this supposed to be 2 underscores as in the code or 3 as in the comment?

makeclean: As it is in the code as opposed to comment

makeclean:
This is all good to go right?

ljacobson64:
It appears this has broken our ability to compile DAGMC with Intel compilers.

makeclean:
I think its just the order of the ifdefs, i think intel has to be checked first, ill reorder and try it.

ljacobson64:
Yeah I think `#elif defined(__CNUC__) || defined(__GNUG__)` is going to evaluate to true pretty much all the time meaning the Intel stuff never gets checked

PR #489, opened 2017-03-03
Enforce more strict code formatting

ljacobson64:
This enforces more strict code formatting. The first commit shows the changes to the astyle_google options, and the second commit formats all the code in the repo.

### DIFF
--- a/.run_tests.sh
+++ b/.run_tests.sh
@@ -44,18 +44,18 @@ cd ..
 # remove the bld dir
 rm -rf bld
 # run astyle to see if there are any differences
-tools/astyle_google --style=linux               \
-                    --indent=spaces=2           \
-                    --pad-oper                  \
-                    --pad-header                \
-                    --unpad-paren               \
-                    --convert-tabs              \
-                    --align-pointer=type        \
-                    --suffix=none               \
-                    --recursive                 \
-                    --formatted                 \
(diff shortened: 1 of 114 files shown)


.run_tests.sh
```diff
@@ -44,6 +44,18 @@ cd ..
...
+                    --pad-header                \
+                    --unpad-paren               \
+                    --convert-tabs              \
+                    --align-pointer=type        \
+                    --suffix=none               \
+                    --recursive                 \
+                    --formatted                 \
+                    --lineend=linux             \
+                    --exclude=gtest             \
+                    *.cc *.cpp *.h *.hh *.hpp
```
makeclean: Could you add quote marks around the asterisked items, i.e. "*.cc" "*.cpp" "*.h" etc?

ljacobson64: How come? if you don't mind me asking

makeclean: Without quote marks barfs on a mac

ljacobson64: That is a good reason

ljacobson64:
Closing this, will re-open when it's actually ready

ljacobson64:
Specifically, I'm waiting for [this bug](https://sourceforge.net/p/astyle/bugs/423) to get fixed so we can use the --max-code-length option.

PR #490, opened 2017-03-03
Fix issue where null characters were getting printed in table 70

### LINKED ISSUE #488: Null characters in MCNP outp files
Print table 70 (surface coefficients) in MCNP outp files contain null characters. This table is not useful for DAGMC runs, but this means that any run that contains the "print" card with no entries will result in an outp file that contains null characters. This is a problem because some text editors cannot open files that contain null characters.

ljacobson64:
Fixes #488


ljacobson64:
Any comments @makeclean ?

makeclean:
Looks good. :) Thanks

PR #491, opened 2017-03-07
Enforce more strict code formatting

ljacobson64:
The new version of astyle (3.0) has a few default options files which format code to specific style guides. One of them is google style, which attempts to follow the [google style guide](https://google.github.io/styleguide/cppguide.html) as closely as possible. The options it uses are in [`tools/astyle/file/google.ini`](https://github.com/ljacobson64/DAGMC/blob/strict_formatting/tools/astyle/file/google.ini). Note that this is different from simply using `--style=google`, which only deals with brace formatting.

I have set it up so that the CI will run astyle using the options in `google.ini`. The CI will also run the style guide check before running the rest of the tests, so that we don't have to wait as long to find out about a style guide failure.


pshriwise:
Looks good to me. @makeclean?

ljacobson64:
The CI should have failed on the second-to-last commit... looking into it

pshriwise:
@ljacobson64 is this ready to go?

ljacobson64:
It's ready as far as I'm concerned

makeclean:
Looks good to me, accepting.

PR #504, opened 2017-03-28
Disable check for zero track length in history_neutral_high

ljacobson64:
This basically does the same thing we did in `charged_particle_history`. If the track length is zero, then it's possible for MCNP to give bad trouble and exit; this PR allows it to be zero without issue.


makeclean:
Could you rebase this against develop?

makeclean:
Great!

PR #505, opened 2017-03-28
Add option to enable mw regression tests. Local builds of dagmc were …

pshriwise:
…failing these tests because the test files did not exist. Removing these tests from a normal build so users aren't alarmed by the failing tests.

### DIFF
--- a/.build_dagmc.sh
+++ b/.build_dagmc.sh
@@ -7,6 +7,7 @@ cmake ../. -DBUILD_TALLY=ON \
            -DBUILD_TESTS=ON \
            -DBUILD_GEANT4=ON \
            -DGEANT4_DIR=/root/geant4.10.00.p02 \
+           -DENABLE_MW_REGRESSION_TESTS=ON \
            -DCMAKE_INSTALL_PREFIX=/root/dagmc
 make 
 make install
--- a/.build_dagmc.sh
+++ b/.build_dagmc.sh
@@ -4,10 +4,9 @@ export LD_LIBRARY_PATH="/root/moab/lib"
 mkdir bld
 cd bld
 cmake ../. -DBUILD_TALLY=ON \
-           -DBUILD_TESTS=ON \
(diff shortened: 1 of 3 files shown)


ljacobson64:
The `-DBUILD_TESTS` option is meant to be used only on the CI; perhaps it would be a good idea to use that option instead of adding a new one.

pshriwise:
I was under the impression that the tests are built and installed by default. This has been my experience anyways. Is this the behavior we expect?

ljacobson64:
The tests are built and installed by default; perhaps `-DBUILD_TESTS` is a bad name. Its purpose is to make sure that the MCNP tests are built on CI even though we don't actually build MCNP on CI. Maybe `-DBUILD_TESTS_CI` would be a better name.

pshriwise:
After looking at the CMake file, I think I see what you're getting at and I agree that `BUILD_CI_TESTS` would be a better name. Shall I update this variable name, place the make_watertight regression tests under that if statement and update the `.build_dagmc.sh` file accordingly?

ljacobson64:
That would be splendid. I'll merge after that's done, assuming it's ok to merge even though the CI is failing because of MOAB master.

pshriwise:
Sounds good. I have a fix for the MOAB issue. PR'ing shortly.

makeclean:
Could you rebase this against develop? Your upstream changes for the CI fix need to go in.

pshriwise:
@makeclean Done

makeclean:
There are no changes needed in run_test.sh right?

pshriwise:
I don't think so. I'll take a look at the CI build to make sure everything is there..

pshriwise:
Travis CI logs are temporarily down :( I'll check this again later and let you know if all of the tests are appearing or not.

pshriwise:
@makeclean Travis logs are working again and the MCNP5 tests are being run, so this is good to go.

makeclean:
Thanks @pshriwise

PR #507, opened 2017-03-31
Adding search suffixes for MOABConfig file in cmake build of MOAB.

pshriwise:
Adding search locations to find MOABConfig.cmake file if MOAB was built and installed using cmake. (This will be the new location for the autotools build as well in MOAB 5.0 so we'll need it)


ljacobson64:
thanks @pshriwise

PR #508, opened 2017-04-07
Mat3 fix

pshriwise:
MOAB is deprecating the Matrix3 constructor used in TrackLengthMeshTally. This PR updates how the Matrix3 object is constructed. Should fix our current problem building against MOAB master.


makeclean:
Thanks!

PR #513, opened 2017-05-09
Adding link flags needed for os/x

makeclean:
OS/X needs to include explicit gfortran flags for the mcnp executables.


makeclean:
This is redundant now.

PR #514, opened 2017-06-06
Add DAGMC library

ljacobson64:
This PR imports the DagMC library from MOAB. The dagmciface library has been removed, and its functionality has been incorporated into the new libdagmc.

I was able to get git to remove all the commits in MOAB that didn't deal with files important to us, so that left us with 240 commits from MOAB out of >7000.

The relevant tests have also been imported from MOAB, but they require some additional MOAB header files which are not published to moab/include, so right now they are not being built.

I expect the CI to pass for the cases built with master and fail for the cases built with MOAB 4.9.2.


ljacobson64:
Or they could all fail too

ljacobson64:
The PR now also removes the embedded lapack files. They are no longer needed as moab now has a lapack dependency anyways.

The 3 tests originally in `moab/test/dagmc` have been ported from moab's testing framework to gtest. They pass on my machine and they have been added to the CI. The 3 tests originally in `moab/dagmc/test` are included in our repo now as well, but they aren't being built because they don't really appear to be unit tests.

ljacobson64:
Travis didn't seem to notice that .travis.yml now wants it to build moab 5.0 instead of 4.9.2

ljacobson64:
Closing and re-opening the PR seemed to do the trick

dagmc/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+# Public headers
+list(APPEND DAGMC_PUB_HEADERS DagMC.hpp)
+list(APPEND DAGMC_PUB_HEADERS dagmcmetadata.hpp)
+
+# Libraries
+set(LINK_LIBS)
+set(LINK_LIBS_STATIC)
+list(APPEND LINK_LIBS        ${DAG_LINK_LIBS})
+list(APPEND LINK_LIBS        pyne_dagmc)
+list(APPEND LINK_LIBS_STATIC ${DAG_LINK_LIBS})
```
makeclean, line 16: Couldnt we just say list(APPEND LINK_LIBS_STATIC ${LINK_LIBS})

dagmc/tools/pt_vol_test.cc
```diff
@@ -0,0 +1,109 @@
+#include "moab/Interface.hpp"
```
makeclean: lets rename the the *.cc files to *.cpp

dagmc/tools/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+# Libraries
+set(LINK_LIBS)
+if (BUILD_STATIC)
+  list(APPEND LINK_LIBS dagmc-static)
+else ()
+  list(APPEND LINK_LIBS dagmc)
+endif (BUILD_STATIC)
+list(APPEND LINK_LIBS ${DAG_LINK_LIBS})
+
+build_tool(pt_vol_test   cc "${LINK_LIBS}")
```
makeclean: Lets renamed the *.cc files to *.cpp

doc/install/dependencies.rst
```diff
@@ -61,7 +61,7 @@ Redhat linux users can do likewise with:
 MOAB installation
 ~~~~~~~~~~~~~~~~~
 
-As of DAGMC version 2.0, MOAB version 4.9.2 or higher is required. The following
+As of DAGMC version 2.0, MOAB version 5.0 or higher is required. The following
```
makeclean: This is DAGMC 3.0 now

dagmc/DagMC.hpp
```diff
@@ -0,0 +1,463 @@
...
+  double density, importance;
+  std::string comp_name;
+};
+
+
+namespace moab {
+
+class CartVect;
+
+#define DAGMC_VERSION 2.0
```
makeclean: Lets bump the version number to 3.0

fluka/CMakeLists.txt
```diff
@@ -12,18 +12,18 @@ set(PUB_HEADERS fluka_funcs.h)
 set(LINK_LIBS)
 set(LINK_LIBS_STATIC)
 list(APPEND LINK_LIBS        ${FLUKA_DIR}/libflukahp.a)
-list(APPEND LINK_LIBS        dagmciface)
+list(APPEND LINK_LIBS        dagmc)
 list(APPEND LINK_LIBS        pyne_dagmc)
 list(APPEND LINK_LIBS        gfortran)
 list(APPEND LINK_LIBS_STATIC ${FLUKA_DIR}/libflukahp.a)
```
makeclean, line 18: again, having fully defined LINK_LIBS we set LINK_LIBS_STATIC, to be LINK_LIBS. In fact does it even have to be defined differently? Dont we just set the compile linkage later to be static, or arhive type for libs?

mcnp/meshtal_funcs.h
```diff
@@ -22,27 +22,12 @@ extern "C" {
...
-#elif ( __GNUC__ == 4)
-/* gcc/gfortran < 4.3: name mangling is '__<module>__<function>' */
-#define FORT_FUNC( mod, func ) __##mod##__##func
 #else
-#error "Some unknown GCC compiler version"
-#endif
-#else
-#error "Some unknown C++ compiler"
-/* Comment out this error to force compile to proceed; it may or may not work */
-#error "DagMC: unknown compiler with unknown fortran name mangling scheme."
```
makeclean: We need to keep the unknown one

makeclean: Especially since you've deprecated the old GCC one, does it need to be deprecated?

pyne/CMakeLists.txt
```diff
@@ -6,14 +6,27 @@ set(SRC_FILES pyne.cpp)
 # Public headers
 set(PUB_HEADERS pyne.h)
 
-# Linked libraries
-set(LINK_LIBS        ${DAG_LINK_LIBS})
-set(LINK_LIBS_STATIC ${DAG_LINK_LIBS})
+# Optimization flags: always use -O0
+set(CMAKE_CXX_FLAGS_RELEASE "-O0")
```
makeclean, line 10: Is this for compile speed? Maybe if build type is Debug we should do O0, maybe we should really optimize it otheriwse? Don't know really, most of the pyne calls arent too deep/slow

makeclean (commented):
I think very good in general, but I think there is too much duplication of link flags. I think we can use the same link flags even with static linkage by using the appropriate cmake command

mcnp/meshtal_funcs.h
ljacobson64: I mainly deprecated it because it was for gfortran 4.3 and earlier, and even the ancient compiler ACI has as its default is 4.4.7. I find it hard to believe that there is anybody out there who is savvy enough to try to compile DAGMC but still has gfortran 4.3.

I will bring the unknown one back though.

ljacobson64: Also, just a FYI, the reason I'm setting these as preprocessor directives and not using, say, `__INTEL_COMPILER` or `__GNUC__` is that those variables are set based on the C++ compiler, not the fortran compiler, because they appear in a C++ file. Since the name mangling in this case depends entirely on the choice of fortran compiler, this seemed like the best way to do it. This wouldn't matter at all if you used gcc/gfortran or icc/ifort, but it becomes a problem if you want to use something like gcc/ifort.

pyne/CMakeLists.txt
ljacobson64: It's an intel thing. If you try to compile pyne.cpp with intel and anything other than `-O0`, it compiles for a few minutes and then crashes my computer due to it running out of memory (I have 16 GB).

ljacobson64:
Someone should probably look over the developer's theory guide as well to make sure the info in there is still accurate. The first sentence currently reads

>This guide is meant to give developers a comprehensive understanding of how DAGMC works as it exists in MOAB and the key components one may use during development.

makeclean:
All done @ljacobson64?

makeclean:
Can confirm builds fine with OS/X, and FC25, as well as on our CI

ljacobson64:
I have a few more commits in [another branch](https://github.com/ljacobson64/DAGMC/compare/add_dagmc...ljacobson64:add_dagmc_2) but I didn't add them to this one because it looks like you've already finished reviewing this one. These are changes I think are good to make, but aren't critical.

The main extra stuff I did is remove unneeded cmake commands, move duplicated commands to the main CMakeLists.txt, and remove the internal libraries. You can see it's a net decrease of 65 lines.

The reason I think we should remove the internal libraries is that they have to be compiled statically, and this means that when we compile shared executables, they don't link against all the things they should. For example, the MCNP executables are currently not linked to libdagmc because the only calls to libdagmc are in the internal mcnpfuncs.a library. The same thing occurs with the internal uwuw_preprocessor.a library.

I can confirm that the new branch works properly for me when compiling with gcc, clang, and intel, all both shared and static. Let me know if you're ok with me adding the commits or if you'd just prefer to merge now that everything works.

makeclean:
I think another PR would be best.

ljacobson64:
Sounds good. Sorry haha I can never seem to leave well enough alone

PR #515, opened 2017-06-12
Add g4hier

makeclean:
This PR adds a class for DagGeant4 that allows the underlying geometry to have a hierarchy imposed upon it with knowledge of the imprint and merge information.


pshriwise:
@makeclean there are some style guide failures here. An easy fix. 

Also, some of the dagsolid test are failing.

I'm aware that this probably isn't at the top of your list to work on, should we try to figure out who might be able to take a look at this? (If anyone else is qualified to do so, that is 😄  )

makeclean:
super stale - if it ever works ill pr again

PR #517, opened 2017-06-26
Add fix for broken test

makeclean:
This should fix the issue seen on CI regarding running out of disk space. This issue is correlated with  the Travis image upgrade - https://blog.travis-ci.com/2017-06-21-trusty-updates-2017-Q2-launch


makeclean:
Note, to test correctly, this needs to be merged to develop, then wait for daily testing to run. If it passes no further action is needed, but if it fails, changes should be reverted and another fix attempted.

makeclean:
Merging, hopefully in time for daily CI

pshriwise:
Question: The intent of committing the docker image is to allow one to open a container and debug failing tests, right? Will one have to retrieve those make_watertight regression test files on their own?

makeclean:
The purpose of committing (in this usage) is to allow store the saved state. We dont actually push the image anywhere such that we can retrieve it. We use it such that the CI image can continue running from where it left off.

ljacobson64:
It looks like the CI doesn't fail until it tries to `sudo docker commit` after running the tests. Since running the tests is the last thing it does, perhaps we could remove that last commit?

PR #519, opened 2017-06-27
Don't docker commit after running tests

ljacobson64:
Perhaps this will fix the problem?


makeclean:
I think a broader refactor into seperate images is better for the long term, and for when we add more linux versions.

ljacobson64:
What about in the mean time?

makeclean:
In the mean time it should serve pretty well, but structurally we need to refactor the docker aspect too.

PR #520, opened 2017-06-29
Remove geant4 10.1 and 10.2 from docker

ljacobson64:
This cuts the size of the docker image from 8.431 GB to 5.946 GB. It also cuts the number of runs in the build matrix from 11 to 9.


makeclean:
Superb, thanks @ljacobson64

ljacobson64:
We could cut out another 1-2 GB by not having it apt-get the compiler versions we're not using; i.e. gcc 4.4, 4.6, 4.7 and clang 3.5, 3.6, 3.7. I'm not sure why I put those in the Dockerfile in the first place since we don't use those compilers on the CI at all.

PR #524, opened 2017-08-01
Fix for MCNP reflecting boundary bug

makeclean:
The PR fixes the bug introduced in the mcnp_funcs refactor, previously due to the faulty logic it was producing incorrect behaviour. This test, and the incoming unit test, addresses this issue.


makeclean:
@ljacobson64 this is good to go, merge when you're happy, or ask for changes :)

ljacobson64:
Looks good, I'll merge once the CI passes. (Looks like the style guide failed)

PR #525, opened 2017-08-17
[bug fix] DAGMC Advanced tallies were not producing vector tagged mes…

makeclean:
This PR modified the underlying routines to fix a number of issues regarding the DAGMC mesh tallies interfacing with pyne.r2s. The routines produced n_group+2 scalar tags, when we should have produced a single n_group sized vector tag. The +2 tags comes form 0-bottom group, and a total group tag. This has been modified for both the KDE tally and Tetmesh tracklength tally. 

A future fix could figure out how automatically tag the total group into a separate TOTAL_TAG, it may need some reworking of the logic, refactor into a common tagging function maybe?

### DIFF
--- a/src/tally/CMakeLists.txt
+++ b/src/tally/CMakeLists.txt
@@ -6,6 +6,8 @@ file(GLOB PUB_HEADERS "*.hpp")
 set(LINK_LIBS lapack dagmc)
 set(LINK_LIBS_EXTERN_NAMES)
 
+set(CMAKE_CXX_STANDARD 11)
+
 dagmc_install_library(dagtally)
 
 if (BUILD_TESTS)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 project(DAGMC)
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.1)
 enable_language(Fortran)
 
 # Set DAGMC version
--- a/src/mcnp/meshtal_funcs.cpp
+++ b/src/mcnp/meshtal_funcs.cpp
@@ -141,7 +141,8 @@ void dagmc_fmesh_setup_mesh_(int* fm_ipt, int* id, int* fmesh_idx,
   std::vector<double> energy_boundaries;
   // if there is more than 1 bin, dont want 0->bottom bin
   int bottom = 0;
-  if(*n_energy_mesh > 2) bottom = 1;
+  if (*n_energy_mesh > 2)
+    bottom = 1;
   for (int i = bottom ; i < *n_energy_mesh; ++i) {
     energy_boundaries.push_back(energy_mesh[i]);
   }
--- a/src/mcnp/meshtal_funcs.cpp
+++ b/src/mcnp/meshtal_funcs.cpp
@@ -139,7 +139,10 @@ void dagmc_fmesh_setup_mesh_(int* fm_ipt, int* id, int* fmesh_idx,
 
   // Copy emesh bin boundaries from MCNP (includes 0.0 MeV)
   std::vector<double> energy_boundaries;
-  for (int i = 0; i < *n_energy_mesh; ++i) {
+  // if there is more than 1 bin, dont want 0->bottom bin
+  int bottom = 0;
+  if(*n_energy_mesh > 2) bottom = 1;
+  for (int i = bottom ; i < *n_energy_mesh; ++i) {
     energy_boundaries.push_back(energy_mesh[i]);
   }
 
--- a/src/tally/KDEMeshTally.cpp
+++ b/src/tally/KDEMeshTally.cpp
@@ -183,7 +183,14 @@ void KDEMeshTally::write_data(double num_histories) {
     moab::EntityHandle point = *i;
     unsigned int point_index = get_entity_index(point);
 
-    for (unsigned int j = 0; j < data->get_num_energy_bins(); ++ j) {
+    unsigned int num_ebins = data->get_num_energy_bins();
+    // if there is a total, dont do anything with it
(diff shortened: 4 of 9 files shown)


makeclean:
Dont merge yet, I want to add a unit test for this.

mcnp/meshtal_funcs.cpp
```diff
@@ -139,7 +139,10 @@ void dagmc_fmesh_setup_mesh_(int* fm_ipt, int* id, int* fmesh_idx,
 
   // Copy emesh bin boundaries from MCNP (includes 0.0 MeV)
   std::vector<double> energy_boundaries;
-  for (int i = 0; i < *n_energy_mesh; ++i) {
+  // if there is more than 1 bin, dont want 0->bottom bin
+  int bottom = 0;
+  if(*n_energy_mesh > 2) bottom = 1;
+  for (int i = bottom ; i < *n_energy_mesh; ++i) {
```
gonuke: Is this because `*n_energy_mesh` includes the bottom of the first bin (0.0 MeV) and thus had N+1 boundaries for N groups?

tally/TrackLengthMeshTally.cpp
```diff
@@ -272,13 +276,31 @@ void TrackLengthMeshTally::write_data(double num_histories) {
         rel_err = sqrt((error / (tally * tally)) - (1. / num_histories));
       }
 
-      rval = mb->tag_set_data(error_tags[j], &t, 1, &rel_err);
-      assert(rval == MB_SUCCESS);
+      tally_vect[j] = score;
+      error_vect[j] = rel_err;
+
+      // TODO: It would be convenient for users to be able to get the total
```
gonuke: But this version doesn't include the total at all... right?  Couldn't we just add an additional tag that is also the total?  In fact, you create the `total_tally_tag` above, but don't use it.

gonuke (changes requested):
Don't we have all the pieces for including the total tally, but just fail to use it?

mcnp/meshtal_funcs.cpp
makeclean: Yes indeed

tally/TrackLengthMeshTally.cpp
makeclean: So right now the Tally API forbids one from accessing tally group data with bins higher than N_GROUPS-1, we could easily change the API and then it would be a trivial change.

makeclean:
Ok, I think this is good to go.

tally/CMakeLists.txt
```diff
@@ -26,6 +26,9 @@ endif (BUILD_SHARED AND ${CMAKE_CXX_COMPILER_ID} STREQUAL Clang)
 add_library(dagtally        SHARED ${SRC_FILES})
 add_library(dagtally-static STATIC ${SRC_FILES})
 
+target_compile_features(dagtally PRIVATE cxx_range_for)
+target_compile_features(dagtally-static PRIVATE cxx_range_for)
```
ljacobson64: What are these for?

makeclean: This is the most cmakey way of enabling c++std's like c++11

ljacobson64: Hmm ok. I don't see anything in here that looks like it needs c++11 though. I can build it fine without these 2 lines with gcc 4.4

ljacobson64: Actually I lied, gcc-4.4 doesn't like std::to_string. I guess my main qualm here is that this PR would make it so it's no longer possible to build DAGMC on ACI with the default compiler

ljacobson64 (approved):
I changed my mind, I don't care about gcc 4.4 anymore.

makeclean:
Just rebased this PR @cadangelo needs this for her work and so does a user here.

ljacobson64:
CI seems to be passing for gcc-6 and gcc-7 but failing for gcc-5, clang-4.0, and clang-5.0. It doesn't seem to like the usage of `to_string`

makeclean:
Yeah I've lost enable c++11 at some point, I'll fix and push again

Get Outlook for Android<https://aka.ms/ghei36>

cadangelo:
@makeclean I built this branch, ran neutron transport, then expanded the vector tag (TALLY_TAG).  There is one less group than there should be.  I haven't fully dug in, but I think the number of energy bins in TallyData.cpp is incorrect.

makeclean:
@cadangelo could you send me associated inputs - or post them somewhere for me to download?

cadangelo:
This PR is good to go!  My previous issue with the number of energy bins was my mistake.

shimwell:
> This PR is good to go! My previous issue with the number of energy bins was my mistake.

Are the outstanding reviews for an outdated versions of this PR?

ljacobson64:
@gonuke are you still looking to see changes here? Or this ok to merge?

gonuke:
This is prob OK.  I was hoping to double check the treatment of the total flux, but that could be a different issue/PR if necessary - this seems to do what it’s supposed to

gonuke (approved):

ljacobson64:
One last thing before I merge -- @makeclean can you add a file to the `news` directory explaining what this does?

gonuke:
Please rebase against the latest updates so we can double check that this passes tests with the right versions of MOAB

src/tally/CMakeLists.txt
```diff
@@ -6,6 +6,8 @@ file(GLOB PUB_HEADERS "*.hpp")
 set(LINK_LIBS lapack dagmc)
 set(LINK_LIBS_EXTERN_NAMES)
 
+set(CMAKE_CXX_STANDARD 11)
```
ljacobson64, line 9: Don't need this anymore because we now have C++11 enabled by default

PR #527, opened 2017-09-29
Cleanup cmake

ljacobson64:
I did this a few months ago but then forgot about it. It only touches CMakeFiles, not any source code. The main things it does are

- moves some duplicated commands into the main cmake file; e.g. `enable_fortran` and some stuff that showed up every time Clang/Intel needed a different setting than everything else
- removes all instances of `enable_testing()` and the `run_test` macro because as far as I can tell they don't do anything
- no longer builds any internal libraries as they were essentially statically linking libraries to executables and other libraries even when static building was disabled
- builds the former internal uwuw library into libdagmc instead
- defaults to a Release build with -O1 for MCNP and -O2 for everything else


pshriwise:
@ljacobson64 why the close?

PR #533, opened 2017-12-01
SDF Class

pshriwise:
This adds the SignedDistanceField (SDF) class to DAGMC. 

The purpose of this class is to store a flat set of data for an implied uniform, structured grid implementation, and its use is to accelerate geometry queries for Monte Carlo codes in which an exact answer is not necessary or to provide a fast safety estimate.

Right now preconditioning for various queries can be added by un-commenting the set of defs at the top of `DagMC.cpp`. After working on implementing this a bit in FLUDAG, I think it makes more sense to call these functions in wrappers to the codes so that more information about the particle's state can be used.


gonuke:
I'm ready to take another look at this if you think it's all still good to go - which I assume you do... (subject to rebase)

pshriwise:
@gonuke sounds good! Rebasing...

ljacobson64:
Don't forget to add a news .rst file!

pshriwise:
@ljacobson64 just added. Thanks for the reminder! Let me know if that looks like the right kind of information we're looking for in those files. Also LMK if the formatting is off. My editor's .rst mode isn't amazing.

ljacobson64:
Looks like the formatting isn't working properly https://github.com/pshriwise/DAGMC/blob/sdf_transfer/news/PR-0533.rst

pshriwise:
Thanks @ljacobson64! Got that taken care of I think.

gonuke:
So.... I just merged @ljacobson64's big file move PR... so I suppose this is now out of date...

pshriwise:
Unless there are any objections, I'm going to close this PR. I'll make sure this work stays available in my fork if needed in the future, but its production applications are limited.

pshriwise:
This PR has been hanging around for a long time and there hasn't been a huge call for it to get in in terms of use cases. I'm going to close it for now. I've protected the branch in my fork, so it'll be around if there's need/interest in the future.

PR #535, opened 2018-01-16
Cleanup cmake

ljacobson64:
I guess there's no better time to submit this than during a hackathon...

* Default to a Release build. This results in optimization flags being used everywhere as appropriate. MCNP is still configured to use no more than `-O1`.
* Rearrange directory structure. Anything with source code is now in its own subdirectory under the `src/` directory.
* Add macros to `cmake/DAGMC_macros.cmake`. This results in much less duplicated cmake elsewhere. For example, the `dagmc_install_library` macro contains all the functionality needed to install a static and shared library at the same time without requiring lots of extra code to be added for every library we build.
* Change how we find HDF5 and MOAB libraries. Previously they were required to be in users' `LD_LIBRARY_PATH`, but the intended use for `LD_LIBRARY_PATH` is for executables to locate libraries at runtime, not compile time. Now, users can specify a `-DHDF5_ROOT` and `-DMOAB_ROOT` when running cmake. The old `LD_LIBRARY_PATH` functionality still works but is deprecated.
* Stop using the CMake file provided by MOAB because it always seems to cause problems for people. Now the DAGMC cmake simply looks for files called `libmoab.so` and `libmoab.a` in the directory specified by `-DMOAB_ROOT`.
* Add `RPATH` functionality so that all executables and libraries automatically know where their dependencies, including compilers, are located. This removes the need for users to add anything to their `LD_LIBRARY_PATH`. This is currently not working for anything that links to Geant4 libraries, and I don't know why.
* Add options to disable building optional targets such as astyle and make_watertight. The default is still to build them as before.
* Remove the mcnpfuncs internal library and replace it with an object library. I think this is a more appropriate way to handle it.
* Remove old, unused, and/or broken things.

Everything listed here works on Linux, but I can't vouch for Mac. Although, from what I have heard, the Mac build is currently broken anyways so this doesn't really change anything in that regard.


pshriwise:
@ljacobson64 looks like the Geant4 build is failing on Travis

pshriwise:
Based on the outline provided above I think that these are generally really helpful changes with the exception of a concern about this item:

  - Stop using the CMake file provided by MOAB because it always seems to cause problems for people. Now the DAGMC cmake simply looks for files called libmoab.so and libmoab.a in the directory specified by -DMOAB_ROOT.

How does DAGMC figure out what dependent libraries MOAB was built with?

pshriwise:
Following up on that previous comment, I guess I'm thinking it would be better submit a to fix the MOAB cmake configuration file than abandon it. (Not that you should necessarily be the one to do it)

ljacobson64:
Yeah it's actually only the static Geant4 build that's failing... I'm looking into it.

As far as I can tell, the only dependent libraries MOAB is built with that matter to us are HDF5 and HDF5's own dependent libraries. Since all the HDF5 libraries are still found using the `FindHDF5` cmake command, the only extra library required by MOAB is libmoab itself. However, I imagine that this could cause problems if a user specified an `HDF5_ROOT` different from the one that was used to build MOAB.

pshriwise:
> As far as I can tell, the only dependent libraries MOAB is built with that matter to us are HDF5 and
> HDF5's own dependent libraries. 

This is true... for now. If we want to support reading or writing to different formats in DAGMC via MOAB, we'll need to link to those libraries in the future.

> Since all the HDF5 libraries are still found using the FindHDF5 cmake command, the only extra library 
> required by MOAB is libmoab itself. However, I imagine that this could cause problems if a user 
> specified an HDF5_ROOT different from the one that was used to build MOAB.

Agreed. We definitely want to make sure we're looking in the same place as the MOAB build.

ljacobson64:
I removed the ability for users to specify where to find HDF5; it now gets that information from the MOAB CMake config file. However, I'm still wary about using the MOAB CMake config file for much else.

The static Dag-Geant4 executable build is still failing for some reason. I thought that updating the docker image to the newest version of Geant4 (10.4) would fix it, but I guess not.

ljacobson64:
Updates:

* Removed the ability for users to specify the location of HDF5. It now looks in the MOAB cmake config file for that information.
* Updated the docker image to include the latest versions of things. Geant4 is also now only built and tested with 2 compilers instead of 7 to decrease the size of the docker image.
* Fixed the Geant4 problem by using `-DGEANT4_USE_SYSTEM_EXPAT=OFF` when building Geant4.
* Updated the documentation to reflect all the changes.
* Removed astyle completely. It's pretty easy to get through a package manager, and if you're on a machine that doesn't have a package manager, then you probably don't need to run astyle anyways. The reason I removed it is that it was having problems being compiled by non-cutting edge compilers.

ljacobson64:
The more I look at this PR, the more I think it really should be multiple PRs. But it would also be a massive headache to try to break it up at this point.

pshriwise:
@ljacobson64 I agree this would've been better off as multiple PRs. There is a lot going on here. The macros for library addition and `RPATH` are good changes, but its hard to sort those out from other things going on in this PR (directory structure changes, docker image & compiler version updates, documentation update to cmake variables). etc. 

These are all valuable changes, but hard to separate and review in this PR. Not sure what the right answer is to make this more review-able...

ljacobson64:
Submitted 2 new PRs to remove the docker and astyle stuff from this PR.

ljacobson64:
I have rebased this PR to make it merge-able again

gonuke:
It is really hard to tell where all the files moved to in this PR.  As much as I don't want to force this to be broken into smaller PR's, I can think of at least three things that are covered here (maybe more?) that are separable:
1. deleting files entirely
2. moving files to a new structure
3. moving CMake macros to a separate place
Are there other things left (now that you already changed the docker and astyle stuff)?

gonuke:
I'm most concerned about knowing which files were removed vs which files were simply moved.  That's very hard to figure out with so many changes...

ljacobson64:
Paul,

I agree with your concerns. Putting all these things in the same PR was easy to do at the time but it does make it a lot harder to review. I'll try to break it up further.

gonuke:
Where are we in the processing of getting all of these changes PR'ed separately?

ljacobson64:
I think there have been 3 PRs merged so far that contained things that were once in this PR:
1. Update the docker image
2. Remove astyle
3. Delete unused files

The next PRs that should be submitted are
1. Rearrange directory structure
2. Make the rest of the changes to the cmake (this could potentially be broken up but I'd have to look it over again to see if there are any obvious changes that wouldn't depend on other changes)

ljacobson64:
I also think that this PR itself should be closed. It would be a nightmare to rebase it at this point. I think it would be easier to make the rest of the changes on top of a fresh branch.

gonuke:
I'm closing this, but it will be around for posterity among closed PRs (don't delete the branch yet...)

PR #536, opened 2018-01-20
Update docker

ljacobson64:
This PR updates the docker image similarly to what I did in #535. Once this is merged, I will update #535 to not include these changes anymore.

Here are the important changes this makes:
* Add gcc-7 and clang-5.0 to supported compilers, remove gcc-4.9, clang-3.8, and clang-3.9 from supported compilers
* Only build and test Geant4 with gcc-7 and clang-5.0; this really helps cut down on the size of the docker image

### DIFF
--- a/docker/run_tests.sh
+++ b/docker/run_tests.sh
@@ -41,7 +41,10 @@ cd ${DAGMC_dir}/tests
 ./dagmc_pointinvol_test
 ./dagmc_rayfire_test
 ./dagmc_simple_test
-./dagsolid_unit_tests
+# not every CI run builds dagsolid
+if [ -f dagsolid_unit_tests ]; then
+  ./dagsolid_unit_tests
+fi
 #./fludag_unit_tests  # no fludag yet
 ./test_CellTally
 ./test_KDEKernel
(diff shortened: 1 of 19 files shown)


pshriwise:
@ljacobson64 thanks for going through the effort of breaking up that PR! I hope it wasn't too painful.

docker/run_tests.sh
```diff
@@ -1,47 +1,55 @@
...
-                          --exclude=tools/astyle \
-                          --exclude=mcnp/mcnp5/Source \
-                          --exclude=mcnp/mcnp6/Source \
-                          --ignore-exclude-errors \
-                          --recursive \
-                          --verbose \
-                          --formatted \
-                          "*.cc" "*.cpp" "*.h" "*.hh" "*.hpp"
-git diff --exit-code  # Exit if astyle found diffs
+astyle_deb=astyle_3.0.1-1ubuntu1_amd64.deb
```
pshriwise, line 25: So after this PR, we still build `astyle` with DAGMC but don't use it to test right?

pshriwise: That's fine by me btw. The removal of `astyle` is its own PR.

pshriwise (approved):
LGTM

docker/run_tests.sh
ljacobson64: Yes. Also you'll see that we have to download a .deb straight from the internet instead of using the package manager. This is because we need to use astyle version 3, while all ubuntu versions before 18.04 can only get version 2.06.

pshriwise: Personally, I don't think that's a big deal but why do we need version 3?

ljacobson64: Version 2 was segfaulting somewhere in our code

ljacobson64:
Thanks for the quick review!

PR #537, opened 2018-01-20
Remove astyle

ljacobson64:
This PR removes astyle similarly to what I did in #535. Once this is merged, I will update #535 to not include these changes anymore.

This PR depends on #536.


doc/contribute/contributeguide.rst
```diff
@@ -179,12 +179,27 @@ to isolate exactly which type is being referenced.
...
-DAGMC conforms to the Google C++ style guide, and we have included a C++ style
-guide formatter to make developers' lives much easier. When you have added all
-the features you want to add, the style guide formatter should be run.
+DAGMC conforms to the Google C++ style guide. We use the Astyle_ code formatter
+to make developers' lives easier. Here is how to install Astyle on Ubuntu:
 ::
 
-    $ astyle --style=linux --indent=spaces=2
+    $ wget http://archive.ubuntu.com/ubuntu/pool/universe/a/astyle/astyle_3.0.1-1ubuntu1_amd64.deb
+    $ dpkg -i astyle_3.0.1-1ubuntu1_amd64.deb
```
pshriwise: Does there need to be a `sudo` in front of this command?

ljacobson64: I guess in the documentation it should have sudo, yes. I'll fix that. I just copied those lines from the docker scripts which automatically run with sudo.

ljacobson64: This is ready now imo

pshriwise:
So that I understand the purpose of this PR, was there a reason that the `astyle` program was added to DAGMC in the first place? If so, does that reason no longer exist? 

Or is this simply to remove support of unrelated code in the repo?

ljacobson64:
I'm not entirely sure why it was added in the first place. It looks like astyle was available from apt at least as far back as ubuntu 14.04.

One of the reasons I want to get rid of it is that v3 requires c++11, which makes it impossible to compile with older compilers. For example, if you're trying to build DAG-MCNP5/6 on ACI, which has gcc 4.4.7, everything will work fine except astyle, which will kill the build. There is currently no way to turn off building astyle, although support for turning off individual components is added in #535.

The main reason I want to get rid of it, however, is that it's 25000 lines that don't have anything to do with DAGMC. And I guess it's my opinion that anybody doing development on DAGMC probably already has the wherewithal to get astyle by themselves, so we don't need to include it in our repo.

pshriwise:
I'm really going back and forth on this. I think astyle is pretty straightforward to install on Linux, but I'm not sure about Windows/OSX. My specific concern is creating a barrier to contribution for outside developers. It could be frustrating to them to have PR's held up because of format testing. The installation of extra software which they won't use often just to submit code might not be attractive either.

On the other hand, it would be nice to remove unrelated code form the repository, and in practice it's largely CNERG users who contribute to the repo (a.k.a. users with significant linux experience). These users also likely will contribute to the repo repeatedly so asking them to install the program makes more sense.

I dunno @ljacobson64, what are your thoughts?

makeclean:
So as @pshriwise the original intent of the addition of astyle was to lower the bar to access, particularly for non linux users i.e. OS/X (and to some extent windows). I felt like the easier we made it for people to contribute to the project the better, which for us is always an issues regarding our dependency stack. In my mind it was one less thing someone has to install, and it always just there.

pshriwise:
After a lot of discussion on this, @ljacobson64 and I came to the conclusion that support for building `astyle` in DAGMC can be removed as long as issues #539 and #540 are addressed. These two proposed changes will: 

  1. make it easier for submitters to identify CI failure due to style guide requirements

  2. make it easier for the PR creator to apply those changes without having to install or build `astyle` themselves, regardless of their development platform

Any other thoughts @makeclean @ljacobson64?

If not, I think this PR is good to do and we can get to work on those issues shortly.

ljacobson64:
Closed since all the commits in this PR were merged as part of #539

PR #538, opened 2018-01-22
update hdf5-1.8.13 source directory

zxkjack123:
Update the document as the source directory of hdf5-1.8.13 was changed from "https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.8.13/src/hdf5-1.8.13.tar.gz" to "https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.8/hdf5-1.8.13/src/hdf5-1.8.13.tar.gz".

### DIFF
--- a/doc/install/dependencies.rst
+++ b/doc/install/dependencies.rst
@@ -35,7 +35,7 @@ HDF5 from source.
 
     $ mkdir -p $HOME/dagmc_bld/HDF5/bld
     $ cd $HOME/dagmc_bld/HDF5
-    $ wget https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.8.13/src/hdf5-1.8.13.tar.gz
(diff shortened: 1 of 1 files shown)


ljacobson64:
Great catch!

ljacobson64 (approved):

pshriwise:
I'm really going back and forth on this. I think astyle is pretty straightforward to install on Linux, but I'm not sure about Windows/OSX. My specific concern is creating a barrier to contribution for outside developers. It could be frustrating to them to have PR's held up because of format testing. The installation of extra software which they won't use often just to submit code might not be attractive either.

On the other hand, it would be nice to remove unrelated code form the repository, and in practice it's largely CNERG users who contribute to the repo (a.k.a. users with significant linux experience). These users also likely will contribute to the repo repeatedly so asking them to install the program makes more sense. 

I dunno @ljacobson64, what are your thoughts?

Issue #539, opened 2018-01-23
[Style Guide] Create separate build for testing style


pshriwise:
Stemming from discussions in #537, it would be beneficial to create a separate build in Travis for checking the code formatting and style in a pull request. This will make it easier to identify when CI has failed due to style guide issues.

pshriwise:
One specific improvement:

  - Error in the Travis output should be clear about why the test has failed using some standard message rather than displaying the diff

ljacobson64:
Addressed by #541

PR #541, opened 2018-01-23
Create a separate CI run that only runs the style guide checker

ljacobson64:
Addressing #539. I don't know whether this is going to work or not, but the only way to test it is to submit the PR.

This PR depends on #537.


docker/run_astyle.sh
```diff
@@ -0,0 +1,27 @@
...
+       --verbose \
+       --formatted \
+       "*.cc" "*.cpp" "*.h" "*.hh" "*.hpp"
+
+# Exit if astyle found diffs
+diffs=`git status --porcelain`
+if [ -z "${diffs}" ]; then
+  echo "Style guide checker passed!"
+else
+  echo "Style guide checker failed. Please run astyle."
```
pshriwise: A couple small things here:

  1. Can you add ERROR (or something along those lines) at the beginning of the "Style guide checker failed..." statement? This will make it easier to find the error for those who aren't used to all that other output coming from travis.

  2. I think we should output the diff after the error message here if the test fails.

pshriwise (commented):
A couple small things. Then this looks ready to go to me!

docker/run_astyle.sh
ljacobson64: Done and done

pshriwise:
Awesome. Thanks @ljacobson64. This is a nice improvement to the CI.

PR #552, opened 2018-04-27
Rearrange dirs

ljacobson64:
This PR moves all source code into the `src` directory. This is part of the effort to break up the now-closed #535 into smaller chunks.


gonuke:
I don't think it's really possible to review this, and since the tests pass.... I'll go for it.

gonuke:
thanks @ljacobson64

PR #556, opened 2018-08-05
Remove mangling

ljacobson64:
This PR removes the need for name mangling in `meshtal_funcs.h` by making use of `bind(c)` in `fmesh_mod.F90`. Previously, there was a macro in `meshtal_funcs.h` that attempted to determine the correct name mangling, but I've found situations where it didn't work properly. This is a much simpler way to do it.

(spoilers)

The real reason I did this is that MCNP6.2 uses Fortran2003, and from what I can tell, Intel's name mangling scheme for Fortran2003 is slightly different than it is for Fortran90. (It could be caused by something else, but regardless, the mangling is different between MCNP6.2 and all other versions of MCNP.) Trying to fix the name mangling macro would be pretty ugly because a different name mangling would need to be used for DAG-MCNP5 (Fortran90) than DAG-MCNP6.2 (Fortran2003) when trying to build both at the same time. This PR completely removes that problem, and it will remove any future problems related to name mangling as well.

I've gotten DAG-MCNP6.2 working (or at least compiling) on top of these changes, but support for that will come in a future PR.

### DIFF
--- a/news/PR-0556.rst
+++ b/news/PR-0556.rst
@@ -3,6 +3,7 @@
 **Changed:**
 
 * Use bind(c) in fmesh_mod.F90 to avoid the need for name mangling on the C++ side
+* Rename MCNP patch files to mcnpXXX.patch, where XXX is the version turned into a 3-digit number
 
 **Deprecated:** None
 
--- a/news/PR-0556.rst
+++ b/news/PR-0556.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:**
+
+* Use bind(c) in fmesh_mod.F90 to avoid the need for name mangling on the C++ side
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/src/mcnp/meshtal_funcs.h
+++ b/src/mcnp/meshtal_funcs.h
@@ -18,28 +18,8 @@ extern "C" {
  * and should only be called from within meshtal_funcs.cpp
  */
 
-/** FORT_FUNC:
- * Macro to access symbol of fortran function 'func' in module 'mod'
- */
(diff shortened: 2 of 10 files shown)


pshriwise:
This is great! I agree - this is the right way to handle name mangling long-term. Will review shortly (once Travis has completed).

src/mcnp/meshtal_funcs.h
```diff
@@ -18,28 +18,8 @@ extern "C" {
...
-#warning "Unknown fortran compiler with unknown name mangling scheme, \
-proceeding with gfortran's scheme"
-#define FORT_FUNC( mod, func ) __##mod##_MOD_##func
-#endif
-#endif
-
-#define FMESH_FUNC( func ) FORT_FUNC( fmesh_mod, func )
-
 /* Make a valid Fortran pointer to a C array */
-extern void FMESH_FUNC(dagmc_make_fortran_pointer)(void* fort_ref, double* array, int* size);
```
pshriwise, line 42: Good to see this go...

news/PR-0556.rst
```diff
@@ -0,0 +1,13 @@
+**Added:** None
+
+**Changed:**
+
+* Use bind(c) in fmesh_mod.F90 to avoid the need for name mangling on the C++ side
```
pshriwise, line 5: Can you add renaming of the patch files here? This will be important for external users.

pshriwise (commented):
LGTM! Just one small request. 

Will wait until the end of the day on Monday to merge so others can take a look.

news/PR-0556.rst
ljacobson64: Good call. Done.

pshriwise (approved):
Thanks for the updates @ljacobson64! LGTM

PR #560, opened 2018-09-14
Correcting namespace of CartVects for tracing DAGMC calls.

pshriwise:
This is a small correction to a couple CartVect variables (adding moab namespace). They are only used when tracing DagMC calls which is why we haven't run into this before.


pshriwise:
Looking for a quick look here. It's a pretty minor change...

ljacobson64 (approved):

ljacobson64:
Thanks @pshriwise

PR #563, opened 2018-10-01
Adding check for DAGMC_ROOT variable in environment.

pshriwise:
This PR adds the option for users to provide a DAGMC install location via the environment variable `DAGMC_ROOT`.

### DIFF
--- a/cmake/FindDAGMC.cmake
+++ b/cmake/FindDAGMC.cmake
@@ -10,7 +10,7 @@
 message(STATUS ${DAGMC_ROOT})
 
 find_path(DAGMC_CMAKE_CONFIG NAMES DAGMCConfig.cmake
-          HINTS ${DAGMC_ROOT}
+          HINTS ${DAGMC_ROOT} $ENV{DAGMC_ROOT}
           PATHS ENV LD_LIBRARY_PATH
           PATH_SUFFIXES lib Lib cmake lib/cmake/
           NO_DEFAULT_PATH)


cmake/FindDAGMC.cmake
```diff
@@ -10,7 +10,7 @@
 message(STATUS ${DAGMC_ROOT})
 
 find_path(DAGMC_CMAKE_CONFIG NAMES DAGMCConfig.cmake
-          HINTS ${DAGMC_ROOT}
+          HINTS ${DAGMC_ROOT} $ENV{DAGMC_ROOT}
```
ljacobson64, line 13: Does this work? I thought the correct syntax was `HINTS ${DAGMC_ROOT} ENV DAGMC_ROOT`

pshriwise: Just tested it out with the OpenMC project and it seems to work. 

Documentation on this use of `ENV` is [here](https://cmake.org/cmake/help/v3.0/variable/ENV.html). The syntax you included may be equivalent though, I'm not sure.

ljacobson64: I was going off of what it says [here](https://cmake.org/cmake/help/v3.8/command/find_path.html). Regardless, I guess they do the same thing.

ljacobson64 (approved):

ljacobson64:
Thanks @pshriwise!

PR #566, opened 2018-10-22
Metadata Structure Documentation

pshriwise:
This PR adds information about how metadata is structured in our DAGMC/MOAB files for future reference and a little bit of information about the DAGMC Metadata interface used to access this info. 

I wasn't sure which section should come first, so if anyone has a strong opinion on that please let me know.


gonuke:
Thanks @pshriwise - this looks like a great start.  One question that I have always forgotten to write down: what is the flow of this metadata from Cubit/Trelis ot simulation time?  That is, I know it is something like this:
* Cubi/Trelis: groups, metadata encoded in group names, groups contain geom entities to which the meta data applies
* dagmc_export: EntitySets, metadata encoded in "NAME" tag of EntitySet, EntitySet contains topological geometry sets to which the meta data applies
* uwuw_preproc (?): ????

pshriwise:
The first bullet is covered in the user's guide under the UWUW section. Which makes sense (in my brain) because model generation in Trelis is a recommended workflow.

I think the second bullet is covered in the `Metadata Structure in DAGMC Files` section. I could certainly word it differently and am happy to take suggestions on that.

I debated about whether or not this fit into the dev theory guide. A shallow explanation of what `uwuw_preproc` does is provided in the UWUW section of the user's guide. While I could go into more detail as to what uwuw_preproc does here, I'm not sure it would be very helpful.

pshriwise:
Thinking about this some more this morning, I *could* (and will) provide some information on the UWUW class and how it is used in various DAGMC implementations.

pshriwise:
Added some info on use of the UWUW interface. Let me know if there is anything else I should include!

pshriwise:
Is there anything I should include here or be more specific on @gonuke @makeclean?

doc/contribute/devtheory.rst
```diff
@@ -202,3 +202,112 @@ the known current volume, the next volume is determined by looking at the other
 volume tagged on that surface (as described in the Sense Tags section above).
 This assumes that a valid surface and volume are provided. If no next volume
 exists, then the call will return 0 for the next volume.
+
+DAGMC Metadata
+~~~~~~~~~~~~~~
+
+DAGMC's Metadata Interface
```
gonuke, line 230: I feel as though this section may be better coming after the next section.  That is, first describe the data in file and then describe the first s/w interface to access this data.

gonuke: With that change, you can also include a phrase/sentence in this section about the fact that this data is unpacked from the MOAB entity sets described in the other section.

doc/contribute/devtheory.rst
```diff
@@ -202,3 +202,112 @@ the known current volume, the next volume is determined by looking at the other
...
+entry are contained by the metadata entity set.
+
+Metadata entity sets are also tagged with a `CATEGORY` tag (similar to geometric
+sets) with the value "Group" to indicate that the entity set's purpose is to
+group geometric entities together. These "Group" sets can be gathered using this
+tag and value to identify "Group" entity sets in the MOAB instance.
+
+.. [1] only used in DAG-MCNP simulations.
+
+The University of Wisconsin Unified Workflow (UWUW) Interface
```
gonuke, line 258: How does the metadata API interact with the UWUW API?  (If they do...) That's not clear to me.

gonuke (commented):
I have a few small comments, but I think it's otherwise good for now.

doc/contribute/devtheory.rst
pshriwise: I like it. I'll make that change shortly.

doc/contribute/devtheory.rst
pshriwise: I'll have to look into that. I think they do. I know the data model in MOAB quite well, but I think @makeclean knows most about these classes and how they connect. I'll take a look at the code and double check my understanding with him then submit some updates.

gonuke: Not looking for too much - I don’t want to hold it up too much

pshriwise:
@gonuke, I've just pushed some changes. A couple things I'd appreciate some comments on: 

  - I've moved to using "EntitySet" instead of "entity set" in the sections I wrote. I've gone back and forth on this, but I think that using "EntitySet" is best. Because this area of documentation is oriented toward developers and contribution. I feel it's good to consistently use terms with that format given they'll run into them when working with the code (in MOAB documentation and other places). Right now, other areas of the document use "entity set" and we should be consistent so I'd be happy to update those as well, but I wanted to discuss it here first.

  - More of an FYI, some of the information reports the current state, not necessarily how I think it should be. For example, I'd probably recommend that `DAGMCMetadata::unpack_string` live somewhere neutral if we're going to use it in both classes.

gonuke:
Thanks @pshriwise  - this looks good to me.

Issue #571, opened 2018-10-29
Required MOAB shared/static build.


pshriwise:
[This line](https://github.com/svalinn/DAGMC/blob/develop/cmake/FindMOAB.cmake#L84) in FindMOAB.cmake causes us to require that MOAB was built with shared libraries... do we want to always require this. Can we require it only if DAGMC is being built with static libs? 

I know a shared static and shared build is now the default for DAGMC, but as it stands we don't present a pathway to building with shared libraries at all. Same is true for a static lib only pathway...

ljacobson64:
I can try to come up with a way to fix this. I imagine it'll look something like adding the cmake variables `CMAKE_BUILD_STATIC_LIBS` and `CMAKE_BUILD_SHARED_LIBS` that default to `ON` but are allowed to be turned off. Then, we'll only search for static/shared external libraries (e.g. MOAB) as needed.

pshriwise:
Sounds good. The current state breaks some downstream work so the sooner the better. :)

ljacobson64:
@pshriwise same question here... did #572 solve this?

pshriwise:
I think so.... let me double check and then I'll close it.

pshriwise:
It seems to me we're good, yup. For anyone who might find this issue in the future, the default is to build both static and shared libraries as #572. To build DAGMC with only static or shared libraries the `BUILD_STATIC_LIBS` and `BUILD_SHARED_LIBS` options will need to be turned off in the cmake command. (e.g. `-DBUILD_STATIC_LIBS=OFF`)

PR #572, opened 2018-10-30
Allow users to disable building static or shared libraries

ljacobson64:
Addressing #571


cmake/DAGMC_macros.cmake
```diff
@@ -69,6 +72,16 @@ macro (dagmc_setup_options)
     set(BUILD_GEANT4 ON)
     set(BUILD_FLUKA  ON)
   endif ()
+
+  if (BUILD_SHARED_LIBS AND NOT BUILD_STATIC_LIBS)
+    set(BUILD_STATIC_EXE OFF)
```
pshriwise, line 77: Is it possible (without a massive effort) to have CMake build both?

gonuke:
Any good ideas for how to test this? @pshriwise?

ljacobson64:
Which part of this are you looking to test?

The CI already builds DAGMC twice to make sure it can build both shared and static executables. This PR didn't change anything in the CI, so it still builds both types of libraries both times. We could configure it to only build shared libraries for one and only build static libraries for the other, but then we wouldn't be testing whether we can build both at the same time. This also wouldn't test whether it's possible to build DAGMC if MOAB only has one type of library.

gonuke:
Do we build anything that links to the DAGMC libraries?

ljacobson64:
In the `cmake/test_config` directory, we compile a sample program that links to the `DAGMC_LIBRARIES` and prints out the DAGMC version string. The CI runs this.

pshriwise:
@ljacobson64 alluded to the core issue here. We currently expect that both sets of libraries are built with MOAB. This is what caused problems in applications that don't need the static DAGMC libs (pyne, openmc). 

Adding builds where MOAB is built shared/static/both and autotools/cmake and corresponding DAGMC configurations is how we'd cover all of the possibilities here, thought I'm not sure what kind of CI hit we'd take as a result of all these added branches. There may be ways around this, however.

ljacobson64:
A hacky way to do it would be to move the MOAB static libraries to a temp directory when building DAGMC shared libraries and move the MOAB shared libraries to a temp directory when building DAGMC static libraries.

ljacobson64:
@pshriwise now that I think about it, I'm confused as to why this wasn't a problem before. My PR did not introduce the idea of requiring DAGMC to be built with both static and shared libraries; it was already like that. How was your workflow not broken before my PR?

pshriwise:
The main change is that now DAGMC requires MOAB to be built with both static and shared libraries. This may not be necessary? Perhaps we really only need changes in `FindMOAB.cmake`?

ljacobson64:
I thought it was already like that though. Before my PR, DAGMC was already required to build both static and shared libraries, and the only way that's possible is if MOAB is also built with both static and shared libraries.

pshriwise:
It doesn't appear so...

From [here](http://docencia.ac.upc.edu/FIB/USO/Bibliografia/unix-c-libraries.html)

Static Linking Vs. Dynamic Linking
----------------------------------------------------

When we discussed static libraries we said that the linker will try to look for a file named 'libutil.a'. We lied. Before looking for such a file, it will look for a file named 'libutil.so' - as a shared library. Only if it cannot find a shared library, will it look for 'libutil.a' as a static library. Thus, if we have created two copies of the library, one static and one shared, the shared will be preferred. This can be overridden using some linker flags ('-Wl,static' with some linkers, '-Bstatic' with other types of linkers. refer to the compiler's or the linker's manual for info about these flags).

pshriwise:
I haven't combed through the CMake to check if we set this preference, but it's possible(likely?) that we could be linking to the shared MOAB libs even if the static ones are present.

pshriwise:
So... I'm not sure where we're at here... should we have a live discussion about our expectations of the MOAB build?

This morning I built DAGMC on top of a MOAB built with both autotools (static/shared) and CMake (shared only). Previous versions of DAGMC were happy to build with shared MOAB libs only and the output of `nm --demangle libdagmc.so | grep Core` showed that MOAB symbols were undefined in both cases:

```
U moab::Core::Core()
```

This indicates to me that regardless of whether the MOAB static libs are present, the linker is using the shared libraries.

pshriwise:
ping. we should resolve this...

pshriwise (approved):
For posterity's sake: @ljacobson64 and I talked a bit and we're happy with these changes or now to allow users to build static, shared, or both DAGMC libraries.

We'll continue the discussion below about MOAB library expectations in issue #577, but it'd be best to get these changes added IMO.

cmake/DAGMC_macros.cmake
ljacobson64: Probably, but we'd need to have a discussion about where to install them. It's convenient to install both types of libraries because they have different filenames (`.so` vs. `.a`), but the executables have the same name.

gonuke: Do we need both sets of executables?  Seems like we need both libraries but not 2 sets of executables

ljacobson64: I can't think of a reason to build both outside of testing purposes. I can't think of any other package that lets you build both either

gonuke:
Thanks for this update @ljacobson64

PR #578, opened 2018-11-28
Support for new MOABConfig.cmake

ljacobson64:
@pshriwise tells me that the MOAB team is planning to update its cmake configure file such that the variables `MOAB_LIBRARY_DIRS` and `MOAB_INCLUDE_DIRS` are set properly. This PR updates our `FindMOAB.cmake` file to use those variables as opposed to trying to figure them out for itself.

I'm still not a super big fan of MOAB's `MOAB_LIBRARIES` variable, so this PR does not attempt to have DAGMC use it.

This PR does one more thing: it changes the `MOAB_DIR` variable to `MOAB_ROOT`. This is a more conventional variable name.

This PR should probably not be approved until these changes are merged into the main development branch of MOAB. Additionally, after approving this this PR, DAGMC will not be backwards compatible with MOAB 5.0. So we will need to update the CI and have a discussion about what versions of MOAB we want to support.


pshriwise:
Hmmm this doesn't seem to be passing tests...

ljacobson64:
@pshriwise and I discussed this. The reason the CI is failing is that some of its tests use MOAB v5.0, which is no longer supported in this PR. The tests using MOAB master are passing. We decided to wait until the next version number (5.1) is released, and then I will change the CI config to use 5.1 instead of 5.0.

gonuke:
Recommended tests:
- build MOAB-master with and without EIGEN3 flag, on systems with and without EIGEN3 existing
- build DAGMC against those MOABs using this file - all combinations should work

gonuke:
MOAB with cmake:
`mkdir bld; cd bld; cmake .. -DENABLE_HDF5=ON -DENABLE_BLASLAPACK=ON|OFF`

ljacobson64:
This branch appears to work correctly with these moab configurations:

1. Autotools, without eigen, system has eigen
2. CMake, without eigen, system has eigen
3. Autotools with eigen, system has eigen
4. CMake with eigen, system has eigen
5. Autotools, without eigen, system does not have eigen
6. CMake, without eigen, system does not have eigen

moab does not (and should not) build with these configurations:

1. Autotools, with eigen, system does not have eigen
2. CMake, with eigen, system does not have eigen

pshriwise:
This is on my list to test with external apps in the next week or so. I'll merge when I've done that successfully if no one else has made comments by then.

ljacobson64:
I see that MOAB v5.1.0 has been released. I have updated this PR to use 5.1.0 on the CI.

gonuke (approved):
LGTM - Let's see if @pshriwise wants to comment/test further, or we can just merge

pshriwise:
After a little investigating, I've found that the CMake convention for supporting `<packagename>_ROOT` variables in `find_package` is relatively new, and that a user-defined variable `MOAB_ROOT` may be ignored in older versions of CMake, causing some existing  (more info [here](https://cmake.org/cmake/help/git-stage/policy/CMP0074.html)).

Do we want to circumvent this problem somehow, or move forward and support only newer versions of CMake?

ljacobson64:
I just tested this branch with cmake 3.0.2 (the earliest version we support) and it seemed to compile fine. I believe the CI uses 3.5.2, and it's fine there as well. Ubuntu 18.04 uses 3.10.2 and it's also fine there.

pshriwise:
True... I was just looking at CI, my system etc. and they all work fine on older versions.

But I may have been misinterpreting the policy documentation. It may actually be a problem going forward in CMake versions. For example, I'm having problems on OpenMC's CI system using 3.12.4. I think that policy related to `find_packages` (CMP0074) needs to be set in newer versions of CMake where it didn't need to be before?

ljacobson64:
Ah, I think I see what's going on here. In cmake 3.12+, `<package_name>_ROOT` variables are supposed to be intrinsic variables used by `find_package()` and are not meant to be used directly.

So perhaps it would be better to go back to using `MOAB_DIR` instead of `MOAB_ROOT`, because cmake doesn't assign any special meaning to `<package_name>_DIR` variables.

pshriwise:
The only problem there is that they use the variable `MOAB_DIR` in their `MOABConfig.cmake`. This was causing problems, and is why I wanted to get ahead of this change. I think we can still use `MOAB_ROOT` we just have to enable this policy for CMake to support it. I'm verifying this on OpenMC's CI right now and will get back to you.

ljacobson64:
This seems to make the warning go away (and have it still work in versions of cmake that don't know what CMP0074 is):

```
if (CMAKE_VERSION VERSION_GREATER 3.11)
  cmake_policy(SET CMP0074 OLD)
endif ()
```

pshriwise:
CMP is short for "CMake Policy". This particular policy defines the behavior of `find_package` with respect to our `*_ROOT` variables. I linked the formal description of the policy above.

gonuke:
I’m now thoroughly confused... I don’t see `MOAB_DIR` in the `MOABConfig.cmake` files in v5.1.0

ljacobson64:
We're talking about the name of the cmake variable we tell users to set when configuring DAGMC so `FindMOAB.cmake` (in DAGMC) can find MOAB.

gonuke:
@pshriwise said: "The only problem there is that they use the variable MOAB_DIR in their MOABConfig.cmake" and I was trying to understand where that was defined and how it affected us

pshriwise:
Hrmmm you're right @gonuke! It's gone. I wonder if I missed that being removed after our discussion about this in the MOAB CMake PR. Either way this makes life a little simpler, we can just go back to using `MOAB_DIR`, yes?

pshriwise:
@ljacobson64 Are you ok with going back to `MOAB_DIR`? I think we can get this merged once that is complete.

ljacobson64:
Yes I'm fine with that. I'll make the change

ljacobson64:
@pshriwise @gonuke any last requests before we merge?

ljacobson64:
Should we maybe bump the DAGMC version number to 3.1 with this PR?

pshriwise:
Just one - that I get one more hour to try it out :smile:

gonuke:
Yes - let's bump the version number and @pshriwise can have the final say on merging

pshriwise:
Thanks @ljacobson64!

PR #579, opened 2018-11-28
Allow cmake 3.10+ when configuring MPI DAG-MCNP6

ljacobson64:
We use the variable `MPI_Fortran_INCLUDE_PATH` when configuring MPI-enabled DAG-MCNP6. However, starting with CMake 3.10, this variable is no longer set. This PR removes the reliance on this variable, opting instead to use `MPI_Fortran_COMPILER`.


src/mcnp/mcnp6/CMakeLists.txt
```diff
@@ -199,12 +199,8 @@ if (BUILD_MCNP_MPI)
...
   endif ()
   if (MCNP_VERSION STREQUAL "611")
-    list(GET MPI_Fortran_INCLUDE_PATH 0 ENV_MPI_ROOT)
-    get_filename_component(temp "${ENV_MPI_ROOT}" NAME)
-    if (${temp} STREQUAL include OR ${temp} STREQUAL lib)
-      get_filename_component(ENV_MPI_ROOT "${ENV_MPI_ROOT}" DIRECTORY)
-    endif ()
-    get_filename_component(ENV_MPI_ROOT "${ENV_MPI_ROOT}" ABSOLUTE)
+    get_filename_component(ENV_MPI_ROOT "${MPI_Fortran_COMPILER}" DIRECTORY)
+    get_filename_component(ENV_MPI_ROOT "${ENV_MPI_ROOT}" DIRECTORY)
```
pshriwise, line 203: @ljacobson64 what is the purpose of line 203?

ljacobson64: `ENV_MPI_ROOT` is supposed to be the path to the location of the MPI package we're using. So something like `/home/lucas/openmpi`. `MPI_Fortran_COMPILER` is an automatically set variable that gives the path to the fortran compiler, i.e. `/home/lucas/openmpi/bin/mpifort`. So the reason we have 2 lines here is that each one goes down 1 directory level so that the end state of `ENV_MPI_ROOT` is `/home/lucas/openmpi`.

Note that the only purpose of most of the `ENV_X` variables, including `ENV_MPI_ROOT`, is to fill in the details that get printed when you run `mcnp6 -v` with MCNP versions 6.1.1 or 6.2.

ljacobson64: So, the way this is set up does depend on the fortran compiler being exactly 2 levels below the base MPI path. This is the way OpenMPI and MPICH both operate by default. I'm not sure what happens if you try to use system OpenMPI... I should try that. But regardless, the worst thing that would happen if this doesn't work correctly is that a wrong path will be printed when you run `mcnp6 -v`.

pshriwise: Right on. Thanks for the background!

I'm guessing it's safe to assume that the `MPI_Fortran_COMPILER` is always an absolute path once the executable's location has been resolved by CMake. 

Looking at my system install of `mpifort` it looks like the the "2 levels up" assumption will hold, but it might be nice to give it a shot if it isn't too difficult. If it is, this is probably ok the way it is.

ljacobson64: Yeah it just sets `ENV_MPI_ROOT` to `/usr` if you use system `mpifort`

pshriwise:
Looks good to me @ljacobson64! I’ll merge at the end of the day today if there are no other comments.

pshriwise:
Thanks @ljacobson64!

PR #580, opened 2018-12-04
Some small fixes to include directories and installed headers.

pshriwise:
A couple small changes to allow the UWUW to be used outside of the DAGMC project after installation.

PR #583, opened 2018-12-13
Adding include to fix the fluka build.

pshriwise:
@ljacobson64 alerted me to the fact that #580 broke the fludag build. This should take care of that problem.


gonuke:
Thanks @pshriwise - I'll go ahead and merge this since it seems logical and consistent and CI is currently unreliable

PR #584, opened 2018-12-13
Do not build against moab-master on CI

ljacobson64:
The purpose of this PR is to prevent the CI from failing for the next month or so until moab 5.1 is released.


gonuke:
Thanks @ljacobson64  - this is a good stop-gap. We should add an issue to explore reactivating this for testing how changes to MOAB affect DAGMC.

PR #585, opened 2018-12-14
Pyne source merge

ljacobson64:
This PR merges the pyne mesh source routine functionality into `develop`. It can be enabled with `-DMCNP_BUILD_PYNE_SOURCE`. This removes the need for the `pyne-source` branch to exist.


src/mcnp/mcnp5/pyne_source/measure.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+#include <math.h>
+
+#include "measure.h"
+
+class CartVect {
+ private:
+  double coords[3];
+
+ public:
```
pshriwise, line 34: Is this the same as the `CartVect` from MOAB?

gonuke: Probably - this code is lifted from MOAB I think

ljacobson64:
@gonuke While `measure` is indeed in MOAB, it is not included in any of the MOAB libraries. It looks like it's only used to build the `mbsize` executable.

So if we wanted to remove `measure.cpp` and `measure.h` here and use MOAB's instead, we'd have to modify moab to make it put it inside one of its libraries.

I think we should skip that and just use our own `measure` files.

gonuke:
Thanks for checking in to the measure methods @ljacobson64

PR #586, opened 2018-12-19
Make HDF5 path to material library a runtime options

gonuke:
The material library generated by PyNE can appear in different places inside an HDF5 file.  This allows the user to specify the HDF5 path to find such a material library, with the default location being "/materials".


ljacobson64:
Perhaps this could use a unit test?

kkiesling:
@ljacobson64 - not sure if it needs it's own tests, but I had previously updated all the tests so that this would work, but now I don't see the changes in this PR.. (see [here](https://github.com/gonuke/DAGMC/pull/2))

bam241:
LGTM! let's go forward and add unit test later if we believe we need some...

bam241:
thx @gonuke

PR #590, opened 2018-12-20
[WIP] H5m explorer -- looking for feedback...

bam241:
This script is based on `pyMOAB` to query the information from an `h5m` file and  build two PandaDataFrames:
- Material Table with material list and assignments
- Volume Table with volume list and corresponding materials assignments

Later I will add a Surface Table (and a Vertex Table?) and maybe the surface information in the Volume Table

This is completely not done. (But I am not sure to do the h5m queries the right way... )

### DIFF
--- a/misc/h5m_explorer.py
+++ b/misc/h5m_explorer.py
@@ -9,7 +9,6 @@ from pymoab import types
 from pymoab.tag import Tag
 
 
-
 class h5mexplorer:
 
     def __init__(self, filename):
@@ -20,17 +19,15 @@ class h5mexplorer:
 
         self.root_set = self.mb.get_root_set()
         self.handles = self.mb.get_entities_by_handle(self.root_set)
-        
+
         self.materials = BuildMaterialTable()
         self.volumes = BuildVolumeTable()
 
-
     def GetCATEGORYHandle(self, category):
-        category_tg_name=self.mb.tag_get_handle(
+        category_tg_name = self.mb.tag_get_handle(
             "CATEGORY", 32, types.MB_TYPE_OPAQUE, False)
         return self.mb.get_entities_by_type_and_tag(self.root_set, types.MBENTITYSET,
-                category_tg_name, np.array([category]))
-
+                                                    category_tg_name, np.array([category]))
 
     def BuildMaterialTable(self):
         mat_handle = []
@@ -47,7 +44,7 @@ class h5mexplorer:
         for handle in group_handle:
             try:
                 name = self.mb.tag_get_data(name_tg_name, handle)[
-                                            0][0].decode('utf-8')
+                    0][0].decode('utf-8')
                 if name[0:3] == "mat":
                     mat_handle.append(handle)
                     mat_name.append(name)
@@ -57,13 +54,14 @@ class h5mexplorer:
                         assignent.append(vol_hld)
                     mat_assignment.append(assignent)
 
-                    mat_id.append(self.mb.tag_get_data(id_tg_name, handle)[0][0])
+                    mat_id.append(self.mb.tag_get_data(
+                        id_tg_name, handle)[0][0])
             except RuntimeError:
                 continue
 
         index = range(0, len(mat_handle))
         columns = ('ID', 'Handle', 'Name', 'Assignment')
-        mat_pdf = pd.DataFrame( columns=columns)
+        mat_pdf = pd.DataFrame(columns=columns)
         for i in range(len(mat_handle)):
             for assigment in mat_assignment[i]:
                 row = [mat_id[i], mat_handle[i], mat_name[i], assigment]
@@ -76,33 +74,32 @@ class h5mexplorer:
         vol_mat_names = []
         vol_mat_handles = []
         vol_surface_handle = []
-        
+
         vol_handle = self.GetCATEGORYHandle("Volume")
-        
+
         tg_name = self.mb.tag_get_handle(
             "GLOBAL_ID", 1, types.MB_TYPE_INTEGER, False)
-        
+
         for i in vol_handle:
             try:
                 vol_id.append(self.mb.tag_get_data(tg_name, i)[0][0])
-                
+
                 vol_mat_handles.append(self.GetMaterialHandleInVolume(i))
                 vol_mat_names.append(self.GetMaterialNameInVolume(i))
-                
+
             except RuntimeError:
                 continue
-        
+
         index = range(0, len(vol_handle))
         columns = ('ID', 'Handle', 'MaterialName', 'MaterialHandle')
-        vol_pdf = pd.DataFrame( columns=columns)
+        vol_pdf = pd.DataFrame(columns=columns)
         for i in range(len(vol_handle)):
             for j in range(len(vol_mat_names[i])):
                 row = [vol_id[i], vol_handle[i], vol_mat_names[i][j],
-                        vol_mat_handles[i][j]]
+                       vol_mat_handles[i][j]]
                 vol_pdf.loc[len(vol_pdf)] = row
 
         return vol_pdf
-
 
     def GetMaterialHandleInVolume(self, volume_handle):
         df_vol = self.materials[self.materials['Assignment'] == volume_handle]
@@ -112,8 +109,7 @@ class h5mexplorer:
         elif len(df_vol) > 1:
             print("!!WARNING!! Multiple material assignements for a single Volume")
         return df_vol['Handle'].tolist()
-    
-    
+
     def GetMaterialNameInVolume(self, volume_handle):
         df_vol = self.materials[self.materials['Assignment'] == volume_handle]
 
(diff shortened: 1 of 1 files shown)


bam241:
I am also looking for feedback about what kind of information to put in the Table (we might want to avoid some redundancy - i.e: mat assignment are present in both tables for the moment)

bohmt:
For the volume table, I suggest that for each volume entity the table contains:
volume of the entity (units=cm3), material name, material mass density (units=g/cm3)
I don't think material atom density (units=nuclei/(barn cm)) is supported by our workflow anymore but that should be included if it is.
For the material table, I suggest that for each material, the table contains:
material mass density, list of volumes that contain that material (if not too long)

misc/h5m_explorer.py
```diff
@@ -0,0 +1,173 @@
...
+        return mat_pdf
+
+    def BuildVolumeTable(self):
+        """
+        Build the Volume Table containing the list of volume, their ID, handle,
+        material name and handle.
+
+        Returns
+        -------
+        PandaDataFrame
```
pshriwise: —> “Pandas DataFrame”

misc/h5m_explorer.py
```diff
@@ -0,0 +1,173 @@
...
+
+        return vol_pdf
+
+    def GetMaterialHandleInVolume(self, volume_handle):
+        """
+        Retrieve the Material Handle for a volume
+
+        Parameters
+        ----------
+        volume_handle : MOAB EntityHandle (long)
```
pshriwise: I’d remove `long` here. It is no longer supported after Python 3.5.

misc/h5m_explorer.py
```diff
@@ -0,0 +1,173 @@
...
+        self.h5m_file = filename
+        self.mb.load_file(filename)
+
+        self.root_set = self.mb.get_root_set()
+        self.handles = self.mb.get_entities_by_handle(self.root_set)
+
+        self.materials = BuildMaterialTable()
+        self.volumes = BuildVolumeTable()
+
+    def GetCATEGORYHandle(self, category):
```
pshriwise: Because this returns a list of handles, maybe we can call it `GetCATEGORYHandles`.

misc/h5m_explorer.py
```diff
@@ -0,0 +1,173 @@
...
+        category : string
+            Value of the category tag to select
+
+        Returns
+        -------
+            MOAB Range of EntityHandles
+
+        """
+        category_tg_name = self.mb.tag_get_handle(
+            "CATEGORY", 32, types.MB_TYPE_OPAQUE, False)
```
pshriwise: PyMOAB has some of this conventional tag information in its `types` module (tag names, sizes etc.). Though it’s possible this info isn’t formatted correctly if you’re using an older version of PyMOAB.

misc/h5m_explorer.py
```diff
@@ -0,0 +1,173 @@
...
+        -------
+        PandaDataFrame
+        """
+        mat_handle = []
+        mat_name = []
+        mat_assignment = []
+        mat_id = []
+
+        group_handle = self.GetCATEGORYHandle("Group")
+        name_tg_name = self.mb.tag_get_handle(
```
pshriwise, line 62: The 4th argument to `tag_get_handle` is specifying a storage type for the tag (either sparse or dense - these are in the `types` module as well). I’m not sure how it’s interpreting the `False` value being passed here. Apparently it works alright for this purpose, but it might have unintended results down the line. 

My guess is that the intent is to indicate that this tag should not be created if it doesn’t exist in the MOAB instance. If that’s the case, it is set to `False` by default so this value shouldn’t be needed.

misc/h5m_explorer.py
```diff
@@ -0,0 +1,173 @@
...
+            "NAME", 32, types.MB_TYPE_OPAQUE, False)
+        id_tg_name = self.mb.tag_get_handle(
+            "GLOBAL_ID", 1, types.MB_TYPE_INTEGER, False)
+
+        for handle in group_handle:
+            try:
+                name = self.mb.tag_get_data(name_tg_name, handle)[
+                    0][0].decode('utf-8')
```
pshriwise: Using `flat=True` in the `tag_get_data` call here will help to avoid one of the `[0]`s here. This might be helpful in a couple other locations as well.

pshriwise (commented):
Just few comments on PyMOAB usage.

misc/h5m_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+        """ Constructor """
+
+        self.mb = core.Core()
+        self.h5m_file = filename
+        self.mb.load_file(filename)
+
+        self.root_set = self.mb.get_root_set()
+        self.handles = self.mb.get_entities_by_handle(self.root_set)
+
+        self.materials = BuildMaterialTable()
```
pshriwise, line 24: Should these two lines be `self.Build*`?

misc/h5m_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+        -------
+            Pandas DataFrame
+        """
+        mat_handle = []
+        mat_name = []
+        mat_assignment = []
+        mat_id = []
+
+        group_handle = self.GetCATEGORYHandles("Group")
+        name_tg_name = self.mb.tag_get_handle(
```
pshriwise, line 62: Calls to tag_get_handle should either only use a name (1st argument) or an entire tag specification (name, length, type, storage_type). These calls need to either include a storage type or only a name should be used.

misc/h5m_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+                                                    category_tg_name, np.array([category]))
+
+    def BuildMaterialTable(self):
+        """
+        Build the Material Table containing the list of material and one line
+        per assignment.
+
+        Returns
+        -------
+            Pandas DataFrame
```
stompsjo, line 54:
```suggestion
            mat_pdf: Pandas DataFrame
```

Since a variable is returned, it might be useful to have that variables name in the function docstring.

misc/h5m_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+        return mat_pdf
+
+    def BuildVolumeTable(self):
+        """
+        Build the Volume Table containing the list of volume, their ID, handle,
+        material name and handle.
+
+        Returns
+        -------
+        Pandas DataFrame
```
stompsjo, line 101:
```suggestion
        vol_pdf: Pandas DataFrame
```

Again it might be useful to see the variable name in the function docstring.

misc/h5m_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+        -------
+        list of material names
+        """
+        df_vol = self.materials[self.materials['Assignment'] == volume_handle]
+
+        if len(df_vol) == 0:
+            return [None]
+        elif len(df_vol) > 1:
+            print("!!WARNING!! Multiple material assignements for a single Volume")
+        return df_vol['Name'].tolist()
```
stompsjo, line 172: As above, is it okay for this function to return something if the warning is triggered?

misc/h5m_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+        -------
+        list of material handles
+        """
+        df_vol = self.materials[self.materials['Assignment'] == volume_handle]
+
+        if len(df_vol) == 0:
+            return [None]
+        elif len(df_vol) > 1:
+            print("!!WARNING!! Multiple material assignements for a single Volume")
+        return df_vol['Handle'].tolist()
```
stompsjo, line 152: Should this function return something if elift len(df_vol) > 1: is executed?

misc/h5m_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+        """
+        category_tg_name = self.mb.tag_get_handle(
+            "CATEGORY", 32, types.MB_TYPE_OPAQUE)
+        return self.mb.get_entities_by_type_and_tag(self.root_set, types.MBENTITYSET,
+                                                    category_tg_name, np.array([category]))
+
+    def BuildMaterialTable(self):
+        """
+        Build the Material Table containing the list of material and one line
+        per assignment.
```
stompsjo, line 50: Should this include a list of explicit material parameters added to the table (e.g. handle, id, name)? See docstring for `BuildVolumeTable`.

misc/h5m_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+        self.root_set = self.mb.get_root_set()
+        self.handles = self.mb.get_entities_by_handle(self.root_set)
+
+        self.materials = BuildMaterialTable()
+        self.volumes = BuildVolumeTable()
+
+    def GetCATEGORYHandles(self, category):
+        """
+        Get the list of handle with the CATEGORY TAG matching the user
+        spacification
```
stompsjo, line 30:
```suggestion
        specification
```

misc/h5m_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+
+import pandas as pd
+import numpy as np
+
+from pymoab import core
+from pymoab import types
+from pymoab.tag import Tag
+
+
+class h5mexplorer:
```
stompsjo, line 12: The class could use its own docstring summary.

stompsjo (approved):
Great job with this class!

PR #591, opened 2018-12-20
MCNP6 pyne mesh source

ljacobson64:
Adds pyne mesh source capability for DAG-MCNP6. @Shimwell @makeclean can you test this?


ljacobson64:
Closing this. I will submit a new PR that add this functionality in what I think is a better way.

PR #594, opened 2019-01-15
Update amalgamated pyne

pshriwise:
## Description
This PR updates the amalgamated PyNE source code in DAGMC to include methods for writing PyNE materials/material libraries to an OpenMC format.

## Motivation and Context
To support the UWUW workflow for DAGMC problems in OpenMC.

## Changes
Several methods are added to the PyNE material and nucname classes to support writing of OpenMC xml elements.

Addition of the C++11 standard to the DAGMC project.

## Other Information
Should we add C++11 for all of DAGMC at this point? Is there a reason not to?

If we don't want to use it globally in the project, can we use this standard in our compilation of `libpyne`?

### DIFF
--- a/cmake/DAGMC_macros.cmake
+++ b/cmake/DAGMC_macros.cmake
@@ -88,6 +88,8 @@ endmacro ()
 macro (dagmc_setup_flags)
   message("")
 
+  set(CMAKE_CXX_STANDARD 11)
+
   if (BUILD_PIC)
     set(CMAKE_POSITION_INDEPENDENT_CODE ON)
   endif ()
--- a/news/PR-0594.rst
+++ b/news/PR-0594.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+* OpenMC material writing capability for UWUW workflow via updates to the
+  amalgamated PyNE source.
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/src/pyne/CMakeLists.txt
+++ b/src/pyne/CMakeLists.txt
@@ -6,6 +6,8 @@ set(PUB_HEADERS pyne.h)
 set(LINK_LIBS)
 set(LINK_LIBS_EXTERN_NAMES HDF5_LIBRARIES)
 
+set(CMAKE_CXX_FLAGS_RELEASE "-std=c++11")
+
 if (CMAKE_Fortran_COMPILER_ID STREQUAL "Intel")
   set(CMAKE_CXX_FLAGS_RELEASE        "-O0")
   set(CMAKE_CXX_FLAGS_DEBUG          "-O0 -g")
--- a/src/pyne/CMakeLists.txt
+++ b/src/pyne/CMakeLists.txt
@@ -6,8 +6,6 @@ set(PUB_HEADERS pyne.h)
 set(LINK_LIBS)
 set(LINK_LIBS_EXTERN_NAMES HDF5_LIBRARIES)
 
-set(CMAKE_CXX_FLAGS_RELEASE "-std=c++11")
-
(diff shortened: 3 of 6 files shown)


ljacobson64:
Is it our intent to keep the amalgamated pyne included with DAGMC more or less up to date with the main pyne repo?

pshriwise:
Good question. I'm not sure. This is sort of a critical update for the motivation listed above rather than a full update of the PyNE source.

gonuke:
One of the reasons for including amalgamated PyNE is that we can update it as often (or as rarely) as we like.  Thus, we only update when something in DAGMC needs a ne feature from PyNE

ljacobson64:
But in the process, we're missing out on bug fixes from upstream. And more importantly, any time we fix a bug in our amalgamated pyne, it doesn't get propagated upstream (see [here](https://github.com/svalinn/DAGMC/commit/9b80a04cdb2c36cd48bceec3d86d9e8c70d575db#diff-feb96d6c87eeec2ff119a53848895636), [here](https://github.com/svalinn/DAGMC/commit/a5ac49de082ea6f7bfb399d2faf78f0bac4aad2f#diff-feb96d6c87eeec2ff119a53848895636), and [here](https://github.com/svalinn/DAGMC/commit/6908108e0b568dd2d1142516d5031be2dd73fdeb#diff-feb96d6c87eeec2ff119a53848895636)). This makes it difficult to maintain.

I think it would be a good idea to create a policy where if we need to make a change to amalgamated pyne, it should be done by submitting a PR to pyne, re-amalgamating, and then updating it here. I will volunteer to backport the changes we've made to amalgamated pyne and submit a PR to pyne.

That being said, if this PR is critical path, then I'm fine with merging it and worrying about this issue later.

pshriwise:
@ljacobson64 agreed! Any bug fix/change that is discovered here should go into PyNE. Changes to PyNE in the meantime might cause us to bring in more features than are necessary for DAGMC, but I think keeping the code as in sync as possible (AISAC? ;) ) is important. 

This PR is in fact on a critical path for some work that I and @makeclean are interested in, so I definitely would like to get it reviewed and merged sooner rather than later, especially given that we're developing this policy just now.

ljacobson64:
Sounds good to me. We probably don't need to review the added c++ code, since I'm assuming it was already reviewed when it went into pyne, right?

The only other change was adding the c++11 standard, which I think is fine. The only place I can think of where it's a hassle to get access to c++11 is on the HPC, but we already use a custom-built compiler there anyways. (Not sure about HTC though.)

ljacobson64 (approved):

gonuke:
Is this waiting for the backport PR in PyNE?  Do you want that to be merged there first and then build another amalgamated PyNE for here?

On a related note, we probably want to never change amalgamated PyNE on this end and always PR something to PyNE and re-amalgamate if we need changes.

ljacobson64:
>Is this waiting for the backport PR in PyNE? Do you want that to be merged there first and then build another amalgamated PyNE for here?

Eventually we should merge that PR and re-amalgamate, but I'm fine with this PR being merged before then. @pshriwise says this is critical path for him.

>On a related note, we probably want to never change amalgamated PyNE on this end and always PR something to PyNE and re-amalgamate if we need changes.

Wholeheartedly agree. There is one snag though -- there is a function in pyne where we currently rely on turning it into a dummy function in order to get pyne.cpp to compile.  Not sure the best way to make sure the dummy function doesn't get overwritten every time we re-amalgamate.

ljacobson64:
But if you want to wait until the pyne PR gets merged and then do it the right way, I'm fine with that too.

gonuke:
I'm fine to merge this - I just want to make sure it's not rolling back anything that's in the backport PR from PyNE.  If this amalgamation is tweaked appropriately, then let's go for it.

gonuke:
Thanks @pshriwise

PR #595, opened 2019-01-16
Update amalgamated pyne to match the pyne repo

ljacobson64:
This PR updates the DAGMC version of amalgamated pyne to match the main pyne repo. It also adds a script to automate this process.

Copying what I wrote in #594:

>I think it would be a good idea to create a policy where if we need to make a change to amalgamated pyne, it should be done by submitting a PR to pyne, re-amalgamating, and then updating it here.

>I will volunteer to backport the changes we've made to amalgamated pyne and submit a PR to pyne. 

Update: this PR has been submitted and merged. Thanks for the quick merge Paul!

>There is one snag though -- there is a function in pyne where we currently rely on turning it into a dummy function in order to get pyne.cpp to compile. Not sure the best way to make sure the dummy function doesn't get overwritten every time we re-amalgamate.

Update: I wrote a short script which modifies these functions to be the dummy functions.


gonuke:
I see which functions are a problem & understand why.  This is something else we should address on the PyNE end eventually.

gonuke:
LGTM - thanks @ljacobson64

PR #574, opened 2018-11-09
Pyne source

shimwell:
There is now an MCNP6 source.F90 which was converted from the MCNP5 source.F90

Also there are some CMakeLists.txt for both mcnp5 and mcnp6 which will allow these source.F90 files to be included in the build.

PR #604, opened 2019-02-18
Add source_sampling to amalgamated pyne

ljacobson64:
This PR adds the `measure` and `source_sampling` routines to amalgamated pyne. This removes the need for us to build them separately when building DAG-MCNP with pyne mesh source sampling support.

This PR also adds a version of `source.F90` that works for MCNP6.

This PR should not be approved until [pyne PR 1098](https://github.com/pyne/pyne/pull/1098) is merged into pyne to avoid having our amalgamated pyne become out of sync with pyne/develop.


ljacobson64:
The added bonus for this is that whenever changes are made to `measure` or `source_sampling` in pyne (such as @zxkjack123's [PR 1099](https://github.com/pyne/pyne/pull/1099) in pyne), all we have to do is update amalgamated pyne to get these changes in DAGMC.

ljacobson64:
I'll explain the purpose of these latest commits. `libpyne_dagmc` was originally only required to be linked to HDF5, not MOAB, but with the addition of `measure`, it is now required to be linked to MOAB. For whatever reason, this caused the position of the libraries `blas`, `lapack`, `quadmath`, and `gfortran` to start appearing in the wrong place in the linking command for executables, which caused them to not be found, which caused the CI to fail.

Originally, those 4 libraries were listed as required libraries for `libdagmc`. I changed it so those libraries are now part of the `MOAB_LIBRARIES` variable, and this made them appear in the correct order again and made the CI pass.

gonuke:
(similar to my comment in #574 ) should we include the `Source.F90` files in our repo or just pull them in from PyNE?

Maybe that can be part of an amalgamated PyNE as well?

ljacobson64:
I think pulling MCNP5/MCNP6 `source.F90` files directly from pyne could work. But I don't think they could be part of an amalgamated pyne because I think we still want to pass files called `source.F90` to MCNP, not some sort of amalgamated `pyne.F90`. Also, that could get hairy since MCNP5 and MCNP6 need different versions of `source.F90`.

So perhaps we could alter the `update_pyne.sh` script to also copy over the `source.F90` files for MCNP5 and MCNP6 from pyne, but not rename them.

shimwell:
Keen to use this nice feature, thanks for the upgrades @ljacobson64

ljacobson64:
@gonuke are you looking for this PR to contain an automated way to pull the source.F90 files directly from pyne, or can that be a separate PR?

If this PR is fine as is, then I think the only thing that needs to happen before merge is that [this PR](https://github.com/pyne/pyne/pull/1098) in pyne is merged first.

gonuke (approved):
Thanks @ljacobson64 - let’s leave the `source.F90` issue to another PR

PR #605, opened 2019-02-20
Throw fatal error for invalid cmake combinations

ljacobson64:
@pshriwise noticed that when he tried to turn `BUILD_STATIC_LIBS` off while turning `BUILD_STATIC_EXE` on, the build proceeded smoothly despite the expectation that it would fail. It turns out that what was happening is that cmake was silently changing `BUILD_STATIC_EXE` to off if it noticed that `BUILD_STATIC_LIBS` was off. We decided that this should probably be a fatal error instead.


pshriwise:
Waiting on CI, but looks good to me as long as that passes!

pshriwise (commented):
CI looks good and the changes capture a combination of options we don't want to allow. Thanks @ljacobson64!

PR #610, opened 2019-02-21
Update Docker/CI

ljacobson64:
* The `ASTYLE_ONLY` Travis variable has been replaced with a `HOUSEKEEPING_ONLY` variable. If this variable is on, DAGMC will not be built and it will instead only perform 3 housekeeping checks:
    * News file: the CI will fail if a news file with the correct filename is not included.
    * Astyle: the version of astyle we use on the CI has been upgraded to 3.1. This is the version that is default on Ubuntu 18.04.
    * Documentation: the CI will now attempt to build the DAGMC documentation and will fail if it finds any errors or warnings.
* A Dockerfile has been added for Ubuntu 18.04.
* The new build matrix for the non-housekeeping run is 2x2x2:
    * Ubuntu 16.04 vs. 18.04
    * gcc vs. clang
        * gcc-5.3 on 16.04; gcc-7.3 on 18.04
        * clang-3.8 on 16.04; clang-6.0 on 18.04
    * MOAB 5.1.0 vs. master
* The builds that use MOAB master are allowed to fail without the entire CI failing. The CI will show as having passed once the housekeeping build and the four MOAB 5.1.0 builds have passed.
* MOAB 5.1.0 is now included in the docker image so it does not need to be built every time the CI is run.
    * This is to save time, since we expect that previous versions of MOAB will not change. If it does change, we can update the Docker images.
    * MOAB master is still built every time it is needed.
* MOAB is now built with pymoab support. This is for future-proofing in case DAGMC ever needs access to this functionality.
* MOAB is now built against both custom-built HDF5 (1.10.4, up from 1.8.13) and against system HDF5.
    * The MOAB built against system HDF5 is currently unused, however, as there is currently a bug that makes it so DAGMC cannot build static executables if using system HDF5. If/when this bug is fixed, then building DAGMC with system HDF5 can be added to the build matrix.
* Geant4 has been upgraded to version 10.5.
* Building the documentation will throw an error if it encounters any warnings or errors. The previous warnings and errors that were occurring have been fixed.
* The `make_watertight_regression_tests` should now be run if the CI is not doing a PR build.
    * I believe this was broken for an undetermined amount of time; I do not believe they were ever getting run regardless of whether the CI was doing a PR build or not. This is because intrinsic Travis variables like `$TRAVIS_PULL_REQUEST` are only available to `.travis.yml`; if they are needed in other scripts, they need to be passed manually, and this was not happening before.


ljacobson64:
From the housekeeping run: 
```
ERROR: News file news/PR-0610.rst not found. Please create a news file.
```

docker/Dockerfile-Ubuntu-16.04
```diff
@@ -0,0 +1,58 @@
+FROM ubuntu:16.04
```
bam241: As both your docker images look identical you could use the ARG command, which allows to set a default,  like:
```
ARG UBUNTU_VERSION=16.04
FROM ubuntu:$UBUNTU_VERSION
```
and to specify it when building the container for `18.04`:
`docker build -t container_tag --build-arg UBUNTU_VERSION="18.04"`
[[see](https://www.jeffgeerling.com/blog/2017/use-arg-dockerfile-dynamic-image-specification)]

docker/Dockerfile-Ubuntu-16.04
```diff
@@ -0,0 +1,58 @@
...
+RUN apt-get -y install git
+RUN apt-get -y install libblas-dev
+RUN apt-get -y install libhdf5-dev
+RUN apt-get -y install liblapack-dev
+RUN apt-get -y install libpython2.7-dev
+RUN apt-get -y install libtool
+RUN apt-get -y install python-numpy
+RUN apt-get -y install python-pip
+RUN apt-get -y install python-setuptools
+RUN apt-get -y install wget
```
bam241: Using a single command will allow to build a lighter docker image (with less single process to load),
something like:
RUN apt-get -y install \
  autoconf \
  clang \
 ...

docker/build_moab.sh
```diff
@@ -1,34 +1,39 @@
...
-../src/configure --enable-dagmc \
-                 --disable-ahf \
+../src/configure --enable-pymoab \
                  --enable-shared \
                  --enable-optimize \
                  --disable-debug \
-                 --with-hdf5=${install_dir}/hdf5-${hdf5_version} \
-                 --prefix=${install_dir}/moab-${moab_version} \
+                 --with-hdf5=${hdf5_install_dir} \
+                 --prefix=${moab_install_dir} \
```
bam241, line 26: why not using cmake for MOAB (comment from @gonuke in a similar PR)?

ljacobson64:
And now...
```
News file news/PR-0610.rst found!
```

docker/Dockerfile-Ubuntu-16.04
```diff
@@ -0,0 +1,58 @@
...
+# Get some pip packages
+RUN pip install cython
+RUN pip install sphinx
+
+# Copy the environment variable script
+RUN mkdir -p /root/etc/
+COPY env.sh /root/etc/
+
+# Build HDF5
+COPY build_hdf5.sh /root/etc/
```
bam241: can't the bash macros be run form their home folder ?

ljacobson64:
Note that the new docker images are on my personal docker account. They would need to be pushed to the CNERG docker account at some point.

docker/build_geant4.sh
```diff
@@ -1,24 +1,27 @@
 #!/bin/bash
 
-# $1: compiler (gcc-4.8, gcc-5, gcc-6, gcc-7, clang-4.0, clang-5.0)
+# $1: compiler (gcc, clang)
```
bam241: Maybe an `if` statement could be nice to ensure than $1 is either `gcc` or `clang` ?

docker/build_hdf5.sh
```diff
@@ -1,23 +1,32 @@
 #!/bin/bash
 
-# $1: compiler (gcc-4.8, gcc-5, gcc-6, gcc-7, clang-4.0, clang-5.0)
+# $1: compiler (gcc, clang)
+# $2: hdf5 version (system, 1.10.4)
```
bam241: Same here, maybe a test to check the values of the arguments match the requirements

docker/build_moab.sh
```diff
@@ -1,34 +1,39 @@
 #!/bin/bash
 
-# $1: compiler (gcc-4.8, gcc-5, gcc-6, gcc-7, clang-4.0, clang-5.0)
-# $2: moab version (5.1.0, master)
+# $1: compiler (gcc, clang)
+# $2: hdf5 version (system, 1.10.4)
+# $3: moab version (5.1.0, master)
```
bam241: same here test for the arguments ?

.travis.yml
```diff
@@ -5,41 +5,49 @@ notifications:
...
-  - commit_id=$(sudo docker ps -l | tail -n1 | awk '{print $1}')
-  - sudo docker commit $commit_id cnerg/dagmc-ci-ubuntu-16.04
+  - docker_image=ljacobson64/dagmc-ci-ubuntu-${UBUNTU_VERSION}
+  - build_dir=/root/build/${COMPILER}/DAGMC-moab-${MOAB_VERSION}-hdf5-${HDF5_VERSION}
+  # Get the docker image
+  - docker pull ${docker_image}
+  # Make the build directory
+  - docker run ${docker_image} /bin/bash -c "mkdir -p ${build_dir}"
+  - commit_id=$(docker ps -l | tail -n1 | awk '{print $1}')
+  - docker commit ${commit_id} ${docker_image}
```
bam241, line 63: are the commits required at every step ? ?

.travis.yml
```diff
@@ -5,41 +5,49 @@ notifications:
...
 script:
-  - sudo docker run cnerg/dagmc-ci-ubuntu-16.04 /bin/bash -c "cd /root/build/$COMPILER/DAGMC-moab-$MOAB_VERSION/DAGMC;
-                                                                    bash docker/travis.script.sh $ASTYLE_ONLY $COMPILER $MOAB_VERSION $BUILD_GEANT4"
+  # Run unit tests
+  - docker run ${docker_image} /bin/bash -c "if [ \"${HOUSEKEEPING_ONLY}\" == \"true\" ]; then
+                                               bash ${build_dir}/DAGMC/docker/travis.housekeeping.sh \"${COMPILER}\" \"${HDF5_VERSION}\" \"${MOAB_VERSION}\" \"${TRAVIS_REPO_SLUG}\" \"${TRAVIS_PULL_REQUEST}\";
+                                             else
+                                               bash ${build_dir}/DAGMC/docker/travis.tests.sh        \"${COMPILER}\" \"${HDF5_VERSION}\" \"${MOAB_VERSION}\" \"${TRAVIS_REPO_SLUG}\" \"${TRAVIS_PULL_REQUEST}\";
+                                             fi"
```
bam241: It might be more human readable if you split this into 2 travis command, and do the if statement in `travis` instead of in the docker container, something like:
```
# Housekeeping
- if ${HOUSEKEEPING_ONLY} then 
 docker run ${docker_image} /bin/bash -c " travis.housekeeping.sh ...."
 fi
# Run unit tests
-  if ! ${HOUSEKEEPING_ONLY} then 
 docker run ${docker_image} /bin/bash -c " /travis.tests.sh ...."
 fi

**Review by bam241 — COMMENTED** [review]
thx for putting this together @ljacobson64 .

I am not sure if my suggestions for the travis workflow (`housekeeping` vs `unit_test`) are possible or make sense.
Also the test in the different bash script might not be necessary....

File: docker/Dockerfile-Ubuntu-16.04 (hunk 1 above)
Comment (ljacobson64) [reply docker/Dockerfile-Ubuntu-16.04]: The docker images are not identical. 16.04 can't simply apt-get astyle so it has to download the .deb from the internet.

File: docker/Dockerfile-Ubuntu-16.04 (hunk 2 above)
Comment (ljacobson64) [reply docker/Dockerfile-Ubuntu-16.04]: I didn't know that that would increase performance. I can certainly make this all one command.

File: docker/build_moab.sh (hunk 3 above)
Comment (ljacobson64) [reply docker/build_moab.sh:26]: Because then you have to build it twice to get the shared and static libraries. CMake can't build them at the same time, while autotools can.

That being said, it might be a good idea to switch to CMake always though, since I get the feeling the MOAB CMake build is maintained much more than the autotools build.

File: docker/Dockerfile-Ubuntu-16.04 (hunk 4 above)
Comment (ljacobson64) [reply docker/Dockerfile-Ubuntu-16.04]: Not sure what you mean. Do you mean like `COPY build_hdf5.sh ~`?

File: .travis.yml (hunk 8 above)
Comment (ljacobson64) [reply .travis.yml:63]: Committing at every step is something that predates my work on this. I don't know if there is a better way. If there is, I haven't searched hard enough to find it yet.

File: .travis.yml (hunk 9 above)
Comment (ljacobson64) [reply .travis.yml]: Good idea. I will change this. For some reason it didn't occur to me that you could have an `if` statement in `travis` directly.

File: docker/Dockerfile-Ubuntu-16.04 (hunk 4 above)
Comment (bam241) [reply docker/Dockerfile-Ubuntu-16.04]: nop I meant calling them from where they are originally

File: docker/Dockerfile-Ubuntu-16.04 (hunk 1 above)
Comment (bam241) [reply docker/Dockerfile-Ubuntu-16.04]: Would the deb version working for 18.04 ? 

Maybe it would be too complicated... I don't know but I usually prefer to limit the amount of file we have to maintain....

File: docker/Dockerfile-Ubuntu-16.04 (hunk 4 above)
Comment (ljacobson64) [reply docker/Dockerfile-Ubuntu-16.04]: They are originally not in the docker image so I don't know how you would call them from inside the docker image without copying them

Comment (bam241) [reply docker/Dockerfile-Ubuntu-16.04]: Oh I was confused...
They come from the repo but before the repo gets copied into the docker container...
Right?
If so I am not sure there is an other option... 
You could maybe copy all of them in the same time...

File: docker/Dockerfile-Ubuntu-16.04 (hunk 2 above)
Comment (ljacobson64) [reply docker/Dockerfile-Ubuntu-16.04]: This is done.

File: docker/build_geant4.sh (hunk 5 above)
Comment (ljacobson64) [reply docker/build_geant4.sh]: I changed the entire paradigm of how variables are passed into scripts. I decided to `export` the variables `COMPILER`, `HDF5_VERSION`, and `MOAB_VERSION` so they are available to every script. None of the scripts accept command line arguments anymore.

File: .travis.yml (hunk 9 above)
Comment (ljacobson64) [reply .travis.yml]: This is done.

File: docker/build_geant4.sh (hunk 5 above)
Comment (ljacobson64) [reply docker/build_geant4.sh]: This is done in `.travis.yml`:
```
before_install:
  - |
    printf "COMPILER=${COMPILER}
    HDF5_VERSION=${HDF5_VERSION}
    MOAB_VERSION=${MOAB_VERSION}
    TRAVIS_REPO_SLUG=${TRAVIS_REPO_SLUG}
    TRAVIS_PULL_REQUEST=${TRAVIS_PULL_REQUEST}
    docker_env=${docker_env}" > ${travis_env}
```

Then whenever a command needs to be run inside the Docker image...
```
docker run --env-file ${travis_env} ${docker_image} /bin/bash -c "echo hello world"
```

**Comment by ljacobson64** [comment]
capture3

Note how the build is considered to have passed despite the 4 MOAB-master runs not having finished yet.

File: .travis.yml (hunk 8 above)
Comment (ljacobson64) [reply .travis.yml:63]: Regardless, I found a shorter command to get the commit ID:

`commit_id=$(docker ps -lq | tail -n1)`

Comment (bam241) [reply .travis.yml:63]: maybe you could try to comment out those commit lines, se what append ?

File: .travis.yml [hunk 10]
```diff
@@ -5,41 +5,61 @@ notifications:
...
-  # make the build directory
-  - sudo docker run cnerg/dagmc-ci-ubuntu-16.04 /bin/bash -c "mkdir -p /root/build/$COMPILER/DAGMC-moab-$MOAB_VERSION"
-  - commit_id=$(sudo docker ps -l | tail -n1 | awk '{print $1}')
-  - sudo docker commit $commit_id cnerg/dagmc-ci-ubuntu-16.04
-  # copy the repo in the docker instance
-  - sudo docker cp ../DAGMC $commit_id:/root/build/$COMPILER/DAGMC-moab-$MOAB_VERSION
-  - commit_id=$(sudo docker ps -l | tail -n1 | awk '{print $1}')
-  - sudo docker commit $commit_id cnerg/dagmc-ci-ubuntu-16.04
+  # Set some local variables
+  - docker_image=ljacobson64/dagmc-ci-ubuntu-${UBUNTU_VERSION}
```
bam241: this needs to change to `cnerg/damci-ubuntu-${UBUNTU_VERSION}`

(I know you meant to, this is just to avoid forgetting it:))

bam241 (changes requested):

gonuke:
Are the new docker images overwriting old ones, or adding new images?  If the latter, you can just push them now and update this.  (Assuming you have permission...)

Related: should we have a svalinn org on dockerhub?

Can you add MOAB develop just like you’ve done for MOAB master (non-fatal)?  Could it be done only for nightly builds?

docker/Dockerfile-Ubuntu-16.04
gonuke: You could layer the dockerfiles and produce intermediate docker images
* base Ubuntu 16.04 & 18.04 through the astyle install 
* all the common steps in a dockerfile that takes a version argument

(At some point we may just want to stop testing on u16.04)

gonuke (commented):
Not sure about whitespace changes in KDENeighborhoodTest?

docker/Dockerfile-Ubuntu-16.04
ljacobson64: Could also just put an `if` statement in the Dockerfile so if building with 18.04, it gets astyle as usual and if building with 16.04, it downloads the .deb

.travis.yml
ljacobson64: I don't think removing those lines is going to work https://stackoverflow.com/a/19586345

ljacobson64:
Whitespace changes are from updating the version of astyle

ljacobson64:
Speaking of astyle, I overall haven't been very impressed with it... there is another code formatter called `clang-format` that appears to be much more stable and in my opinion produces better-formatted code. `clang-format` is developed by the clang team, while astyle is mainly maintained by just one person.

This would be a separate PR (or it could not even happen at all); I'm just musing.

ljacobson64:
@gonuke The new docker images should overwrite the old ones, yes. I think I do have permission to push to the CNERG dockerhub, but I don't want to push things there until we're just about ready for this to be merged, because ideally they should happen at the same time.

I guess using a svalinn org might be more appropriate than CNERG since this is a project that involves more people than just CNERG. But that can be up to you.

As for building moab develop, it can certainly be done just like it is for moab master. I'm still looking into whether it's possible to have it only be done in the nightlies as opposed to PRs.

ljacobson64:
I got it to only do the housekeeping + moab 5.1 builds when in PR mode, and it's supposed to be set up to add the moab-master builds when in non-PR mode (i.e. push and cron builds), but that functionality can't be tested in this PR, because... it's a PR. So I'll set up a dummy branch and merge this branch into the dummy branch and see what happens then.

ljacobson64:
The dummy branch thing worked. So now it's set up so that it only builds the housekeeping build and the moab-5.1 builds for PRs, and it builds those plus the moab-master and moab-develop builds for non-PR builds. And the moab-master and moab-develop builds are still set up so that them failing won't cause the whole CI run to fail.

gonuke:
I made the svalinn org on dockerhub and you should have permission @ljacobson64

docker/Dockerfile-Ubuntu-16.04
ljacobson64: This has been addressed. There is only 1 dockerfile now.

ljacobson64:
the new docker images have been pushed to the svalinn org

gonuke:
More great infrastructure work! Thanks @ljacobson64

docker/build_dagmc.sh
```diff
@@ -1,34 +1,32 @@
...
+  build_mw_reg_tests=OFF
+fi
 
-mkdir -p ${build_dir}/DAGMC-moab-${moab_version}/bld
-rm -rf ${install_dir}/DAGMC-moab-${moab_version}
-cd ${build_dir}/DAGMC-moab-${moab_version}
+rm -rf ${dagmc_build_dir}/bld ${dagmc_install_dir}
+mkdir -p ${dagmc_build_dir}/bld
+cd ${dagmc_build_dir}
 #git clone https://github.com/svalinn/DAGMC -b develop
```
gonuke, line 16: remove commented lines?  Is there a reason they are here?

docker/build_dagmc.sh
```diff
@@ -1,34 +1,32 @@
...
              -DCMAKE_Fortran_COMPILER=${FC} \
-             -DCMAKE_INSTALL_PREFIX=${install_dir}/DAGMC-moab-${moab_version}
-make -j`grep -c processor /proc/cpuinfo`
+             -DCMAKE_INSTALL_PREFIX=${dagmc_install_dir}
+make -j${jobs}
 make install
-#rm -rf ${build_dir}/DAGMC-moab-${moab_version}
-rm -rf ${build_dir}/DAGMC-moab-${moab_version}/bld
+cd
+#rm -rf ${dagmc_build_dir}
```
gonuke, line 32: remove commented line? or uncomment this line?

gonuke (commented):
Just 2 comments about comments...

gonuke:
If @bam241 approves we can merge this...

bam241:
LGTM!

bam241 (approved):

gonuke:
Thanks @ljacobson64

PR #611, opened 2019-02-23
Adding option to read starting weights from mesh

moatazharb:
This PR updates pyne-source sampling routine to allow for user supplied starting particles weights. The provided starting particles weights are mainly used for error propagation from mesh-based sources to sdr via random sampling in the R2S workflow. 
While this update will mainly be used as part of the thesis research of @moatazharb , it can also be used in cased both the biased source density distribution and weights are supplied by the user in which case this additional mode will be identical to the mode in which a user supplies both the source and biased source distributions.

- updated description of code.
- corrected number of modes as on the idum first entry on mcnp input 0 refers to analog, 1 to uniform, etc.. In a comment line analog was referred to by 1 mistakenly. 
 > source_sampling.cpp
- Added mode "3" to source_sampling.cpp
- changed some conventions; from "USER" to "UBIASED" and "UWEIGHT". The former refers to a User supplied BIASED source density distribution along with the unbiased, while the later refers to a User supplied WEIGHT along with the source density distribution. 

@ljacobson64 @gonuke I know that dagmc on chtc is rebuilt per new merged PRs. This is important to finish the thesis, so ... :)

### DIFF
--- a/news/PR-0611.rst
+++ b/news/PR-0611.rst
@@ -1,0 +1,16 @@
+**Added:** None
+
+**Changed:**
+source_sampling.h
+source_sampling.cpp
+
+* Throw a fatal error if trying to build static executables but not static
+  libraries, or shared executables but not shared libraries
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/src/mcnp/mcnp5/pyne_source/source_sampling.cpp
+++ b/src/mcnp/mcnp5/pyne_source/source_sampling.cpp
@@ -18,7 +18,10 @@ void pyne::sampling_setup_(int* mode) {
       sampler = new pyne::Sampler(filename, src_tag_name, e_bounds, true);
     } else if (*mode == 2) {
       std::string bias_tag_name("biased_source_density");
-      sampler = new pyne::Sampler(filename, src_tag_name, e_bounds, bias_tag_name);
+      sampler = new pyne::Sampler(filename, src_tag_name, e_bounds, bias_tag_name, false);
+    } else if (*mode == 3) {
+      std::string bias_tag_name("weight");
+      sampler = new pyne::Sampler(filename, src_tag_name, e_bounds, bias_tag_name, true);
     }
   }
 }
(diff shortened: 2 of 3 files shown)


ljacobson64:
@moatazharb I think that it would be best if you made these changes in pyne, and then we approved #604 which adds `source_sampling` into our own `pyne.cpp` and `pyne.h`.

That being said, if you need this change on HPC ASAP, I can build this branch for you while we wait for all these PR's to be merged.

zxkjack123:
A new sampler has been added to the source_sampling.cpp.
See this function:
https://github.com/pyne/pyne/blob/7db8a587c460c986e356bfb10bd42d607f00d477/src/source_sampling.cpp#L90

There are two types of sampler exist and used in the source_sampling.cpp. 
https://github.com/pyne/pyne/blob/7db8a587c460c986e356bfb10bd42d607f00d477/src/source_sampling.cpp#L9
- Old one: sampler for only default/voxel R2S (which you are modifying). They are:
```
if (*mode == 0) {
      sampler = new pyne::Sampler(filename, src_tag_name, e_bounds, false);
    } else if (*mode == 1) {
      sampler = new pyne::Sampler(filename, src_tag_name, e_bounds, true);
    } else if (*mode == 2) {
      std::string bias_tag_name ("biased_source_density");
      sampler = new pyne::Sampler(filename, src_tag_name, e_bounds, bias_tag_name);
    }
```
- sampler for both voxel and sub-voxel R2S:
```
else {
      sampler = new pyne::Sampler(filename, tag_names, e_bounds, *mode);
    }
```

Currently, both of these two types of sampler works, and the old samplers are the default choice of voxel mode. However, as the new sampler works for both voxel and sub-voxel R2S, the old sampler functions could be removed in the future.

It's better to make changes based on the new sampler. By the way, the mode number 3 has been used to represent the sub-voxel analog mode. I suggest you to change the mode number, such as 6.

moatazharb:
@ljacobson64 I guess that would be a good compromise for now since I'll need it on HPC to perform few runs in the coming days. It'll be great if you can build that on HPC. I just want to make sure that dag-mcnp5 is built also as that's what I used for all my calculations. 

@zxkjack123 Thanks for referencing your PR. I'll make a PR based on your changes once yours is approved. That just might take me sometime since now I'm pretty occupied with finishing my thesis.

ljacobson64:
@moatazharb I built this branch on HPC here: `/home/group/dagmc/opt/gcc-8/DAGMC-moataz`

moatazharb:
Closing this PR. A new PR will be issued in pyne after latest changes to source sampling from @ljacobson64 & @zxkjack123 are merged.

PR #612, opened 2019-03-08
Modification to allow finding of moab on Fedora/Redhat systems where …

makeclean:
## Description
Changed one line in FindMOAB.cmake to allow finding of MOAB libraries in their default named positions on Fedora systems

## Motivation and Context
Build is currently broken for Fedora/RedHat based systems

## Changes
Bug fix

## Behavior
Current behaviour - broken - New Behaviour - fixed

## Other Information
N/A

## News file
Incoming


ljacobson64:
LGTM, thanks @makeclean!

PR #616, opened 2019-03-16
Openmc mat assign by name

pshriwise:
## Description
An addition to documentation on model preparation for OpenMC.

## Motivation and Context
A recent [PR](https://github.com/openmc-dev/openmc/pull/1199) into OpenMC allows for materials to be assigned by name rather than ID. This is going to be the recommended workflow for DAG-OpenMC model generation.

## Changes
Updates to the OpenMC section of the documentation.

## Behavior
No behavior is changed.

### DIFF
--- a/doc/usersguide/codes/openmc.rst
+++ b/doc/usersguide/codes/openmc.rst
@@ -63,12 +63,12 @@ OpenMC's ``materials.xml`` file:
 This method for assigning materials is recommended for use with OpenMC as it
 provides a more verbose description of the material definition and purpose.
 
-This method of assignment also allows an easy transition to the UWUW workflow
-for future models. One can embed a PyNE material library in the DAGMC model at
-any point using ``uwuw_preproc`` without modification to the material
-assignments to obtain a working UWUW model, provided that materials with
-corresponding names and appropriate definitions are in the PyNE material
-library.
+This method of assignment also allows for straighforward model conversion to the
+UWUW workflow without material assignment modification, re-faceting, or
+re-sealing. To do this one can simply embed a PyNE material library in the
+original .h5m file at any point using ``uwuw_preproc``. No modification to the
+material assignments is necessary, provided that materials with appropriate
+names and definitions are in the PyNE material library.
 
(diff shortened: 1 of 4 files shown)


src/uwuw/uwuw_preprocessor.cpp
```diff
@@ -152,7 +152,7 @@ void uwuw_preprocessor::process_materials() {
 
     std::string mat_name = dmd->return_property(*s_it, "mat");
     std::string density  = dmd->return_property(*s_it, "rho");
-
+    mat_name = "mat:"+mat_name;
```
gonuke: This seems like a big difference, what am I missing here?

Looks like ASTYLE fail here too 🙂

doc/usersguide/codes/openmc.rst
```diff
@@ -41,6 +41,38 @@ then add them to a group called, "mat:Vacuum":
...
+     <material id="40" name="fuel">
+       <density units="g/cc" value="11" />
+       <nuclide ao="1.0" name="U235" />
+     </material>
+   </materials>
+
+This method for assigning materials is recommended for use with OpenMC as it
+provides a more verbose description of the material definition and purpose.
+
+This method of assignment also allows for straighforward model conversion to the
```
gonuke: What happens if the same material is defined in the `materials.xml` file and the embedded MaterialLibrary?  Or the same material name, but different definition?

gonuke (commented):
Thanks @pshriwise - I think I now understand what you were telling me yesterday....

src/uwuw/uwuw_preprocessor.cpp
pshriwise: Yikes. That was a personal patch, how embarrassing... :grimacing: will remove :smile:

doc/usersguide/codes/openmc.rst
pshriwise: I'll clarify that situation.

gonuke:
Cool feature @pshriwise

PR #617, opened 2019-03-20
Update amalgamated pyne

ljacobson64:
First time updating our version of pyne to match the real pyne using our new `update_pyne.sh` script.

### DIFF
--- a/news/PR-0617.rst
+++ b/news/PR-0617.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:**
+
+* Updated amalgamated pyne
+
+**Deprecated:** None
+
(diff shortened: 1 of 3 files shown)


gonuke:
I'm surprised by how few changes... shouldn't we see the include guards, for example, in measure.h?

ljacobson64:
Sorry, my description here was kind of inaccurate. #604 did include a number of changes. The changes in this PR are only the ones that occurred between the time #604 was submitted and now. The reason the include guards aren't present in this PR is because I already manually added them into #604; they were required to make #604 work.

gonuke:
Thanks @ljacobson64

PR #621, opened 2019-03-22
Make the MW_REG_TEST_MODELS_URL variable available to the docker image

ljacobson64:
This fixes the problem where all the nightly CI builds were failing. [This Travis build](https://travis-ci.org/svalinn/DAGMC/builds/510079202) shows that this PR fixes the issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,7 @@ before_install:
     MOAB_VERSION=${MOAB_VERSION}
     TRAVIS_REPO_SLUG=${TRAVIS_REPO_SLUG}
     TRAVIS_PULL_REQUEST=${TRAVIS_PULL_REQUEST}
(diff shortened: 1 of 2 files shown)


pshriwise (approved):
Looks good @ljacobson64.

pshriwise:
I'll merge on Monday mid-day if there are no other comments.

gonuke:
Thanks @ljacobson64 

For future reference, was there something that you expected to work that didn’t?  Or was it simply a bug in your script from the beginning?

ljacobson64:
As far as I can tell, the make_watertight regression tests were never being run in the past because that environment variable was never being propagated into the docker image, and there was some logic present that said that if that environment variable was empty, don't run the test. My PR caused the make_watertight regression tests to actually attempt to run, and the lack of access to this environment variable inside the docker image led to the tests failing.

PR #632, opened 2019-08-24
Fix error in documentation where cmake was not pointing to the DAGMC source dir as it should

ljacobson64:
Adds the line `ln -s DAGMC src` and changes all instances of `cmake ..` to `cmake ../src`


pshriwise (approved):
These changes look good to me! I'll merge tomorrow afternoon if there are no further comments.

pshriwise:
Thanks @ljacobson64!

PR #634, opened 2019-09-04
Do not include lapack/blas/gfortran/quadmath in MOAB_LIBRARIES

ljacobson64:
This PR removes the line which included `lapack`, `blas`, `gfortran`, and `quadmath` in `MOAB_LIBRARIES`, and instead directly links those libraries as needed. `libdagtally` requires all 4, while `libfludag` only requires `quadmath`.


gonuke:
This is failing on the build of `build_obb` because of missing libraries.  I think we need to explicitly add the libraries when building executables - although I thought we'd get those from some kind of MOAB variable since these are dependencies because of MOAB

ljacobson64:
The CI is only failing during the part of the build where it builds static executables. It builds the shared executables fine.  I don't seem to have any issues building static executables on my machine, so I'm not sure what the issue is.

ljacobson64:
Well, I figured out a way to make the CI work again, but I'm not sure it's the "right" thing to do. If you add `libeigen3-dev` to the list of packages for apt to install in the Dockerfile, then it will build MOAB with eigen instead of lapack/blas, and then DAGMC won't complain about not lapack/blas not being linked.

ljacobson64:
I think the functionality in MOAB that supposedly lets users choose whether they want to use blas/lapack or eigen is broken. I can't seem to find a way to get MOAB to build without eigen on a system that contains eigen. The only way I can force it to not use eigen is to remove it from the computer entirely.

Regardless of this problem, I think that what we need to do in DAGMC is have a better way to determine whether libmoab.so is linkned to blas/lapack, and if so add the lapack/blas/gfortran/quadmath libraries to the MOAB_LIBRARIES variables. Right now (previous to this PR), those libraries are _always_ added, and as the PR stands right now, those libraries are _never_ added.

gonuke:
It seems that this is an inconsistency between CMake and autotools implementations in MOAB.  The autotools throws an error if both are enabled.  I've PR'ed a patch to provide similar behavior in CMake

gonuke:
Perhaps a root cause here is that we don't use the information provided by `MOABConfig.cmake` even once we find it...?

ljacobson64:
Something is definitely wrong with how MOAB determines whether it's going to use blas/lapack or eigen. Based on my testing, independent of what you put on the command line (and whether you use autotools or cmake), if it finds the eigen headers, it uses eigen, and if it doesn't find the eigen headers, it uses blas/lapack.

If it finds the eigen headers, `libMOAB.so` is linked to the following libraries: `libhdf5`, `libstdc++`, `libm`, `libc`, `libgcc_s`. You can see what libraries are linked by running `readelf -d libMOAB.so`. If it does not find the eigen headers, it is linked to all those libraries, plus `liblapack`, `libcblas`, and `libf77blas`.

The only differences in the `MOABConfig.cmake` file between an eigen and a blas/lapack build are the following:
```
47,48c47,48
< set(MOAB_USE_EIGEN yes)
< set(EIGEN3_DIR "/usr/include/eigen3")
---
> set(MOAB_USE_EIGEN no)
> set(EIGEN3_DIR "")
70c70
< set (MOAB_PACKAGE_INCLUDES2 "    -I/home/lucas/dagmc-build/opt/hdf5/include -isystem /home/lucas/dagmc-build/opt/hdf5/include -I/home/lucas/dagmc-build/opt/hdf5   -I/usr/include/eigen3  " )
---
> set (MOAB_PACKAGE_INCLUDES2 "    -I/home/lucas/dagmc-build/opt/hdf5/include -isystem /home/lucas/dagmc-build/opt/hdf5/include -I/home/lucas/dagmc-build/opt/hdf5     " )
```
And both of them contain these lines:
```
set(MOAB_USE_LAPACK yes)
set(LAPACK_LIBRARIES "-llapack")
```
despite the fact as far as I can tell, the eigen build does not use lapack.

ljacobson64:
So what my latest commit did is it now uses `MOAB_USE_EIGEN` to determine whether MOAB was built with eigen or with blas/lapack. I was able to get it to work properly with all combinations of eigen vs. blas/lapack and shared vs. static on my machine, and it now passes the CI.

gonuke:
Let me clarify: 
* I 100% agree that MOAB is not behaving correctly with respect to Eigen/LAPACK.  I'll push a PR to MOAB today to fix that
* I also think our `FindMOAB.cmake` does too many customized things after it successfully finds MOAB instead of relying on MOAB's variables as set in `MOABConfig.cmake`

ljacobson64:
(Full disclosure that I did this testing with MOAB's autotools build, not cmake. I will test it with cmake to see if anything is different.)

I disagree with the second point. After looking at `MOABConfig.cmake` again, all I see is more proof of why it's a good idea for us to not use some of MOAB's variables. The main variable that we used to use but don't use anymore is this:

```
set (MOAB_PACKAGE_LIBS "         -L/home/lucas/dagmc-build/opt/hdf5/lib     -lhdf5_hl  -lz -lhdf5      -llapack -lcblas -lf77blas -latlas -llapack -lcblas -lf77blas -latlas  -ldl -lz -lm -lm -lm")
```

That variable is the same regardless of whether MOAB was built with blas/lapack or eigen. So even if MOAB is built with eigen, and it has no dependency on blas/lapack, it still includes `-llapack -lcblas -lf77blas -latlas -llapack -lcblas -lf77blas -latlas` in its list of libraries.

Similarly, these variables appear to always be set no matter what:
```
set(MOAB_USE_BLAS @enableblas@)
set(BLAS_LIBRARIES "-lcblas -lf77blas -latlas")
set(MOAB_USE_LAPACK yes)
set(LAPACK_LIBRARIES "-llapack")
```

gonuke:
I agree that `MOABConfig.cmake` is also not doing everything right, but we can influence that and we should try to move both sides to that ideal

It is also very likely that the version of `MOABConfig.cmake` that is generated by autotools is the worst of all worlds since most people who test with autotools use it all the way through and vice versa

ljacobson64:
Here's the diff for cmake:
```
18,19c18,19
< set(MOAB_CFLAGS "-O2 -DNDEBUG -O3 -DNDEBUG -fPIC -DPIC -pedantic -fpic -Wall -pipe -Wno-long-long -Wextra -Wno-cast-align -Wsign-compare -Wpointer-arith -Wformat -Wformat-security -Wunused-parameter -fstack-protector-all -mtune=native -fpermissive -Wno-ignored-attributes -Wno-variadic-macros -Wno-deprecated-declarations -Wno-unused-local-typedefs -O2 -funroll-loops -funroll-all-loops -fstrict-aliasing -Wno-unused -I/usr/include/eigen3")
< set(MOAB_CXXFLAGS "-O2 -DNDEBUG -O3 -DNDEBUG -fPIC -DPIC -pedantic -fpic -Wall -pipe -Wno-long-long -Wextra -Wno-cast-align -Wsign-compare -Wpointer-arith -Wformat -Wformat-security -Wunused-parameter -fstack-protector-all -mtune=native -fpermissive -Wno-ignored-attributes -Wno-variadic-macros -Wno-deprecated-declarations -Wno-unused-local-typedefs -O2 -funroll-loops -funroll-all-loops -fstrict-aliasing -Wno-unused -I/usr/include/eigen3")
---
> set(MOAB_CFLAGS "-O2 -DNDEBUG -O3 -DNDEBUG -fPIC -DPIC -pedantic -fpic -Wall -pipe -Wno-long-long -Wextra -Wno-cast-align -Wsign-compare -Wpointer-arith -Wformat -Wformat-security -Wunused-parameter -fstack-protector-all -mtune=native -fpermissive -Wno-ignored-attributes -Wno-variadic-macros -Wno-deprecated-declarations -Wno-unused-local-typedefs -O2 -funroll-loops -funroll-all-loops -fstrict-aliasing -Wno-unused ")
> set(MOAB_CXXFLAGS "-O2 -DNDEBUG -O3 -DNDEBUG -fPIC -DPIC -pedantic -fpic -Wall -pipe -Wno-long-long -Wextra -Wno-cast-align -Wsign-compare -Wpointer-arith -Wformat -Wformat-security -Wunused-parameter -fstack-protector-all -mtune=native -fpermissive -Wno-ignored-attributes -Wno-variadic-macros -Wno-deprecated-declarations -Wno-unused-local-typedefs -O2 -funroll-loops -funroll-all-loops -fstrict-aliasing -Wno-unused ")
48c48
< set(EIGEN3_DIR "/usr/include/eigen3")
---
> set(EIGEN3_DIR "EIGEN3_DIR-NOTFOUND")
71c71
< set (MOAB_PACKAGE_INCLUDES_LIST "   /home/lucas/dagmc-build/opt/hdf5/include   -I/usr/include/eigen3  " )
---
> set (MOAB_PACKAGE_INCLUDES_LIST "   /home/lucas/dagmc-build/opt/hdf5/include   -IEIGEN3_DIR-NOTFOUND  " )
```

Notably, these lines are unchanged regardless of whether blas/lapack or eigen are used:
```
set(MOAB_USE_BLAS ON)
set(BLAS_LIBRARIES "/usr/lib/x86_64-linux-gnu/libf77blas.so;/usr/lib/x86_64-linux-gnu/libatlas.so")
set(MOAB_USE_LAPACK ON)
set(LAPACK_LIBRARIES "/usr/lib/x86_64-linux-gnu/liblapack.so;/usr/lib/x86_64-linux-gnu/libf77blas.so;/usr/lib/x86_64-linux-gnu/libatlas.so")
```

And even more notably is that this line is always unchanged as well:
```
set(MOAB_USE_EIGEN )
```

This means that my latest fix, which depends on the value of `MOAB_USE_EIGEN`, will not work if MOAB is built with cmake because that variable is always empty.

gonuke:
What is that diff comparing? Cmake vs autotools? or two configuration of CMake?

ljacobson64:
Both are cmake. The diff is between two different moab builds, one with eigen and the other with lapack/blas.

The second snippet shows that `MOAB_USE_BLAS` and `MOAB_USE_LAPACK` are always set to `ON` regardless of whether blas or lapack are actually used.

gonuke:
What was your `cmake` line in each case?  I am trying to figure out how to fix MOAB's CMake setup

ljacobson64:
I used this for both:

```
-DBUILD_SHARED_LIBS=ON
-DENABLE_PYMOAB=ON
-DENABLE_HDF5=ON
-DHDF5_ROOT=/home/lucas/dagmc-build/opt/hdf5
-DCMAKE_BUILD_TYPE=Release
-DCMAKE_C_COMPILER=/usr/bin/gcc
-DCMAKE_CXX_COMPILER=/usr/bin/g++
-DCMAKE_Fortran_COMPILER=/usr/bin/gfortran
-DCMAKE_INSTALL_PREFIX=/home/lucas/dagmc-build/opt/moab
-DCMAKE_INSTALL_RPATH=/home/lucas/dagmc-build/opt/hdf5/lib:/home/lucas/dagmc-build/opt/moab/lib
```

The only way I've been able to reliably get it to not use eigen is to remove eigen from the computer entirely. So the only difference between the two builds is that in the eigen build, eigen exists on the computer, and in the blas/lapack build, eigen does not exist on the computer.

gonuke:
What about `-DENABLE_BLASLAPACK=ON` or `-DENABLE_BLASLAPACK=OFF`?

ljacobson64:
I will leave eigen installed and try building it with both `-DENABLE_BLASLAPACK=ON` and `-DENABLE_BLASLAPACK=OFF` and report the results

pshriwise:
Have we looked into having MOAB export CMake targets when installing? This would only help the CMake MOAB builds perhaps but it’s been proven to work well in many other packages.

gonuke:
I expect you'll get Eigen in both cases because I have a pending PR to make sure Eigen doesn't show up when LAPACK does, but the case with LAPACK off should only include Eigen

gonuke:
What do you mean by "export CMake targets"?  MOAB already does generate `MOABConfig.cmake` files, but I'm not sure they are correct/optimal

ljacobson64:
First `MOABConfig.cmake` was created with `-DENABLE_BLASLAPACK=OFF`, second file with `-DENABLE_BLASLAPACK=ON`

```
$ diff MOABConfig.cmake_eigen MOABConfig.cmake_lapack
43,46c43,46
< set(MOAB_USE_BLAS )
< set(BLAS_LIBRARIES "")
< set(MOAB_USE_LAPACK )
< set(LAPACK_LIBRARIES "")
---
> set(MOAB_USE_BLAS ON)
> set(BLAS_LIBRARIES "/usr/lib/x86_64-linux-gnu/libf77blas.so;/usr/lib/x86_64-linux-gnu/libatlas.so")
> set(MOAB_USE_LAPACK ON)
> set(LAPACK_LIBRARIES "/usr/lib/x86_64-linux-gnu/liblapack.so;/usr/lib/x86_64-linux-gnu/libf77blas.so;/usr/lib/x86_64-linux-gnu/libatlas.so")
64c64
< set (MOAB_PACKAGE_LIBS "    /home/lucas/dagmc-build/opt/hdf5/lib/libhdf5_hl.so;/home/lucas/dagmc-build/opt/hdf5/lib/libhdf5.so;/usr/lib/x86_64-linux-gnu/libdl.so;/usr/lib/x86_64-linux-gnu/libsz.so;/usr/lib/x86_64-linux-gnu/libz.so;/usr/lib/x86_64-linux-gnu/libm.so;/usr/lib/x86_64-linux-gnu/libcurl.so      " )
---
> set (MOAB_PACKAGE_LIBS "    /home/lucas/dagmc-build/opt/hdf5/lib/libhdf5_hl.so;/home/lucas/dagmc-build/opt/hdf5/lib/libhdf5.so;/usr/lib/x86_64-linux-gnu/libdl.so;/usr/lib/x86_64-linux-gnu/libsz.so;/usr/lib/x86_64-linux-gnu/libz.so;/usr/lib/x86_64-linux-gnu/libm.so;/usr/lib/x86_64-linux-gnu/libcurl.so    /usr/lib/x86_64-linux-gnu/liblapack.so;/usr/lib/x86_64-linux-gnu/libf77blas.so;/usr/lib/x86_64-linux-gnu/libatlas.so /usr/lib/x86_64-linux-gnu/libf77blas.so;/usr/lib/x86_64-linux-gnu/libatlas.so " )
```

The lines pertaining to eigen were the same in both files despite the fact that the blas/lapack build should not be using eigen. The blas/lapack build contained a lot of info about where it's finding blas and lapack, but if you actually look at `libMOAB.so`, it isn't linked to blas or lapack.

gonuke:
You are seeing what I expect prior to my PR (you can try branch [gonuke/eigen_lapack_unique](https://bitbucket.org/fathomteam/moab/branch/gonuke/eigen_lapack_unique) in the upstream repo)

ljacobson64:
I tested your branch and it seems to fix the problem

gonuke:
It looks like MOAB's CI is broken... so I am not sure when that get's merged... :frowning_face:

ljacobson64:
What's the plan for the autotools build? Our website still tells users to build moab with autotools

gonuke:
First, I'd like to steer our users towards CMake, but I thought autotools already produced a valid `MOABConfig.cmake`.

ljacobson64:
It has the same problem where the only way to actually get it to build with blas/lapack is to remove eigen from the system

pshriwise:
Exported targets are a little different than the configuration file. CMake handles these automatically. When targets are exported, their linker, compiler, etc. info is all included in a file that, in this case, would be something like `MOABTargets.cmake`. Then in other packages relying on moab, this target file gets included as part of the `FindMOAB.cmake` and then targets can be used as though they were created as part of any projects that file is included into, in this case, DAGMC.


```
find_package(MOAB REQUIRED) # MOABTargets.cmake included in this step
add_library(dagmc <dagmc_sources>)
target_link_library(dagmc MOAB)
```

(Not sure about the exact syntax/names there. I'm on the go)

More specific info (and a better explanation) is here [exporting targets](https://gitlab.kitware.com/cmake/community/wikis/doc/tutorials/Exporting-and-Importing-Targets)

gonuke:
@pshriwise  - Yes - so I think we definitely need to up our game on the use of CMake to make things more smooth.  

@ljacobson64 - My bad - I misinterpreted a line in `configure.ac`.  I thought it prevented them both being invoked, but it does the opposite and requires at least one of them.  I guess I'll have to update that too

pshriwise:
@gonuke Yeah I'd love to export targets in DagMC too - hoping to do that sometime soon.

ljacobson64:
There is still one problem though. This PR depends on the `MOAB_USE_EIGEN` variable to determine whether moab is built with blas/lapack or eigen. But only the autotools build sets this variable (to `yes`). With the cmake build it is set to empty. So our CI will only pass because it uses the autotools build.

gonuke:
Just pushed a fix for that to my branch.  (MOAB build is a hack job from too many people)

I think my original fix may not actually be right - I think it should be possible to build with both LAPACK and Eigen because `Matrix3` should use the one specified in build, but `HYPRE` only supports Eigen3... ugh!

gonuke:
I think I just made it better for CMake, but now to tackle autotools...

gonuke:
Actually - autotools was probably fine with the change I just made to `Matrix3` in my MOAB branch

ljacobson64:
Just tested your updated branch:

eigen/autotools: looks good
lapack/autotools: looks like it's still using eigen and not blas/lapack
eigen/cmake: looks good
lapack/cmake: long list of build errors. The first error is
```
In file included from /home/lucas/dagmc-build/opt/moab-lapack-cmake/src/src/AffineXform.hpp:22:0,
                 from /home/lucas/dagmc-build/opt/moab-lapack-cmake/src/src/io/ReadABAQUS.cpp:34:
/home/lucas/dagmc-build/opt/moab-lapack-cmake/src/src/moab/Matrix3.hpp:169:3: error: ‘Eigen’ does not name a type
   Eigen::Matrix3d _mat;
   ^~~~~
```

For autotools, I'm using `--disable-blaslapack` to try to use eigen and `--enable-blaslapack` to try to use blas/lapack. For cmake, I'm using `-DENABLE_BLASLAPACK=OFF` and `-DENABLE_BLASLAPACK=ON` respectively.

ljacobson64:
We could potentially say that we (DAGMC) are not going to support the moab autotools build anymore. Then we wouldn't have to maintain it. We'd need to change our website and CI though.

gonuke:
Now that I look more carefully, I definitely missed some lines in that update.

gonuke:
Just cleaned it up...

ljacobson64:
Your changes fixed the build errors

gonuke:
For all 4 combinations? eigen vs lapack & autotools vs CMake?

ljacobson64:
I meant that the compiler errors in the lapack/cmake case were fixed. The lapack/autotools case still results in libraries that are built with eigen, not lapack.

gonuke:
I don't have time to dig into the autotools issues at the moment.  It is kind of complicated to see where those parameters get set.

One thing I don't like about autotools is that you can't use it with a system HDF5 - at least I can't figure out how with Ubuntu 16.04 - because it wants the includes and libs to be in the same hierarchy: some/path/lib and some/path/include.  U16.04 has /usr/lib/x86_64-gnu-linux/hdf5 and /usr/include/hdf5/serial/

ljacobson64:
What exactly does it mean for MOAB to be built with both eigen and blas/lapack? Right now, if you
* use MOAB's cmake build on your new branch,
* tell it to build with blas/lapack, and
* have eigen on your machine,

both `MOAB_USE_EIGEN` and `MOAB_USE_LAPACK` are set to `ON`. From my understanding, one of those should always be `ON` and the other should be `OFF`.

makeclean:
The header file Matrix3.hpp looks for Eigen first, so it would expect Eigen. I guess that would be followed in the linkage, since there would be no matching symbols found when trying to link BLAS/LAPACK. You can check ldd or something, you should see the final library would only depend on Eigen. If I'm right that is :)

ljacobson64:
This was kind of my point. If it "uses both," it really only uses eigen.

gonuke:
I am working on a PR (currently just a branch) that reverses the precedence of EIGEN and LAPACK in `Matrix3.hpp`.  Currently, as I see it, MOAB's HYPRE tool requires Eigen3, but the option to use LAPACK was introduced in `Matrix3`.

In the current code, using the precedence that Andy correctly indicates, you cannot use Eigen in HYPRE and LAPACK in Matrix3.  Rather, you get Eigen everywhere or LAPACK without HYPRE.

My fix will allow the combination.

That said, I'm not sure what the rest of the MOAB team wants/prefers/expects!?!

makeclean:
From my perspective, the current way is better, Eigen for the win - no fortran linkage.

gonuke:
Agreed!  We'll still be able to build a pure Eigen version, which we prefer in our project for sure.

gonuke:
If we assume only CMake builds with `ENABLE_BLASLAPACK=OFF` does this PR fix all our problems?

ljacobson64:
Yes. If we only build moab with cmake, and we disable blas/lapack, then this PR should work.

The way the PR is set up right now is that if `MOAB_USE_LAPACK` evaluates to true (i.e. it is `ON` or `yes`), then it assumes moab is linked to lapack. If it evaluates to false (i.e. it is `OFF`, `no`, or empty), then it assumes moab is not linked to lapack.

The only situation I've been able to find where this PR would _not_ work is if you use moab 5.1.0, enable blas/lapack, have eigen present on the system, and use the cmake build. In this specific case, `MOAB_USE_LAPACK` will be set to `ON`, but moab will not be linked to lapack.

In every other case I tested, the value of `MOAB_USE_LAPACK` is set correctly depending on whether or not moab is linked to lapack.

pshriwise:
It appears that `libblas` and friends are no longer present in the `develop` branch. Does this PR contain other changes we might want to keep or should it be closed?

gonuke:
I think they were removed when I ported DAGtally to Eigen3

pshriwise:
Ok so I think we're good to close this PR if that's alright w/ you @ljacobson64.

src/fluka/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ find_package(Fluka REQUIRED)
 set(SRC_FILES fluka_funcs.cpp)
 set(PUB_HEADERS fluka_funcs.h)
 
-set(LINK_LIBS ${FLUKA_LIBRARIES} dagmc pyne_dagmc uwuw)
+set(LINK_LIBS ${FLUKA_LIBRARIES} dagmc pyne_dagmc uwuw quadmath)
```
ljacobson64, line 8: This is the only line remaining in this PR that I think is still needed. You cannot build a static fludag executable without this change.

ljacobson64: I'm still going to close this PR though. That one line can be its own PR (or we can just ignore it and leave it broken)

Issue #639, opened 2019-09-14
Overlap Check Tool


pshriwise:
Someone at the DagMC workshop in Madison mentioned an overlap check tool for discretized DAGMC models. I'm in the process of polishing up a version of this now - just need to add tests. I'm creating this issue to avoid work duplication and make everyone aware of the method I intend to use. 

First, the tool will check all vertices of the problem for containment by more than one volume. My plan is that this will be the default mode. Optionally, users can provide a certain number of points to check along each triangle edge as well. While more thorough, this takes quite a bit of extra time - hence it being an optional approach. Output from the program will make users aware of this option if it is not enabled, as checking the edges of the triangles will make the search more robust.

My thinking on triangle edges is that as the number of points we check on the edges approaches infinity, we should find all overlap (to within machine precision) - i.e. points in the middle of the triangle shouldn't matter because the extent and orientation of the triangle is defined by its edges. I could definitely be missing something there though, so please let me know if there's something more I should look into.

pshriwise:
Related issue: #629

gonuke:
resolved by #641

Issue #629, opened 2019-06-16
Detect enclosed volumes


gonuke:
Feature/Problem: This feature request is related to the possible problem of completely enclosed volumes.  It may be possible for an imported geometry to include two volumes that overlap because one volume is completely enclosed within another.  This would represent an upstream flaw in the model creation, but it would be valuable if such a flaw were detected before running the problem.  A model with such features will run successfully with no indications of an error, but the results are not reliable.  This is a special case of overlapping geometries, and perhaps a new feature should be prepared to detect all overlapping geometries.

Solutions: Capability can be created either within an existing tool or in a new tool that will test for this kind of overlap (or all kinds of overlaps).  

The brute force solution is to test every vertex of every surface to ensure that it is not included in any volume.  In a watertight geometry, such a test would exclude the volumes that are known to contain the vertex on its boundary.  Such a test is probably prohibitively expensive.

There may be capability within Cubit/Trelis that can be invoked to detect overlaps, and user documentation should be updated to advise such a test in the workflow.

makeclean:
Trellis/Cubit's overlpap check does cover this case, the documentation already says do an overlap check https://svalinn.github.io/DAGMC/usersguide/trelis_workflow.html. is there a case in the wild that you've found that Trelis is not robust to?

gonuke:
See the [recent email from Wei Lu on the DAGMC users list](https://groups.google.com/d/msg/dagmc-users/-uW1Fb5TSww/0mN4865mBwAJ).

PR #641, opened 2019-09-16
Overlap Check Tool

pshriwise:
## Description
This PR implements the overlap checking tool discussed in #639.

I've tested it on FNG, ATR, & UWNR (several models we know are successfully imprinted, merged, and sealed) to ensure that overlaps are *not* found for those cases. I've also included a few test cases here to make sure it will find overlaps successfully - more details below.

### Building

Unlike other executables in the repo, I've opted not to build a library for the overlap tool. I didn't see a compelling case for someone using this as a library, but I could be swayed otherwise.

### Test cases

I've added a few tests here to make sure that we find overlaps when we should.

  - A sphere overlapping a box
  - A cylinder and cube with coincident surfaces, unimprinted
  - A cylinder and cube with coincident surfaces, imprinted
  - One cylinder completely enclosing another
  - A slight overlap (1E-06 thickness) between a cylinder and box*

* - This overlap was too small to be detected using Trelis' default settings, but is caught by the overlap tool. Were imprinting to be used, no overlap would exist, but this was more to represent a case where imprinting might fail due to topological problems in the CAD model.

There are probably more test cases that would be useful here too, but I don't have the bandwidth to look into that right now.

### Bumping locations

I'm sure this will come up in the review, so I'll address it early. There are cases in which MOAB really doesn't like it if rays are fired through vertices. A lot of nasty output occurs which we cannot turn off. Rather than force users to suffer through this output, I bump the test location a bit and make sure we are firing rays away from the vertex's surface to avoid that output. I understand that this isn't ideal, but I think it's the best thing to do until that problem is resolved in MOAB.

Example of this output:

```
Vertex 667:
Global id = 0
Coordinates: (-7.5, 40.57, 5.99)
  Adjacencies:
   Edge 128095, Edge 128096, Edge 128100, Edge 296773, Edge 392728, Edge 413190, Edge 413192, Edge 961661, Edge 961662, Edge 1141656, Edge 1141657, Edge 1190932, Edge 1190933, Edge 1190934, Edge 1190935, Edge 1190944, Edge 1190945, Edge 1190946, Edge 1190947, Edge 1190948, Edge 1190949, Edge 1190950, Edge 1190951, Edge 1190952, Edge 1190953, Edge 1190954, Edge 1190955, Edge 1190956, Edge 1190957, Edge 1190958, Edge 1190959, Edge 1190960, Edge 1190961, Edge 1190962, Edge 1190963, Edge 1190964, Edge 1190965, Edge 1190966, Edge 1190967, Edge 1190968, Edge 1190969, Edge 1190970, Edge 1190971, Edge 1190972, Edge 1190973, Edge 1190974, Edge 1190975, Edge 1190976, Edge 1190977, Edge 1190978, Edge 1190979, Edge 1190980, Edge 1190981, Edge 1190982, Edge 1190983, Edge 1190984, Edge 1190985, Edge 1190986, Edge 1190987, Edge 1190988, Edge 1190989, Edge 1190990, Edge 1190991, Edge 1190992, Edge 1190993, Edge 1190994, Edge 1190995, Edge 1190996, Edge 1190997, Edge 1190998, Edge 1190999, Edge 1191000, Edge 1191001, Edge 1191002, Edge 1191003, Edge 1191004, Edge 1191005, Edge 1191006, Edge 1191007, Edge 1191008, Edge 1191009, Edge 1191010, Edge 1191011, Edge 1191012, Edge 1191013, Edge 1191014, Edge 1191015, Edge 1191016, Edge 1191017, Edge 1191018, Edge 1191019, Edge 1191020, Edge 1191021, Edge 1191022, Edge 1191023, Edge 1191024, Edge 1191025, Edge 1191026, Edge 1191027, Edge 1191028, Edge 1191029, Edge 1191030, Edge 1191031, Edge 1191032, Edge 1191033, Edge 1191034, Edge 1191035, Edge 1191036, Edge 1191037, Edge 1191038, Edge 1191039, Edge 1191040, Edge 1191041, Edge 1191042, Edge 1191043, Edge 1191044, Edge 1191045, Edge 1191046, Edge 1191047, Edge 1191048, Edge 1191049, Edge 1191050, Edge 1191051, Edge 1191052, Edge 1191053, Edge 1191054, Edge 1191055, Edge 1191056, Edge 1191057, Edge 1191058, Edge 1191059, Edge 1191060, Edge 1191061, Edge 1191062, Edge 1191063, Edge 1191064, Edge 1191065, Edge 1191066, Edge 1191067, Edge 1191068, Edge 1191069, Edge 1194003
```
and so on...

## Motivation and Context
See issues #629 and #639 

## Changes
Adds a feature for overlap detection.

## Behavior
Added tool for detecting overlaps in DagMC models.


gonuke:
Any comments on speed?  
Related: can we thread this (in a different PR)?

pshriwise:
Sure! Turns out it wasn't too bad. I need to clean it up a bit though. Parallelism and some timings coming tomorrow.

makeclean:
Have we tested it on Wei's case?

Get Outlook for Android<https://aka.ms/ghei36>

pshriwise:
If someone can provide the model I’d be happy to try it.

makeclean:
I've emailed them to you, not sure about a public share.

Get Outlook for Android<https://aka.ms/ghei36>

pshriwise:
The tool found a number of overlaps in the SNS model - including an overlap between volumes 3 & 6 which is the one Wei Lu presented at the workshop I believe.

CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ dagmc_setup_build()
 dagmc_setup_options()
 
 find_package(MOAB REQUIRED)
+find_package(OpenMP REQUIRED)
```
ljacobson64: I think it would make more sense for this line to be in `src/overlap_check/CMakeLists.txt` instead because it's the only place where OpenMP is used.

Also, should it actually be `REQUIRED`? A cursory glance at the code indicates that it should still work even if OpenMP is disabled.

src/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ endif ()
 # Required libraries
 add_subdirectory(dagmc)
 add_subdirectory(pyne)
+add_subdirectory(overlap_check)
```
ljacobson64: I think it would be better to use the format of the other packages for this tool, since it isn't strictly required to build this to build DAGMC; i.e.
```
if (BUILD_OVERLAP_CHECK)
  add_subdirectory(overlap_check)
endif ()
```

This would require adding the line
```
option(BUILD_OVERLAP_CHECK "Build overlap_check tool" ON)
```
to `cmake/DAGMC_macros.cmake` in the `dagmc_setup_options()` macro.

src/overlap_check/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+message("")
+
+set(LINK_LIBS dagmc)
+set(LINK_LIBS_EXTERN_NAMES)
+
+add_subdirectory(test)
```
ljacobson64: This statement should be surrounded by an `if (BUILD_TESTS)` check. `BUILD_TESTS` is an option that defaults to ON, but all other tests in DAGMC can be disabled by explicitly turning it off.

news/PR-0641.rst
```diff
@@ -0,0 +1,12 @@
+**Added:** None
+Tool for checking DagMC models for overlaps.
```
ljacobson64, line 3: A blank line is needed here or else the formatting will be messed up. Also the word "None" should be removed

src/overlap_check/build/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+else()
+  set(LINK_LIBS dagmc)
+endif()
+
+set(LINK_LIBS_EXTERN_NAMES)
+
+set(INSTALL_INCLUDE_DIR_SAVE ${INSTALL_INCLUDE_DIR})
+set(INSTALL_INCLUDE_DIR ${INSTALL_INCLUDE_DIR}/overlap_check)
+
+set(INSTALL_INCLUDE_DIR ${INSTALL_INCLUDE_DIR_SAVE})
```
ljacobson64: If we're not installing any public headers, then these 3 lines concerning `INSTALL_INCLUDE_DIR` are not needed.

src/overlap_check/ProgressBar.cpp
```diff
@@ -0,0 +1,78 @@
+
```
ljacobson64, line 1: Just out of curiosity... have you tested whether the progress bar affects the speed this tool runs? A few years back I added a progress bar to some project I was working on, and I noticed that it slowed it down. I'm sure if it's implemented properly it won't slow anything down but I want to make sure.

src/overlap_check/build/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+message("")
```
ljacobson64: Very nitpicky but this blank line is not needed

ljacobson64 (changes requested):
Just a few comments on the build system here.

The CI seems to be failing for multiple reasons.
* The housekeeping run is failing because of the style guide.
* The gcc runs are failing because the `overlap_check_test` test is segfaulting. I just ran this test on my machine and it succeeded. Not sure what's going on here.
* The clang runs are failing because CMake is having problems finding OpenMP. I get this same error on my machine. Not sure how to fix it.

pshriwise:
Timings for FNG (haven't had a chance to run other problems today)

| N Threads | Time (s)     |      |      |
|-----------|----------------|------|------|
|           | Tri Verts Only | 1 point per edge | 3 points per edge |
| 1         |      86        | 354 | 757 |
| 4         |      42        | 133  | 318  |
| 8         |      31        | 100  | 302  |

CMakeLists.txt
pshriwise: No it definitely shouldn't be required. Meant to remove that. Thanks!

pshriwise:
Didn't mean to push the parallel parts yet! Thank you for the review @ljacobson64. Most of the CMake comments probably would've been needed anyways.

src/overlap_check/ProgressBar.cpp
pshriwise: Good point. I looked at that a bit, yeah. I saw that it adds 10-15% to the runtime. I think the additional time is worth the indication of progress to our users - especially given how long it could take for the check to complete.

pshriwise:
Figured out that CI issue. `gcc` doesn't like building static executables with OpenMP I guess, so the recommended fix for this is to bring along the whole of `-lpthread` into the executable. This is what I've done to get CI working.

I'm also open to simply not building with OpenMP for a static build to reduce unforeseen problems, yet being able to thread this program does make it considerably faster. Do we think users will be running this in environments where they're using static executables? (i.e. HPC environments - or at least that's my understanding of our main drive for static build support).

makeclean:
Tbh I never ever build static executables, especially.on hpc systems. I don't know about anyone else?

Get Outlook for Android<https://aka.ms/ghei36>

pshriwise:
@makeclean? @ljacobson64? Any thoughts?

CMakeLists.txt
ljacobson64: I think you've addressed all my comments other than the first part of this one. I'll mark this PR as approved once it's taken care of.

pshriwise: I fixed this mistake already I think. The diff here is outdated.

ljacobson64: The first part of my comment, not the second part

pshriwise: Oh, sorry! Yup. Happy to update that.

ljacobson64 (approved):

pshriwise:
Any further comments here @gonuke @makeclean?

.gitignore
```diff
@@ -6,3 +6,7 @@
 
 # Ignore built documentation
 gh-build
+
+# Ignore common build directory names
+bld
+build
```
gonuke, line 12: This seems wise, but I notice that this PR has files in a directory named `build`

src/overlap_check/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+message("")
+
+set(LINK_LIBS dagmc)
+set(LINK_LIBS_EXTERN_NAMES)
+
+find_package(OpenMP)
+
+add_subdirectory(build)
```
gonuke, line 13: Maybe this directory should be `cli` or something more specific? (Although I'm not convinced that it needs its own subdirectory?)

src/overlap_check/test/overlap_test_driver.cc
```diff
@@ -0,0 +1,30 @@
+#include <string>
+#include <stdio.h>
+
+#include <gtest/gtest.h>
+
+int main(int argc, char* argv[]) {
+  for (int i = 0; i < argc; i++) {
+    std::string arg = argv[i];
+    if (arg == "--help") {
```
gonuke: Why did I think that `gtest` automatically generated its own help message?

src/overlap_check/overlap.cpp
```diff
@@ -0,0 +1,184 @@
...
+                   << GTT->global_id(vol)
+                   << " at location " << loc);
+
+    if (result == 1) {
+      vols_found.insert(GTT->global_id(vol));
+    }
+  }
+
+  if (vols_found.size() > 1) {
+    overlap_map[vols_found] = loc;
```
gonuke, line 37: Should this be the original location rather than the bumped location?  Or is there any value in a vertex ID here?

gonuke (commented):
This looks pretty good to me.  Just a couple of discussion questions, but no show-stoppers.

src/overlap_check/CMakeLists.txt
ljacobson64: The paradigm we've been using for the rest of the stuff we build in DAGMC is to have executables built in a subdirectory called `build`

.gitignore
ljacobson64: Good catch. Pretty sure `.gitignore` entires are recursive, so any changes to any files in directories called `build` (of which there are several) will not be picked up if this line remains.

src/overlap_check/CMakeLists.txt
gonuke: I see - I guess I haven't paid much attention to this before - is there a reason to avoid doing it in the same directory?

.gitignore
gonuke: Actually - if they are already tracked, then git will still pick up changes.  If new files are added in such a directory, then they will not be noticed, and a developer will need to be carefully to manually add them

I propose that we stop using `build` as the directory name for source

pshriwise: I agree and created #644 so this can be addressed in a future PR.

src/overlap_check/CMakeLists.txt
pshriwise: Not that I'm aware of. I'd be on board with that model.

src/overlap_check/test/overlap_test_driver.cc
pshriwise: It does! That's what I get for copy-pasting from other drivers without question.

pshriwise: I've removed it from the overlap test, but I'll leave other test drivers to be updated as a group in a separate PR. Any known reason for this set of code reappearing in our drivers? @ljacobson64? @makeclean?

src/overlap_check/overlap.cpp
pshriwise: My only concern is that if a user then tries to inspect the model at this location, they may get a different result than the location actually tested in the overlap check. Which leads us back to ideally testing exact vertex locations, but the verbosity of what comes out of MOAB when we do that is really prohibitive...

gonuke: I guess if we are using a location instead of a vertex ID, the bump doesn't matter that much - it's almost round off error.

src/overlap_check/CMakeLists.txt
ljacobson64: I was mistaken. If there is no library being built, then we don't need the `build` or `app` subdirectory. See the `src/build_obb` directory.

src/overlap_check/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+
+if (BUILD_TESTS)
+add_subdirectory(test)
```
bam241: missing indent ?
(not a big deal but to follow other CMakeFile style)

src/overlap_check/ProgressBar.cpp
```diff
@@ -0,0 +1,78 @@
...
+    bar.append("  0");
+  } else {
+    std::stringstream ss;
+    ss << std::setfill(' ') << std::setw(3) << (int)val;
+    bar.append(ss.str());
+  }
+
+  bar.append("% |");
+  // remaining width of the bar
+  int remaining_width = BAR_WIDTH - bar.size() - 2;
```
bam241: is the bar size not supposed always to be 3 at this point ?
Not that we are concern about speed there, would in theory would it not be faster to use `3` instead of `bar.size()` ?

src/overlap_check/ProgressBar.cpp
```diff
@@ -0,0 +1,78 @@
...
+#ifdef _POSIX_VERSION
+  return isatty(STDOUT_FILENO) != 0;
+#else
+  return false;
+#endif
+}
+
+
+void
+ProgressBar::set_value(double val) {
```
bam241: why a line break there ?

src/overlap_check/ProgressBar.cpp
```diff
@@ -0,0 +1,78 @@
...
+  if (val >= 100.0) {
+    bar.append(remaining_width, '=');
+  } else if (val < 0.0) {
+    bar.append(remaining_width, ' ');
+  } else {
+    int width = (int)((double)remaining_width * val / 100);
+    bar.append(width, '=');
+    bar.append(1, '>');
+    bar.append(remaining_width - width - 1, ' ');
+  }
```
bam241, line 45: Do you need the if/else if /else ? 
could you not at the beginning do something like:
`if val > 100 > val =100`
`else if val < 0 val = 0`

and then just:
```
std::stringstream ss;
ss << std::setfill(' ') << std::setw(3) << (int)val;
bar.append(ss.str());
```
 and then
```
int width = (int)((double)remaining_width * val / 100);
bar.append(width, '=');
bar.append(1, '>');
bar.append(remaining_width - width - 1, ' ');
```

without special cases ?

src/overlap_check/build/overlap_check.cpp
```diff
@@ -0,0 +1,56 @@
...
+  // check for overlaps
+  OverlapMap overlap_map;
+  rval = check_instance_for_overlaps(MBI, overlap_map, points_per_tri_edge);
+  MB_CHK_SET_ERR(rval, "Failure while checking for overlaps");
+
+  // if any overlaps are found, report them
+  if (overlap_map.size() > 0) {
+    report_overlaps(overlap_map);
+  } else {
+    std::cout << "No overlaps were found." << std::endl;
```
bam241, line 70: maybe remind the number point of points per triangle edge there, to make sure people understand that it is not an exhaustive check

src/overlap_check/overlap.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+      EntityHandle vert = all_verts[i];
+      CartVect loc;
+      rval = MBI->get_coords(&vert, 1, loc.array());
+
+      rval = check_location_for_overlap(GQT, all_vols, loc, dir, overlap_map);
+      MB_CHK_SET_ERR_CONT(rval, "Failed to check location " << loc << " for an overlap");
+
+      #pragma omp critical
+      prog_bar.set_value(100.0 * (double) num_checked++ / (double) num_locations);
```
bam241, line 117: If you have a high number of locations it will be updating the progressBar for nothing.

As you progress bar only updates per integer % why not make sure that you hit a new % before updating the progress bar printout ? (here or in progress bar...)

src/overlap_check/overlap.cpp
```diff
@@ -0,0 +1,184 @@
...
+        double t = (double)j / (double)pnts_per_edge;
+        locations.push_back(start + t * edge_vec);
+      }
+
+      // check edge locations
+      for (auto& loc : locations) {
+        rval = check_location_for_overlap(GQT, all_vols, loc, dir, overlap_map);
+        MB_CHK_SET_ERR_CONT(rval, "Failed to check point for overlap");
+        #pragma omp critical
+        prog_bar.set_value(100.0 * (double) num_checked++ / (double) num_locations);
```
bam241, line 159: same here

bam241 (commented):
LGTM!
some minor style comments....

src/overlap_check/ProgressBar.cpp
pshriwise: Just a different style. Doesn't seem to violate our style guide, but I know that's not something we do in DAGMC. I'll put those on the same line.

src/overlap_check/build/overlap_check.cpp
pshriwise: Good idea. I opted to do this at the beginning of the program though, so users can see it earlier in case they had a typo in their command.

src/overlap_check/ProgressBar.cpp
pshriwise: The problem with that is that when we reach 100% `remaining_width` will equal `width` and the final `bar.append` call will have an argument of `-1` (`bar.append(-1, ' ')`). So I could put another if statement there, but I think that what exists delineates the three cases we're concerned with more clearly. Also, this would technically result in move overall calls to `str::append` if we're concerned about performance.

src/overlap_check/overlap.cpp
pshriwise: For sure. I'll add some state to the bar to track that and exit early if we're going to get the same bar already being displayed.

src/overlap_check/ProgressBar.cpp
pshriwise: It might make a bit of a difference. At any rate, it's more succint to just use `int remaining_width = BAR_WIDTH - 5;` here.

pshriwise:
I realized it's a little irresponsible to let the program claim all available threads. I've added an option to specify the number of threads if built with OpenMP.

src/overlap_check/ProgressBar.cpp
gonuke: Not sure I understand why the -1 is there at all? 

What about:
```
val = max(min(100,val),0);
int width = (int)((double)remaining_width * val / 100);
bar.append(width, '=');
bar.append(remaining_width - width , ' ');
```

pshriwise: The `-1` is for the arrowhead (`bar.append(1, '>')`) which I at some point removed accidentally.

I do like the `val = max(min(100.0, val), 0.0)` update quite a bit and moving that `-1` to another location (basically lumping any extra space needed for the prefix, suffix, and arrowhead into the `remaining_width` initialization) let's us get rid of a lot of those `if`'s as @bam241 suggested. This comes at the cost of the arrowhead still appearing when the bar is full rather than getting replaced with another `=` character, but the samll aesthetic change doesn't bother me much.

pshriwise: Haven't had a chance to check out if there are any concurrecy performance-killers with this update. My sense is that the early exits will avoid those but I'll double-check that tomorrow.

src/overlap_check/ProgressBar.cpp
```diff
@@ -0,0 +1,68 @@
...
+  return isatty(STDOUT_FILENO) != 0;
+#else
+  return false;
+#endif
+}
+
+
+void ProgressBar::set_value(double val) {
+
+  if (!is_terminal()) {
```
gonuke, line 35: Couldn't this check be done in the constructor?  It won't change during the run, right?

gonuke: Obviously, you'd still need to check something for an early exit, but instead of running `isatty()` each time, you could just check a boolean member variable.

gonuke:
I think we might be [bikeshedding](https://whatis.techtarget.com/definition/Parkinsons-law-of-triviality-bikeshedding) this PR - so I'm ready to merge once @pshriwise has reviewed for concurrency performance killers.

src/overlap_check/ProgressBar.cpp
pshriwise: No appreciable difference in performance that I could see!

gonuke:
Thanks @pshriwise

pshriwise:
@ljacobson64 I noticed that the Ubuntu 18.04 build wasn't finding OpenMP, but the Ubuntu 16.04 build was (with gcc). From what I can tell, the `dagmc_setup_flags` function was causing it to fail. Specifically, it seemed that calling `set(CMAKE_CXX_STANDARD 11)` before `find_packace(OpenMP)` was causing the problem. As we only apply the OpenMP flags in the `src/overlap_check/` directory I think this should be ok. What do you say?

pshriwise:
Thanks for the merge @gonuke! If we determine we should handle finding OpenMP differently, I'll make a new PR.

ljacobson64:
Just noticed that this PR still used a `build` directory to build the executable instead of the `app` directory as was made standard in #645

pshriwise:
I'll add that change as well.

ljacobson64:
I'll look into the OpenMP thing. I have seen some strange behavior with having CMake try to find OpenMP on Ubuntu 18.04 as well

pshriwise:
Ok I'll wait to hear about that. In the meantime, I'll get that directory name updated. Meant to do so and caught caught up in all the `ProgressBar` updates.

ljacobson64:
It found openmp fine on my machine with Ubuntu 18.04. The CI found it too, on both 16.04 and 18.04 with gcc. It did not find it with clang though. My machine does not find it with clang either, unless I add these lines to the cmake command: 
```
-DOpenMP_C_LIB_NAMES=gomp
-DOpenMP_CXX_LIB_NAMES=gomp
```

pshriwise:
Strange. I thought I saw it fail to OpenMP  with `gcc` in [this job](https://travis-ci.org/svalinn/DAGMC/jobs/600530087). And FWIW I tried a simple project on my own machine with 19.04 and saw the same behavior.

ljacobson64:
Oh, I see what you mean. I only looked at the Travis logs from the latest commit, which is after you moved the `find_package` call. This is very odd

Issue #642, opened 2019-10-05
`DAGMC_LIBRARIES` CMake variable incorrectly populated


pshriwise:
**Describe the Bug**<br/>
Currently, all libraries are added to the `DAGMC_LIBRARIES` CMake variable whether they are enabled or not. This is caused by [these lines](https://github.com/svalinn/DAGMC/blob/8b23e5fd12dfa59e490a661a054b92722db56f18/cmake/DAGMC_macros.cmake#L225). As these macros get evaluated and inserted into the various directories where it is called, the `DAGMC_LIBRARIES` variable is populated regardless of whether or not the library is enabled or disabled. [`add_subdirectory`](https://cmake.org/cmake/help/v3.0/command/add_subdirectory.html) will keep targets from being added to the project, but the subdirectory `CMakeLists.txt` file will still be processed.

**To Reproduce**<br/>
Clone the repo and run `cmake <dagmc_src_dir> -DBUILD_TALLY=OFF`. You will still see `dagtally` in the list of DAGMC libraries in the `DAGMCConfig.cmake` file in the directory where you've run this command.

**Expected Behavior**<br/>
Only enabled libraries are added to the `DAGMC_LIBRARIES` variable.

pshriwise:
@ljacobson64 this is your area of expertise. Any suggestions on how we might update the macro to fix this? We could place these additions to `DAGMC_LIBRARIES` inside of if statements for the different libraries in the `/src/CMakeLists.txt` but there might be a more elegant solution?

gonuke:
I think this is addressed in #634 which is stalled trying to sort out whether MOAB's build system is self-consistent across different options.   At this point, I think a CMake build of MOAB will give correct settings and indications for our DAGMC build.

pshriwise:
Oh I thought #634 was about managing `MOAB_LIBRARIES`. This issue is entirely internal to DAGMC.

ljacobson64:
I was not able to reproduce this behavior. I built DAGMC with `-DBUILD_TALLY=OFF` and this line appeared in `DAGMCConfig.cmake`:
```
set(DAGMC_LIBRARIES MOAB dagmc pyne_dagmc uwuw makeWatertight)
```
`dagtally` is not present.

pshriwise:
Hrmm this might be caused by something other than what I initially thought.

Can you running (in the same build directory):

```bash

$ cmake ../dagmc_src -DMOAB_DIR=...
[...]

PR #643, opened 2019-10-08
Fix DAGMC_LIBRARIES cmake variable

ljacobson64:
Addressing #642


cmake/DAGMC_macros.cmake
```diff
@@ -222,11 +225,7 @@ macro (dagmc_install_library lib_name)
   endif ()
 
   # Keep a list of all libraries being installed
-  if (DAGMC_LIBRARIES)
-    set(DAGMC_LIBRARIES "${DAGMC_LIBRARIES} ${lib_name}" CACHE INTERNAL "DAGMC_LIBRARIES")
-  else ()
-    set(DAGMC_LIBRARIES ${lib_name} CACHE INTERNAL "DAGMC_LIBRARIES")
-  endif ()
+  set(DAGMC_LIBRARIES ${DAGMC_LIBRARIES} ${lib_name} CACHE INTERNAL "DAGMC_LIBRARIES")
```
pshriwise, line 228:
```suggestion
  set(DAGMC_LIBRARIES "${DAGMC_LIBRARIES} ${lib_name}" CACHE INTERNAL "DAGMC_LIBRARIES")
```

pshriwise: The current version gives me a semicolon-delimited list. The change above makes it a whitespace deliniated list. 

If I remember right, I've heard complaints about getting this kind of list in a configuration file because they can cause problems when used in different ways externally.  

A: `target_link_libraries(foo_lib ${DAGMC_LIBRARIES}`
B: `target_link_libraries(foo_lib "${DAGMC_LIBRARIES")`

If we write a semicolon-delimited list, case B will fail as the variable won't be expanded but handed verbatim to the command. But if we use a whitespace delineated list, both will be fine. This line of thought changes when working with other entries in a list that might have spaces (like filenames or paths), but for a list of library names I think we're ok. Of course, it isn't a problem if external applications use the variable correctly, but I don't think it hurts us to accommodate here.

That's my understanding anyway and I'm happy to defer to you on it!

pshriwise (changes requested):
Nice! Does the trick for me. Just one question about semicolons in the variable.

cmake/DAGMC_macros.cmake
pshriwise: Any thoughts on this either way @ljacobson64?

ljacobson64: I guess my experience is that I've had a lot more trouble dealing with lists formatted as strings than I have with lists formatted as lists

pshriwise: Good enough for me!

pshriwise (approved):

PR #657, opened 2020-01-13
consolidate all the PR news

ngranda:
Removes the PR specific news files and adds their content into one news file. 
Authors of those PRs that were not in the Authors file were added


ngranda:
Hoping to start a conversation about a DAGMC release with this PR

gonuke:
It makes sense to me to keep the individual news files, even as their content is consolidated into a main news file.

Alternatively, each of the entries in the consolidated file could indicate which PR it came from.

E.g.
```
* something new was added or a bug fixed (PR#987)
```

gonuke:
(Unless someone has an argument against this strategy?)

bam241:
thx @ngranda 
merging

ljacobson64:
So what is the new practice now? Continue to add PR-XXXX.rst files for every PR, and every time we do a release we delete all the news files and combine them into a single file? Perhaps the combined news file should have the version number in its filename?

AUTHORS.rst
```diff
@@ -4,6 +4,10 @@ Authors
 The DAGMC project began as the CAD-MCNPX project in 2003, and since then has
 included contributions from the following:
 
+* Paul Romano
+* Jonathan Shimwell
+* Chelsea D'Angelo
+* Xiaokang Zhang
```
ljacobson64, line 10: It seems like the rest of the names are in alphabetical order. I feel we should continue that practice

PR #665, opened 2020-03-27
Apply code review changes to MCNP files

kkiesling:
Applying the changes from the professional code review (#660) to the MCNP related files. The changes are as follows:
- cleaning up commented lines in cmake files ([original comment](https://github.com/svalinn/DAGMC/pull/660#discussion_r368664837))
- changes error messages to `std::cerr` instead of `std::cout` ([original comment](https://github.com/svalinn/DAGMC/pull/660#discussion_r368664840))
- move strings for Graveyard and Vacuum into variables ([original comment](https://github.com/svalinn/DAGMC/pull/660#discussion_r368664844))
- uses C++11 standards for converting int/doubles to strings ([original comment](https://github.com/svalinn/DAGMC/pull/660#discussion_r368664849))

The one comment that I did NOT address was this comment on the readability of using `#ifdef`s in function calls (see [here](https://github.com/svalinn/DAGMC/pull/660#discussion_r399540089) for more information on why). It's doable, just will be done in a separate PR since it will likely be a larger more significant change.


src/mcnp/mcnp_funcs.cpp
```diff
@@ -251,33 +254,33 @@ void write_cell_cards(std::ostringstream& lcadfile, const char* mcnp_version_maj
...
         // no splitting can happenin vacuum set to 1
-      } else if (mat_name.find("Vacuum") != std::string::npos) {
+      } else if (mat_name.find(vacuum_str) != std::string::npos) {
         imp = 1.0;
         // otherwise as the map says
       } else {
         imp = DMD->importance_map[entity][particle_name];
       }
-      importances += "imp:" + mcnp_name + "=" + _to_string(imp) + " ";
+      importances += "imp:" + mcnp_name + "=" + std::to_string(imp) + " ";
```
pshriwise: It seems that `std::to_string`  is always going to supply 6 decimal places for floating-point types, even if they aren't needed. The test output of `dagmcnp_unit_tests` expects the lcad output to have integer values for the neutron/photon importances and it's causing CI to fail.

We can either change the expected test output or we could get these values as integers before writing them to the lcad file. Given that MCNP treats these importances as floating-point values, I'd say we should update the expected results in our test suite. I'll defer to you and @makeclean on this though as you're both far more knowledgable than I :)

pshriwise (commented):
These look like positive changes to me @kkiesling. Thanks for getting us started on incorporating the PullRequest changes.

src/mcnp/mcnp_funcs.cpp
kkiesling: [mcnp_funcs.cpp actually defines the importances as doubles.](https://github.com/svalinn/DAGMC/blob/develop/src/mcnp/mcnp_funcs.cpp#L257) We _could_ define it as an int . If the input to `to_string()` is an int, then there's no trailing zeros. But this is probably not recommended because users can define an importance as a doubles.

I am going to update the tests to expect doubles there.

pshriwise: Didn't mean that `mcnp_funcs.cpp` defines the importance as an integer.. just that the expected test output has something that looks like an int. :) In the end, the important part is how MCNP interprets them when determining how we should write them IMO, so I back your call to update the expected test output to expect doubles.

kkiesling: Right- Sorry that was my own processing of what was wrong because I was confused by the trailing zeros (I assumed the importances were defined as ints, but doubles makes sense). Anyway, tests have been updated.

src/mcnp/tests/dagmcnp_unit_tests.cpp
```diff
@@ -113,7 +113,7 @@ TEST_F(DAGMCNP5Test, dagmcinit_test) {
     EXPECT_EQ(expected_lcad[i], input_deck[i]);
   }
   // delete the lcad file
-  std::remove("lcad");
+  //std::remove("lcad");
```
gonuke: Why this change?  Should we just delete it?  or is this an artifact of confirming the change?

gonuke (changes requested):
Thanks for making these changes @kkiesling - just one question.

src/mcnp/mcnp_funcs.cpp
pshriwise: Nice! Only caveat here is that importances will now be truncated to 6 digits... something we should probably be aware of. The [example here](https://devdocs.io/cpp/string/basic_string/to_string) explains in more detail. 

After reading a bit on this subject, it seems that our method in `mcnp_funds.cpp` for higher precision double --> string conversion (with the addition of a `std::setprecision` in the stream) is still recommended. :man_shrugging:

I'd be surprised if more than 6 digits are used for these values very often, so I'm good w/ whatever you decide.

src/mcnp/tests/dagmcnp_unit_tests.cpp
```diff
@@ -76,17 +76,17 @@ TEST_F(DAGMCNP5Test, dagmclcad_test) {
...
-                            "3 9 -12.0 imp:n=1 imp:p=1 ",
-                            "4 9 -12.0 imp:n=1 imp:p=1 ",
-                            "5 3 -2.0 imp:n=1 imp:p=1 ",
-                            "6 3 -2.0 imp:n=1 imp:p=1 ",
-                            "7 3 -2.0 imp:n=1 imp:p=1 ",
-                            "8 3 -2.0 imp:n=1 imp:p=1 ",
-                            "9 1 0.022 imp:n=1 imp:p=1 ",
-                            "12 0  imp:n=0 imp:p=0   $ graveyard",
-                            "13 0  imp:n=1 imp:p=1   $ implicit complement"
+  const char* expected[] = {"1 9 -12.0 imp:n=1.000000 imp:p=1.000000 ",
```
pshriwise: This isn't a comment from PullRequest, but I think it's worth updating and is in the spirit of this PR.

With C++11 we can now define these expected values using an initializer list rather than the variable-length char array we're using now like so:

```c++
std::vector<std::string> expected_lcad = {"1 9 -12.0 imp:n=1.000000 imp:p=1.000000 ",                                                                 
                                                                  "2 9 -12.0 imp:n=1.000000 imp:p=1.000000 ",
...
```

Happens in a few places here I think.

src/mcnp/tests/dagmcnp_unit_tests.cpp
kkiesling: Oops. I accidentally left that in when I was looking at the written file. I'll fix that.

src/mcnp/tests/dagmcnp_unit_tests.cpp
kkiesling: Good idea. I'll search throughout and update.

src/mcnp/mcnp_funcs.cpp
kkiesling: I'll look into the `std::setprecision` option. If we set the precision to match doubles, that's 15 decimal points.

The other thing I just noticed given your example is that `to_string()` won't write out values in scientific notation. So if a value is 1e-3, then it will be written as 0.00100. If a value is smaller than 1e-6, then it will return 0.000000. So that concerns me too. If we set precision to 15, that allows for numbers as small as 1e-15. That doesn't seem like a good idea to limit the range like that.

What do you think? Now I am thinking that given the limitations of this `to_string()` function that we should actually keep what we had.

kkiesling: @gonuke or @makeclean - any thoughts on this?

gonuke: I agree that the `to_string()` is not a good change

pshriwise: I agree as well -- we should go back to our `_to_string` function for floating point values. We can still use `std::to_string` for `int`'s like material number though. Probably best to use the standard library where we can.

kkiesling: This sounds like a good idea- I will update accordingly.

I think `to_string()` will still convert ints to floats to write out (eg, `1` will become `1.000000` by default). But I might be able to play with the precision to avoid that.

pshriwise: You'll find that `std::to_string()` will write ints without trailing zeros. Its implementation is specialized depending on the type passed to it. So no extra work should be needed there :tada:

src/mcnp/mcnp_funcs.cpp
```diff
@@ -198,37 +201,37 @@ void write_cell_cards(std::ostringstream& lcadfile, const char* mcnp_version_maj
...
+          std::cerr << "In the material library" << std::endl;
           exit(EXIT_FAILURE);
         }
 
         pyne::Material material = workflow_data->material_library[mat_name];
         int matnumber = material.metadata["mat_number"].asInt();
-        mat_num = _to_string(matnumber);
-        density = "-" + _to_string(material.density);
+        mat_num = std::to_string(matnumber);
+        density = "-" + std::to_string(material.density);
```
gonuke: should we also use `_to_string()` here?

kkiesling: yes- I haven't gone back through to untangle all the `to_string()`/`_to_string()` uses yet

kkiesling:
I think this is ready now as long as the checks pass.

src/mcnp/mcnp5/CMakeLists.txt
```diff
@@ -52,7 +52,6 @@ set(CMAKE_Fortran_FLAGS_RELWITHDEBINFO "-O1 -g")
 
 # C compiler flags
 if (CMAKE_C_COMPILER_ID STREQUAL "Intel")
-  #set(CMAKE_C_FLAGS "${CMAKE_Fortran_FLAGS} -mcmodel=medium")
```
ljacobson64, line 55: The reason these commented lines are here is that they appear in the makefiles for native MCNP, but I needed to turn them off. I don't remember why.

pshriwise (approved):
Thanks for getting us started on this @kkiesling!

src/mcnp/mcnp5/CMakeLists.txt
kkiesling: Should they be left in and put in use? Or is it okay to remove these lines?

ljacobson64: Ehhhh yeah let's just delete them

gonuke (approved):

gonuke:
Thanks for kicking it off @kkiesling

PR #666, opened 2020-04-01
Apply code review changes to make_watertight files

pshriwise:
Applying the changes from the professional code review (#660) to the `make_watertight` related files. The changes are as follows:

  - remove commented code blocks that are either outdated or are debug statements
  - improvements to some logic for clarity
  - use of standard library containers to avoid potential memory leaks in Arc.cpp/Gen.cpp
  - improvements to struct/variable names
  - declared variables for "magic numbers"
  - passing by const reference where possible to avoid unnecessary memory allocation
  - removed an unused function (`Arc::create_loops_from_oriented_edges_fast`)

I'm not going to link each comment from the original PR here due to the sheer number of them, but I'm happy to answer any questions about the changes and hunt them down for reference as needed.

### DIFF
--- a/src/make_watertight/Arc.cpp
+++ b/src/make_watertight/Arc.cpp
@@ -125,7 +125,8 @@ moab::ErrorCode Arc::remove_opposite_pairs_of_edges_fast(moab::Range& edges, con
 
   // populate edge array, used only for searching
   unsigned n_orig_edges = edges.size();
-  edge* my_edges = new edge[n_orig_edges];
+  std::vector<Edge> my_edges(n_orig_edges);
+  // edge* my_edges = new edge[n_orig_edges];
   unsigned j = 0;
   for (moab::Range::const_iterator i = edges.begin(); i != edges.end(); ++i) {
     // get the endpoints of the edge
@@ -151,7 +152,7 @@ moab::ErrorCode Arc::remove_opposite_pairs_of_edges_fast(moab::Range& edges, con
   }
 
   // sort edge array
-  qsort(my_edges, n_orig_edges, sizeof(struct edge), compare_edge);
+  qsort(my_edges.data(), n_orig_edges, sizeof(struct Edge), compare_edge);
 
   // find duplicate edges
   j = 0;
@@ -171,7 +172,6 @@ moab::ErrorCode Arc::remove_opposite_pairs_of_edges_fast(moab::Range& edges, con
       edges = subtract(edges, duplicate_edges);
       rval = MBI()->delete_entities(duplicate_edges);
       if (moab::MB_SUCCESS != rval) {
-        delete[] my_edges;
         MB_CHK_SET_ERR(moab::MB_FAILURE, "cannot delete edge");
       }
       if (debug) {
@@ -183,7 +183,6 @@ moab::ErrorCode Arc::remove_opposite_pairs_of_edges_fast(moab::Range& edges, con
     j = i;
   }
 
-  delete[] my_edges;
   return moab::MB_SUCCESS;
 }
 
@@ -257,7 +256,7 @@ moab::ErrorCode Arc::create_loops_from_oriented_edges_fast(moab::Range edges,
                                                            std::vector< std::vector<moab::EntityHandle> >& loops_of_edges,
                                                            const bool debug) {
   // place all edges in map
-  std::multimap<moab::EntityHandle, edge> my_edges;
+  std::multimap<moab::EntityHandle, Edge> my_edges;
   moab::ErrorCode rval;
   for (moab::Range::const_iterator i = edges.begin(); i != edges.end(); ++i) {
     // get the endpoints of the edge
@@ -268,11 +267,11 @@ moab::ErrorCode Arc::create_loops_from_oriented_edges_fast(moab::Range edges,
       MB_CHK_SET_ERR(moab::MB_FAILURE, "could not get connectivity");
     }
     // store the edges
-    edge temp;
+    Edge temp;
     temp.edge = *i;
     temp.v0   = endpts[0];
     temp.v1   = endpts[1];
-    my_edges.insert(std::pair<moab::EntityHandle, edge>(temp.v0, temp));
+    my_edges.insert(std::pair<moab::EntityHandle, Edge>(temp.v0, temp));
   }
   std::cout << "error: function not complete" << std::endl;
   return moab::MB_FAILURE;
--- a/src/make_watertight/Arc.cpp
+++ b/src/make_watertight/Arc.cpp
@@ -126,7 +126,6 @@ moab::ErrorCode Arc::remove_opposite_pairs_of_edges_fast(moab::Range& edges, con
   // populate edge array, used only for searching
   unsigned n_orig_edges = edges.size();
   std::vector<Edge> my_edges(n_orig_edges);
-  // edge* my_edges = new edge[n_orig_edges];
   unsigned j = 0;
   for (moab::Range::const_iterator i = edges.begin(); i != edges.end(); ++i) {
     // get the endpoints of the edge
--- a/src/make_watertight/Arc.cpp
+++ b/src/make_watertight/Arc.cpp
@@ -217,7 +217,6 @@ moab::ErrorCode Arc::get_next_oriented_edge(const moab::Range edges,
       std::cout << "result=" << result
                 << " could not get connectivity of edge" << std::endl;
       return result;
-      //print_edge( *i );
     }
     assert(moab::MB_SUCCESS == result);
     assert(2 == n_verts);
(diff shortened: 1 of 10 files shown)


src/make_watertight/CMakeLists.txt
```diff
@@ -1,4 +1,3 @@
-message("")
```
ljacobson64: Apologies in advance for commenting on such a small change.

The purpose of these lines is to break up the CMake output to make it easier to see what CMake output is coming from where. This `message("")` line appears in CMakeLists.txt files throughout the entire repo. If we don't want these lines anymore, then we should delete them from all the CMake files, not just this one. (I would argue we shouldn't delete them.)

pshriwise: Ah, makes sense. Though I think there's something to their comment that it would be even better if these each contained a message indicating their respective directories.
https://github.com/svalinn/DAGMC/pull/660/files#r368664884

All the same, reverting this change now.

src/make_watertight/Arc.cpp
```diff
@@ -664,7 +637,7 @@ moab::ErrorCode Arc::merge_curves(moab::Range curve_sets, const double facet_tol
         result = zip->merge_verts(curve_i_verts.back(), curve_j_verts.back(),
                                   curve_i_verts, curve_j_verts);
         if (moab::MB_SUCCESS != result)
-          std::cout << result << std::endl;
+          std::cout << "Failed to merge vertices with error: " << result << std::endl;
```
kkiesling: Should this be `std::cerr`?

kkiesling (commented):
Nice work! Just a few questions:

There are a lot of `assert(moab::MB_SUCCESS == result)` lines throughout (not necessarily ones that you've added). We discussed that `result`/`rval` should always be checked, but maybe there needs to be more than just an assertion. If the result is not a success, what error message shows up and is it helpful to users at all? Some places are good about having messages, while others still don't.

src/make_watertight/Arc.cpp
pshriwise: Mmm yeah probably. Updating now.

pshriwise:
I had the same thought looking through this and it's been an issue in the `make_watertight` files for a while now, but I felt that that would be better off in a separate PR dedicated solely for that update. for now, I've created issue #669 for this so we don't forget.

kkiesling (approved):
This PR looks good to me now. Thanks @pshriwise!

pshriwise:
Any further comments or thoughts here @ljacobson64 @kkiesling?

kkiesling:
> Any further comments or thoughts here @ljacobson64 @kkiesling?

None from me. I approved the other day.

src/make_watertight/Arc.hpp
```diff
@@ -73,7 +70,7 @@ class Arc {
 
   /// goes through curve_sets and finds any curves with coincident ( dist. apart <= FACET_TOL) front and back points.
```
bam241:
```suggestion
  /// goes through curve_sets and finds any curves with coincident ( dist. apart <= facet_tol) front and back points.
```

src/make_watertight/Cleanup.cpp
```diff
@@ -152,7 +151,7 @@ moab::ErrorCode Cleanup::delete_small_edges(const moab::Range& surfaces, const d
         MBI()->list_entities(duplicate_edges);
       assert(1 == duplicate_edges.size());
 
-      // if the edge length is less than MERGE_TOL do nothing
+      // if the edge length is less than FACET_TOL do nothing
```
bam241, line 154:
```suggestion
      // if the edge length is less than facet_tol do nothing
```

is not this variable lower case now ?

src/make_watertight/Cleanup.cpp
```diff
@@ -225,13 +219,13 @@ moab::ErrorCode Cleanup::delete_small_edges(const moab::Range& surfaces, const d
...
-      if (adj_skin_edges1.empty())
-        endpt1_is_skin = false;
-      else
-        endpt1_is_skin = true;
-      if (endpt0_is_skin && endpt1_is_skin)
+
+      // check to see if the endpoints part of the skin
+      bool endpt0_is_skin = !adj_skin_edges0.empty();
+      bool endpt1_is_skin = !adj_skin_edges1.empty();
+      if (endpt0_is_skin && endpt1_is_skin) {
```
bam241, line 226:
```suggestion
      if (!adj_skin_edges1.empty() && !adj_skin_edges0.empty()) {
```

src/make_watertight/Gen.cpp
```diff
@@ -427,15 +433,14 @@ double Gen::length(std::vector<moab::EntityHandle> edges) {
...
   double dist = 0;
   moab::EntityType type = MBI()->type_from_handle(edges[0]);
 
   // if vector has both edges and verts, only use edges
   // NOTE: The curve sets from ReadCGM do not contain duplicate endpoints for loops!
   moab::EntityType end_type = MBI()->type_from_handle(edges.back());
   if (type != end_type) {
-    for (std::vector<moab::EntityHandle>::iterator i = edges.begin(); i != edges.end(); i++) {
+    for (auto i = edges.begin(); i != edges.end(); i++) {
       if (moab::MBVERTEX == MBI()->type_from_handle(*i)) {
```
bam241, line 444:
```suggestion
      if (MBI()->type_from_handle(*i) == moab::MBVERTEX) {
```
We are testing here if `MBI()->type_from_handle(*i)` is a `moab::MBVERTEX`. For me it make more sense to have the tested object on the left side....

bam241 (commented):
Thx @pshriwise !
Looking good.

I only have few minor comments.

src/make_watertight/MakeWatertight.cpp
```diff
@@ -1525,7 +1520,7 @@ moab::ErrorCode MakeWatertight::delete_merged_curves(moab::Range& existing_curve
   result = MBI()->delete_entities(curves_to_delete);
   assert(moab::MB_SUCCESS == result);
   if (result != moab::MB_SUCCESS) {
-    std::cout << "Houston, we have a problem" << std::endl;
+    std::cout << "Failed to delete curves" << std::endl;
```
bam241: should it be a `cerr` as well ?

src/make_watertight/Arc.cpp
```diff
@@ -125,7 +125,8 @@ moab::ErrorCode Arc::remove_opposite_pairs_of_edges_fast(moab::Range& edges, con
 
   // populate edge array, used only for searching
   unsigned n_orig_edges = edges.size();
-  edge* my_edges = new edge[n_orig_edges];
+  std::vector<Edge> my_edges(n_orig_edges);
+  // edge* my_edges = new edge[n_orig_edges];
```
bam241: why adding this as a comment ? As you removed all the other ?

bam241 (commented):
I missed those 2 :)

src/make_watertight/Arc.cpp
pshriwise: Definitely forgot to pull that line. Thanks!

src/make_watertight/Cleanup.cpp
pshriwise: Ummm it should be IMO, but the line below still uses the uppercase `FACET_TOL` passed into this function.

I think it'd be a worthwhile PR correcting that variable name in the future though.

pshriwise: Or maybe I'm missing the point of your comment?

bam241: you made that change L73 (old l76), in `Arc.hpp`

I was just trying to maintain consistency... (I actually missed there were not in the same file....)

BTW I don't have a strong opinion about this....

src/make_watertight/Cleanup.cpp
pshriwise: Had this thought as well... I left those variables in there for clarity though. The fact that the `Range` `adj_skin_edges1` is empty doesn't quite get across the point that the endpoint of the edges is a skin point like the Boolean variable names do. I can update it if you're feeling strongly about it though.

src/make_watertight/MakeWatertight.cpp
pshriwise: Ah yeah, is should be.

src/make_watertight/Cleanup.cpp
bam241: I don't have a string feeling either....

src/make_watertight/Arc.hpp
pshriwise: Ah here it is, the source of the `FACET_TOL` --> `facet_tol` comments. I'm going to revert this change for now so they can be dealt with all together at another time.

bam241 (approved):
Looking good thx @pshriwise !

gonuke:
Thanks @pshriwise

PR #668, opened 2020-04-02
Addressing docker comments

bam241:
This address comments from #660 about Docker.

I added something from me as well, see if I can separate the housekeeping jobs from the others... and that way simply the travis CI file


bam241:
@kkiesling @pshriwise @gonuke this is ready for review.

docker/build_geant4.sh
```diff
@@ -1,13 +1,13 @@
...
-set -e
+set -ex
 
 source ${docker_env}
 
 rm -rf ${geant4_build_dir}/bld ${geant4_install_dir}
 mkdir -p ${geant4_build_dir}/bld
 cd ${geant4_build_dir}
-wget http://geant4.cern.ch/support/source/geant4.${geant4_version}.tar.gz
+wget https://geant4.cern.ch/support/source/geant4.${geant4_version}.tar.gz --no-check-certificate
```
pshriwise, line 10: In #660 they mention verifying our file against a checksum. Where do we stand on adding that?

pshriwise (commented):
Just one small comment from me. Nice cleanup of the housekeeping check!

docker/build_geant4.sh
bam241: did not found any checksum for geant4 tarball :(

docker/Dockerfile
```diff
@@ -2,30 +2,18 @@ ARG UBUNTU_VERSION=18.04
...
-    else \
-      astyle_deb=astyle_3.1-1ubuntu2_amd64.deb && \
-      wget https://launchpad.net/ubuntu/+source/astyle/3.1-1ubuntu2/+build/14532685/+files/${astyle_deb} && \
-      dpkg -i ${astyle_deb} && \
-      rm -f ${astyle_deb}; \
-    fi
-
 # Get some pip packages
-RUN pip install cython sphinx
+RUN pip install cython
```
ljacobson64: Why was sphinx moved out of this file?

docker/Dockerfile
```diff
@@ -2,30 +2,18 @@ ARG UBUNTU_VERSION=18.04
...
 RUN apt-get -y update
-RUN apt-get -y upgrade
-RUN apt-get -y dist-upgrade
 
 # Get some apt packages
 RUN apt-get -y install autoconf clang cmake g++ gcc gfortran git libblas-dev \
                        libhdf5-dev liblapack-dev libpython2.7-dev libtool \
                        python-numpy python-pip python-setuptools wget
 
-# Get astyle
```
ljacobson64, line 17: Why was astyle moved out of this file? Everything that could possibly mess with system libraries (i.e. apt or pip) was put in this file.

docker/build_dagmc.sh
```diff
@@ -13,7 +13,6 @@ fi
 rm -rf ${dagmc_build_dir}/bld ${dagmc_install_dir}
 mkdir -p ${dagmc_build_dir}/bld
 cd ${dagmc_build_dir}
-#git clone https://github.com/svalinn/DAGMC -b develop
```
ljacobson64, line 16: The reason these lines were commented is that they would normally be part of a DAGMC build script, but in this specific case, we get the DAGMC source code by copying it into the docker image, not cloning it from github.

docker/build_dagmc.sh
```diff
@@ -29,4 +28,3 @@ cmake ../src -DMOAB_DIR=${moab_install_dir} \
 make -j${jobs}
 make install
 cd
-#rm -rf ${dagmc_build_dir}
```
ljacobson64, line 32: The dagmc_build_dir would normally be deleted after building DAGMC, but we need to keep it in this specific case in order to run the tests.

.travis.yml
```diff
@@ -66,17 +34,49 @@ before_install:
...
+    - env: UBUNTU_VERSION=16.04 COMPILER=gcc   HDF5_VERSION=1.10.4 MOAB_VERSION=develop
+      if: type != pull_request
+    - env: UBUNTU_VERSION=16.04 COMPILER=clang HDF5_VERSION=1.10.4 MOAB_VERSION=develop
+      if: type != pull_request
+    - env: UBUNTU_VERSION=18.04 COMPILER=gcc   HDF5_VERSION=1.10.4 MOAB_VERSION=develop
+      if: type != pull_request
+    - env: UBUNTU_VERSION=18.04 COMPILER=clang HDF5_VERSION=1.10.4 MOAB_VERSION=develop
+      if: type != pull_request
+    - stage: housekeeping
+      env: UBUNTU_VERSION=16.04 COMPILER=gcc   HDF5_VERSION=1.10.4 MOAB_VERSION=5.1.0
```
ljacobson64: The housekeeping run should use 18.04

docker/Dockerfile
```diff
@@ -2,30 +2,18 @@ ARG UBUNTU_VERSION=18.04
...
-RUN if [ "${UBUNTU_VERSION}" == "18.04" ]; then \
```
ljacobson64, line 18: Removing these lines will disallow the ability to do the housekeeping run on Ubuntu 16.04. This is fine, as long as the housekeeping run is changed back to being on 18.04.

ljacobson64 (changes requested):

.travis.yml
bam241: oups this was not intended

docker/build_dagmc.sh
kkiesling: @ljacobson64 is there any reason to leave this line in as a comment w/ justification or can we just remove it all together? (see [here](https://github.com/svalinn/DAGMC/pull/660#discussion_r368664684))

docker/build_dagmc.sh
kkiesling: My same question above applies here too- do you think it should be left in as a comment w/ an explanation or removed?

docker/Dockerfile
kkiesling: It looks like anything having to do w/ housekeeping was moved to the housekeeping.sh file.

kkiesling (approved):
I have no major comments. Thanks @bam241!

bam241:
@ljacobson64 to avoid having the housekeeping stuff everywhere and only used for 1 job in ubuntu 18.04. I moved `astyle` and `sphinx` install in the housekeeping scripts.

docker/build_dagmc.sh
bam241: @ljacobson64 

agreed with @kkiesling if this is commented without explanation it should not be part of the build.
I'd be happy to add them back with a line of explanation. 

when you mean run the tests, do you mean run the test locally ? 
Because we run the test in CI without those...

docker/build_dagmc.sh
bam241: If I understand correctly, there are there just because the normal install `DAGMC` procedure should include them. But in this case as we copy the `DAGMC` from the PR (using `Travis`) we don't need them.

If this is correct, I don't think they should be there:
Those scripts are dedicated for CI not to show people how to install `DAGMC`.
It is probably a good idea to have a version of them somewhere to facilitate `DAGMC` install but for me this is definitely not the place.

kkiesling: I agree w/ this assessment @bam241. I think the lines here should be removed and if we want a docker to help users, we need to host that elsewhere.

docker/build_geant4.sh
gonuke: We could generate a checksum against the version we are pulling

docker/travis.housekeeping.sh
```diff
@@ -18,6 +18,9 @@ if [ "${TRAVIS_REPO_SLUG}" == "svalinn/DAGMC" ] && \
   fi
 fi
 
+# install package for style check
+apt-get install astyle
```
gonuke: Why are we moving this to here?

docker/travis.tests.sh
```diff
@@ -19,7 +19,7 @@ cd ${dagmc_build_dir}/bld
 # If this is not a pull request, get files needed for regression tests
 if [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then
   cd src/make_watertight/tests
-  wget ${MW_REG_TEST_MODELS_URL} -O mw_reg_test_files.tar.gz -o wget.out
+  wget ${MW_REG_TEST_MODELS_URL} -O mw_reg_test_files.tar.gz
```
gonuke, line 22: The comment suggested streaming these to `stderr` and `stdout` - do we do this now?  Or is that default behavior?

gonuke:
I'm ambivalent on the housekeeping changes.  They may happen fairly often, which means a lot of `apt install` and `pip installing` for stuff we could do once.  Should we have different docker images? Would that help?

docker/build_geant4.sh
bam241: make sense !

docker/travis.tests.sh
bam241: my understanding of the comment was the default behavior streams them to `stderr`

docker/build_dagmc.sh
ljacobson64: For maintainability purposes, I think it would be good to have some indication of why the source code already exists in this directory. (The current situation with just a commented `git clone` line is also not very helpful.) Perhaps a comment line that says

```
# DAGMC source code already exists in this directory because it was copied into
# the docker image by travis.yml
```

docker/build_dagmc.sh
ljacobson64: I think this line can be deleted.

To answer @bam241's question: I do mean for running the tests in CI. The CI runs `make test` in the DAGMC build directory, so the DAGMC build directory must exist in order to run the tests. See here: https://github.com/bam241/DAGMC/blob/fix_docker/docker/travis.tests.sh#L28

docker/build_dagmc.sh
bam241: I think I understand,

the line is left there to highlight that we don't need to copy DAGMC source, while we normally would in a standard installation process.
So the line is basically there for educational purposes...
Is that correct @ljacobson64 ?

I would definitely prefer to not mix purposes in those scripts. So if there are used for CI install and testing, I would remove the git clone line and add the comment at the head of the file.

bam241:
I was thinking @ljacobson64 @kkiesling @pshriwise @gonuke 
For separation of concern what do you think about:
- creating a `.travis_scripts` folder containing all the scripts required for CI (exept the `update_docker.sh` one)
- keep only the `Dockerfile` and the `update_docker.sh` macro in the `docker` folder
- create a `script` folder, with a version of the install scripts dedicated to installation for users ?

ljacobson64:
That kind of sounds like overkill. I'd rather just delete those 2 lines we were discussing

bam241:
> That kind of sounds like overkill. I'd rather just delete those 2 lines we were discussing

Yes but I feel like it will clarify the purpose of everything.
It is just that it does not seems obvious to me, that the scripts used to install and run the tests in CI would be located in the `docker` folder

ljacobson64:
Would it make more sense if the `docker` folder were renamed `CI`?

bam241:
@gonuke @kkiesling @ljacobson64 this should  be ready for an around of review

bam241:
@ljacobson64 @kkiesling I think I am now satisfied with the way it looks. 

cc @pshriwise @gonuke

gonuke:
Should we test some of this by adding an automatic build to dockerhub?

ljacobson64:
How does that work?

gonuke:
You point dockerhub at a Dockerfile in a github repo and everytime that repo changes, it rebuilds the image.  We have deployed it for 
* https://hub.docker.com/repository/docker/svalinn/pymoab-py2-18.04
* https://hub.docker.com/repository/docker/svalinn/pymoab-py2-16.04
* https://hub.docker.com/repository/docker/svalinn/pymoab-visit-py2-18.04
and it seems to work well.

So far they have all been for images whose configuration is 100% defined in the dockerfile, so I don't know how it will behave for this image that requires scripts...

bam241:
@gonuke this seems like a good idea, will take a look at this

gonuke:
I have something running here: https://hub.docker.com/repository/docker/svalinn/dagmc-ci-ubuntu-18.04/builds

gonuke:
The dockerhub build is taking a long time....  probably not surprising

gonuke:
Looks like it might have timed out...

bam241:
@gonuek if we want to do the automated one, we will probably need 2 dockerfiles one for each ubuntu version

ljacobson64:
We could stop supporting 16.04 since it's 4 years old now

ljacobson64:
Although in that case we may as well start supporting 20.04...

bam241:
@gonuke I think we will have problem....
timeout limits is set to 2h
https://success.docker.com/article/what-are-the-current-resource-limits-placed-on-automated-builds

bam241:
I think the main problem is building geant4... that takes forever....

gonuke:
And we build Geant4 four different times!  It times out on the first try.

gonuke:
This may be a good reason to separate out some of the docker `RUN` lines, and build Geant4 before HDF5.  It is less likely to change in ways that matter, and we can do that locally, push the image and then let the autobuild skip the parts that are already built.

That said, perhaps we make a bunch of different images, each tagged with the info we we want and let dockerhub build them all?

TL;DR: if we want to leverage autobuilds on dockerhub, perhaps we need a different strategy.

bam241:
Few though there and comments:
- the `hooks/build` strategy seems to have worked to allow multiple build using a single `Dockerfile` with different `ENV_VARIABLE`
- I am not sure how the 2h limits is used, is it 2h limits per job (i.e. `hooks/build` script), per image build or per RUN command....
- it seems that the `ci/dockercloud` hook replaced our `CI/CircleCI` hook in the test check :(, not sure if we can have both into the same repo... will look into it...

bam241:
I am not sure what to do with this..., the dockeer build process is wait to long...

my suggestion would be to leave this on the side, maybe make a issue about this and merge this PR....

bam241:
I am not sure how, but travis is back :)

bam241:
discovered a bug in the existing `Dockerfile`, took me a couple of hours to fix it.
I am building and pushing fixed version of the docker container. 
I'll do my "final" push on this PR after :)

bam241:
@gonuke dockerhub auto-build restriction to 2h runs, will force us to redesign our workflow concerning docker image generation and usage in CI. 
(I tried to push the image before hand but still takes a super long time on dockerhub...)
After discussing with @pshriwise we concluded that it is probably better to keep this PR simple, make an issue about dockerhub auto-generation, and redesign docker image workflow in a different PR.


@ljacobson64 @kkiesling is there anything blocking still this PR for you ? (I'll be happy to discuss/tackle those points if any remain.)

I'd like to keep this moving... @pshriwise's PR #674 is waiting for this to be finalized and merged.

gonuke:
I think what I want to see is that the docker image generated by this process will result in an image in which testing is successful.  If you already pushed a new image based on this to dockerhub, maybe that already happened when the most recent commits were tested on Travis??  can you confirm?

bam241:
Yes I pushed my last commit after having build and push the new docker images on `dockerhub`.

so the last set of Travis test were run on docker images generated with this docker image

bam241:
todo:
- [x] move housekeeping the the beginning
- [x] stop CI if one test fail

docker/Dockerfile
bam241: I re-added those line for ubuntu 18.04 only as housekeeping is only done there.

docker/Dockerfile
bam241: I put it back in :)

bam241:
@pshriwise @gonuke 
- I added a fast fail, which will stop CI jobs if one fails. (I try to with a DagMC in which I messed up the formatting, it fails and stoped everything [see](https://travis-ci.org/github/svalinn/DAGMC/builds/687020418))
- I reput housekeeping as the first jobs, so it should run first.

bam241:
@pshriwise @gonuke 
- I added a fast fail, which will stop CI jobs if one fails. (I try to with a DagMC in which I messed up the formatting, it fails and stoped everything [see](https://travis-ci.org/github/svalinn/DAGMC/builds/687020418))
- I reput housekeeping as the first job, so it should run first.


@ljacobson64 I think I addressed all your comments, could you confirm ??

scripts/build_geant4.sh
```diff
@@ -1,17 +1,22 @@
...
 
 source ${docker_env}
 
 rm -rf ${geant4_build_dir}/bld ${geant4_install_dir}
 mkdir -p ${geant4_build_dir}/bld
 cd ${geant4_build_dir}
-wget http://geant4.cern.ch/support/source/geant4.${geant4_version}.tar.gz
+wget https://geant4.cern.ch/support/source/geant4.${geant4_version}.tar.gz --no-check-certificate
+tar_chashum=$(sha256sum ${geant4_version}.tar.gz | cut -d' ' -f1)
+if [ $geant4_shasum != $tar_chashum ]; then
```
ljacobson64: If we're doing the shasum thing for Geant4, would it make sense to do it for HDF5 too?

.travis.yml
```diff
@@ -7,44 +7,12 @@ notifications:
...
-    - env: UBUNTU_VERSION=16.04 HOUSEKEEPING_ONLY=false COMPILER=clang HDF5_VERSION=1.10.4 MOAB_VERSION=develop
-    - env: UBUNTU_VERSION=18.04 HOUSEKEEPING_ONLY=false COMPILER=gcc   HDF5_VERSION=1.10.4 MOAB_VERSION=develop
-    - env: UBUNTU_VERSION=18.04 HOUSEKEEPING_ONLY=false COMPILER=clang HDF5_VERSION=1.10.4 MOAB_VERSION=develop
+
 before_install:
   # Set some local variables
   - docker_image=svalinn/dagmc-ci-ubuntu-${UBUNTU_VERSION}
   - dagmc_build_dir=/root/build/${COMPILER}/DAGMC-moab-${MOAB_VERSION}-hdf5-${HDF5_VERSION}
-  - script_dir=${dagmc_build_dir}/DAGMC/docker
+  - script_dir=${dagmc_build_dir}/DAGMC/scripts
```
ljacobson64: I don't think I understand why the scripts need to be in a separate directory from the Dockerfile. All of the scripts in this directory are CI-specific scripts that are only intended to be used in conjunction with the docker image.

ljacobson64:
One final comment: I think it would be a good idea to rebase this PR against `develop` before merging. There have been several PRs merged since this one was first submitted, so we want to make sure these new PRs also play nice with the new CI.

scripts/build_geant4.sh
bam241: fair point !

.travis.yml
bam241: I think I got confuse because some of the file where `travis.xxx` and some just `build` I though that the build one could be used elsewhere.
@ljacobson64 What about we put all of them in a CI folder as you suggested (including the `Dockerfile`) and add a `docker.` prefix in the build scripts that are intended to be used in the docker building process.

ljacobson64: I like that idea. So there'd be a folder called `CI` and it would contain these files

```
docker.build.dagmc.sh
docker.build.geant4.sh
docker.build.hdf5.sh
docker.build.moab.sh
docker.env.sh
Dockerfile
travis.housekeeping.sh
travis.install.sh
travis.tests.sh
update_docker.sh
```

bam241: build_dagmc is only used in travis, so I named it `travis.`
still working on making sure update_docker still work before pushing it

bam241: `env.sh` is used in both, so I kept it without a prefix

bam241:
@ljacobson64 I added the suggestion about check_sum for hdf5 (and the moving/renaming of the scripts and Dockerfile)

the hdf5 shasum made me think about the version we are using, as when we use an other version than the system one, we only use the `1.10.4` shall we note specify it in the `env.sh` instead of having a environment variable that we set "manually" ?

.travis.yml
gonuke: Sorry to be late here, but why `CI/travis.xxx` and `CI/docker.xxxx` and not `CI/travis/xxxx` and `CI/docker/xxxx`?

CI/docker.build_hdf5.sh
```diff
@@ -13,11 +13,17 @@ fi
 rm -rf ${hdf5_build_dir}/bld ${hdf5_install_dir}
 mkdir -p ${hdf5_build_dir}/bld
 cd ${hdf5_build_dir}
+
 wget https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-${HDF5_VERSION_major}/hdf5-${HDF5_VERSION}/src/hdf5-${HDF5_VERSION}.tar.gz
+tar_chashum=$(sha256sum hdf5-${HDF5_VERSION}.tar.gz | cut -d' ' -f1)
+if [ $hdf5_shasum != $tar_chashum ]; then
```
ljacobson64: `$hdf5_shasum` should be `$hdf5_1_10_4_shasum`

bam241:
@ljacobson64 @gonuke I this I addressed your last comments :)

gonuke (approved):
Thanks @bam241

bam241:
@ljacobson64 do you have any other comments ?

CI/update_docker.sh
```diff
@@ -1,10 +1,10 @@
 #!/bin/bash
 
-set -e
+set -ex
 
 ubuntu_versions="16.04 18.04"
 for ubuntu_version in ${ubuntu_versions}; do
   image_name="svalinn/dagmc-ci-ubuntu-${ubuntu_version}:latest"
-  docker build -t ${image_name} --build-arg UBUNTU_VERSION=${ubuntu_version} .
+  docker build -t ${image_name} --build-arg UBUNTU_VERSION=${ubuntu_version} -f CI/Dockerfile .
```
ljacobson64, line 8: Is this script now intended to be run from outside the `CI` directory?

bam241: I though it would make more sense to run from the top directory

ljacobson64 (approved):

ljacobson64:
Has the docker image on dockerhub already been updated? I'm curious about how the tests are passing if it hasn't. Unless the new docker image is the same as the old one despite having a new Dockerfile.

bam241:
Yes the docker container have been updated/pushed after last dockerfile change.

gonuke (approved):
Thank @bam241 - LGTM

PR #671, opened 2020-04-09
Change pointer to shared_ptr

bam241:
addressing comment from #660:
```
Rather than manually `new`'ing and `deleting` these objects, I would recommend 
using a RAII ("Resource Allocation Is Initialization") technique to manage their 
memory [1].  Specifically, you can use smart pointers [2] such as `std::unique_ptr` 
or `std::shared_ptr` to allocate `GTT` and `GQT` in the constructor, and subsequently 
not be responsible for their deletion.

Unless you are allocating a computationally large number of pointers (such as in a 
core data structure in a game engine), it's typically recommended to use a smart 
pointer when possible to guard against memory leaks.

[1] https://en.cppreference.com/w/cpp/language/raii
[2] https://en.cppreference.com/book/intro/smart_pointers
```
_Originally posted by @PullRequest-Agent in https://github.com/svalinn/DAGMC/pull/660/threads/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMzY4Njc1MDIw/resolve_

Targeting declaration and deletion of pointers in **DagMC()** (see DAGMC.cpp L64)

Not sure we want to change the API (constructor of **DagMC** class) but I was not able to have **shared_ptr<>** in the body of the **DagMC** class and still accept a normal pointer for the **Interface*** because when declaring the point outside of the **DagMC** instance, one need to delete it, but the ownership of the **Interface*** has been transferred to the **shared_ptr<>** which is suppose to take care of the deallocation.
But it is not obvious without looking at the `DagMC.cpp` source code that you should not deallocate your own pointer.
More over if the **Interface*** object would probably been deallocated after deleting the **DagMC** object which will be problematic if you want to used it ( **Interface***) after the **DagMC** deletion....

-------
I created a separate PR for this one, as it will be simpler to review, and I think it might need some discussion about how to address this.

To make that change I was not able to find an other implementation, so I see only 2 solutions:
- keep the code as it is with `pointer` and **delete**
- slightly change the API


pshriwise:
These are great changes IMO! As you noted, @bam241, requiring use of a `std::shared_ptr` for the MOAB interface makes DAGMC a little more rigid. 

One thought might be to have a raw pointer to a `moab::Interface` in the DAGMC class which is set regardless of the constructor used and be returned by the `DAGMC::moab_instance()` method. Our constructors taking a raw `moab::Interface` pointer would then look like this:

```cpp
DagMC::DagMC(Interface* mb_impl, double overlap_tolerance, double p_numerical_precision) {
  // if we arent handed a moab instance, create one
  if (nullptr == mb_impl) {
   mb_shared = std::shared_ptr<Core>();
   MBI = mb_shared.get();
  } else {
   MBI = mb_impl;
  }
  
  // make new GeomTopoTool and GeomQueryTool
  ...
}
```

If the pointer is passed in, we won't delete it on cleanup but if we did create it, the `std::shared_ptr` will take care of itself when the `DagMC` object is destroyed. No `new`'s /`delete`'s to worry about on our end.

Does that sound plausible or am I missing something?

Clarification:

In the example above `MBI` is still our raw `moab::Interface*` and we would add an attribute called `mb_shared` or something similar for the `shared_ptr`.

pshriwise:
Adding to the comment above, if we go this route I think it would be good for us to include a constructor that takes in a `std::shared_ptr<Interface>` and start a discussion about marking the other constructor as deprecated or indicate some kind of preference in docs/comments for the constructor taking a `shared_ptr`. IMO we want to promote the use of these modern C++ practices when possible.

src/dagmc/DagMC.cpp
```diff
@@ -39,36 +39,32 @@ const bool counting = false; /* controls counts of ray casts and pt_in_vols */
...
 
 // DagMC Constructor
-DagMC::DagMC(Interface* mb_impl, double overlap_tolerance, double p_numerical_precision) {
+DagMC::DagMC(std::shared_ptr<Interface> mb_impl, double overlap_tolerance, double p_numerical_precision) {
   moab_instance_created = false;
   // if we arent handed a moab instance create one
-  if (NULL == mb_impl) {
-    mb_impl = new moab::Core();
+  if (nullptr == mb_impl) {
+    mb_impl = std::shared_ptr<Interface>(new moab::Core());
```
pshriwise:
```suggestion
    mb_impl = std::shared_ptr<Core>();
```

src/dagmc/DagMC.cpp
```diff
@@ -39,36 +39,32 @@ const bool counting = false; /* controls counts of ray casts and pt_in_vols */
...
   }
 
   // set the internal moab pointer
-  MBI = mb_impl;
+  MBI = std::shared_ptr<Interface>(mb_impl);
 
   // make new GeomTopoTool and GeomQueryTool
-  GTT = new moab::GeomTopoTool(MBI, false);
-  GQT = new moab::GeomQueryTool(GTT, overlap_tolerance, p_numerical_precision);
+  GTT = std::shared_ptr<GeomTopoTool> (new GeomTopoTool(MBI.get(), false));
```
pshriwise: Should these be `unique_ptr`'s as those objects are going to carry state that is related directly to calls that have been made on this DagMC object (e.g. `init_obbs`)?

src/dagmc/tests/dagmc_simple_test.cpp
```diff
@@ -29,26 +29,24 @@ TEST_F(DagmcSimpleTest, dagmc_load_file) {
 TEST_F(DagmcSimpleTest, dagmc_load_file_dagmc) {
   /* 1 - Test with external moab, load file in DAGMC*/
   // make new moab core
-  Core* mbi = new moab::Core();
+  std::shared_ptr<Core> mbi = std::shared_ptr<Core>(new moab::Core());
```
pshriwise:
```suggestion
  std::shared_ptr<Interface> mbi = std::make_shared<Core>();
```

src/dagmc/tests/dagmc_simple_test.cpp
```diff
@@ -29,26 +29,24 @@ TEST_F(DagmcSimpleTest, dagmc_load_file) {
...
 
   ErrorCode rval;
+
   // load a file
   rval = dagmc->load_file(input_file);
   EXPECT_EQ(rval, MB_SUCCESS);
-
-  // delete dagmc
   delete dagmc;
-  delete mbi;
```
pshriwise, line 43: :tada:

src/dagmc/tests/dagmc_simple_test.cpp
```diff
@@ -29,26 +29,24 @@ TEST_F(DagmcSimpleTest, dagmc_load_file) {
...
 }
 
 TEST_F(DagmcSimpleTest, dagmc_load_file_dagmc_via_moab) {
   /* 2 - Test with external moab, load file in MOAB*/
   // load the file into moab rather than dagmc
   ErrorCode rval;
 
-  moab::Core* mbi = new moab::Core();
+  std::shared_ptr<Core> mbi = std::shared_ptr<Core>(new moab::Core());
```
pshriwise:
```suggestion
  std::shared_ptr<Interface> mbi = std::make_shared<Core>();
```

src/dagmc/tests/dagmc_simple_test.cpp
```diff
@@ -77,7 +74,7 @@ TEST_F(DagmcSimpleTest, dagmc_load_file_dagmc_build_obb) {
   // make new moab core
   ErrorCode rval;
 
-  moab::Core* mbi = new moab::Core();
+  std::shared_ptr<Core> mbi = std::shared_ptr<Core>(new moab::Core());
```
pshriwise:
```suggestion
  std::shared_ptr<Interface> mbi = std::make_shared<Core>();
```

src/dagmc/tests/dagmc_simple_test.cpp
```diff
@@ -88,15 +85,14 @@ TEST_F(DagmcSimpleTest, dagmc_load_file_dagmc_build_obb) {
...
-  delete mbi;
 }
 
 TEST_F(DagmcSimpleTest, dagmc_load_file_dagmc_via_moab_build_obb) {
   /* 2 - Test with external moab, load file in MOAB*/
   // load the file into moab rather than dagmc
   ErrorCode rval;
 
-  moab::Core* mbi = new moab::Core();
+  std::shared_ptr<Core> mbi = std::shared_ptr<Core>(new moab::Core());
```
pshriwise:
```suggestion
  std::shared_ptr<Interface> mbi = std::make_shared<Core>();
```

src/dagmc/DagMC.cpp
pshriwise:
```suggestion
  GTT = std::make_shared<GeomTopoTool>(MBI.get(), false);
```

pshriwise (changes requested):
Thanks for bringing us up-to-date on this front @bam241! A few suggestions on creating the `std::shared_ptr`'s.

src/dagmc/tests/dagmc_simple_test.cpp
```diff
@@ -88,15 +85,14 @@ TEST_F(DagmcSimpleTest, dagmc_load_file_dagmc_build_obb) {
...
   rval = mbi->load_file(input_file);
   EXPECT_EQ(rval, MB_SUCCESS);
   moab::DagMC* dagmc = new moab::DagMC(mbi);
```
pshriwise: Missed this in my initial review, but we'll probably want to update these raw pointers to smart pointers as well.

bam241:
> Adding to the comment above, if we go this route I think it would be good for us to include a constructor that takes in a `std::shared_ptr<Interface>` and start a discussion about marking the other constructor as deprecated or indicate some kind of preference in docs/comments for the constructor taking a `shared_ptr`. IMO we want to promote the use of these modern C++ practices when possible.

the problem with the deprecated constructor, is that coexistence of a pointer and a shared_ptr pointing to the same object will potentially lead to segfault.

I am not an expert about shared_ptr and unique_ptr, but this is what I observed so far...

src/dagmc/tests/dagmc_simple_test.cpp
bam241: I am ok with that suggestion. I just like to understand why sometimes do we use `Core` and some other time `Instance` ?

pshriwise:
True, but that's also currently true right?

If someone were to call for the the `moab::Interface*` using `DagMC::moab_instance`, they could delete the pointer and cause the same problem whether we're using a `shared_ptr` or not.

My understanding is that use of `shared_ptr`/`unique_ptr` isn't intended to protect against that source of segfaults, but more to simplify the cleanup of these objects and avoid memory leaks.

bam241:
> True, but that's also currently true right?
> 
> If someone were to call for the the `moab::Interface*` using `DagMC::moab_instance`, they could delete the pointer and cause the same problem whether we're using a `shared_ptr` or not.
> 
> My understanding is that use of `shared_ptr`/`unique_ptr` isn't intended to protect against that source of segfaults, but more to simplify the cleanup of these objects and avoid memory leaks.

I don't believe it is the case with my change as the pointer is owned by the `share_ptr`

so you don't have to delete the pointer.

The probleme with have a contractor taking a Instance*, is that the ownership of the pointer if transferred to the share_ptr (MBI) that might delete it. which would make an issue for the initial pointer passed to the constructor:
```
Instance* my_exeternal_MBI;

DagMC* dagmc = new DagMC(my_external_MBI);

delete dagmc;
delete my_external_MBI
```

the problem with such constructor is that in it, we do something like:
`MBI = shared_ptr<Core>(my_external_MBI)`


when deleting the `dagmc` object this would also delete the `MBI` (shared_ptr), as it is a member of the `dagmc` object which would delete the location of `my_external_MBI`

but `my_external_MBI` outside of the dagmc object still point to this address.

src/dagmc/tests/dagmc_simple_test.cpp
pshriwise: We talked about this offline, but I'll put it here for future reference. 

This works b/c `moab::Core` inherits from `moab::Interface`. The compiler is ok with us assigning the `shared_ptr` of a child class to a `shared_ptr` of a parent class, but this would not work the other way around.

bam241:
this last comment re-add the constructor with a **Instance*** as an argument. This constructor is deprecated and only exist for backward compatibility.

the `DagMC` class will have 2 members for MBI, one `Instance* MBI` and a `share_ptr<Instance> MBI_shared_ptr` to avoid segaulting when deallocating the `DagMC` instance.

Indeed, having a constructor with a `Interface*` argument will induce pointer owner ship issue.
in the `Instance*` is directly casted into the `shared_ptr` taking the ownership on the pointer then at destruction of the `DagMC` object the `shared_ptr` will also be delete and deallocating the `Interface*` address... which may potentially lead to segfault.

With this construct the following case are possible:
- external `Interface*` (using the deprecated constructor) -> no shared_ptr is defined, the Instance* is owned outside of the DagMC object, so no need to delete it.
```
  Interface* mbi = new Interface();
  DagMC *dagmc = new DagMC(mbi);
  delete dagmc;
  delete mbi;
```


- external `shared_ptr< Interface >`, the chain of shared_ptr owns the address of the Instance and will deal with the deallocation when needed. (no need to delete the Interface*)
```
  shared_ptr<Interface> mbi = make_shared<Interface>();
  DagMC *dagmc = new DagMC(mbi);
  delete dagmc;
```

- internal definition of the Interface, the Interface* is defined and owned through the the `shared_ptr< Interface>` that will take care of its deallocation when needed...
```
  DagMC *dagmc = new DagMC();
  delete dagmc;
```

pshriwise:
>I don't believe it is the case with my change as the pointer is owned by the share_ptr
>
>so you don't have to delete the pointer.
>
>The probleme with have a contractor taking a Instance*, is that the ownership of the pointer if transferred to the share_ptr (MBI) that might delete it. which would make an issue for the initial pointer passed to the constructor:

Seems like we weren't quit talking about the same thing here (confusion about whether `MBI` was a shared pointer or a raw pointer...), but I'm glad to see the changes we talked about are going to work out! :rocket:

Thanks for putting in the effort to keep the original constructor. I see why you were asking about deleting the mesh in the destructor -- it'd be nice to get rid of that `moab_instance_created` variable. Before these changes, we would only delete the mesh right before deleting the instance which seems odd. @gonuke @makeclean can you remember if there was a specific reason for that?

bam241:
@gonuke @pshriwise @makeclean this should be ready for an other round of review

src/dagmc/DagMC.cpp
```diff
@@ -39,36 +39,46 @@ const bool counting = false; /* controls counts of ray casts and pt_in_vols */
...
   moab_instance_created = false;
   // if we arent handed a moab instance create one
-  if (NULL == mb_impl) {
-    mb_impl = new moab::Core();
+  if (nullptr == mb_impl) {
+    mb_impl = std::make_shared<Core>();
     moab_instance_created = true;
   }
 
+  MBI_shared_ptr = std::shared_ptr<Interface>(mb_impl);
```
pshriwise: These should be equivalent.
```suggestion
  MBI_shared_ptr = mb_impl;
```

src/dagmc/DagMC.hpp
```diff
@@ -365,7 +369,9 @@ class DagMC {
  public:
   OrientedBoxTreeTool* obb_tree() {return GTT->obb_tree();}
 
-  GeomTopoTool* geom_tool() {return GTT;}
+  [[ deprecated("Replaced by std::shared_ptr<GeomTopoTool> geom_tool_ptr()") ]]
+  GeomTopoTool* geom_tool() {return GTT.get();}
+  std::shared_ptr<GeomTopoTool> geom_tool_ptr() {return GTT;}
```
pshriwise: We need to be careful about passing back shared pointers for these objects. In the case where we've created the MOAB instance, these objects aren't guaranteed to be useable beyond the life of the DAGMC object. 

```c++
// populates the MBI_shared_ptr and sets MBI
std::unique_ptr<DagMC> dag = std::make_unique<DagMC>();
moab::Errorcode rval = dag->load_file("test.h5m");
MB_CHK_ERR(rval);

rval = dag->init_OBBTree();
MB_CHK_ERR(rval);

// retrieve a shared_ptr GTT from the DagMC instance
std::shared_ptr<GeomTopoTool> gtt = dag->geom_tool_ptr();

// MBI_shared_ptr is cleaned up and the MOAB instance is destroyed as intended
dag.reset();

// the GTT shared_ptr is still available, but any call that relies on it's underlying MBI pointer will fail
rval = gtt->check_model(); // will segfault
MB_CHK_ERR(rval);
```

I think it would be better to keep only the `geom_tool()` method for now.

src/dagmc/DagMC.hpp
```diff
@@ -381,17 +387,23 @@ class DagMC {
   ErrorCode get_root(EntityHandle vol_or_surf, EntityHandle& root);
 
   /** Get the instance of MOAB used by functions in this file. */
+  [[ deprecated("Replaced by std::shared_ptr<Interface> moab_instance_ptr()") ]]
   Interface* moab_instance() {return MBI;}
+  std::shared_ptr<Interface> moab_instance_ptr() {return MBI_shared_ptr;}
```
pshriwise: This method is also a little inconsistent with our current design. What if the deprecated constructor is used and this object is a `nullptr`?

pshriwise (changes requested):
A couple of thoughts on returning shared pointers from the DagMC interface. I think we're going to run into trouble making sure they don't cause problems at this point. We'll get there someday though hopefully.

src/dagmc/DagMC.hpp
bam241: sounds good !

src/dagmc/DagMC.hpp
bam241: it make sense the only `shared_ptr` we shall return is the MBI one....
the other one depends on each other through the raw pointer not the shared pointer, so we can't guaranty they will be working outside of the DagMC instance....

pshriwise:
This is really close I think. Any luck with making the `GQT` and `GTT` unique pointers?

bam241:
> This is really close I think. Any luck with making the `GQT` and `GTT` unique pointers?

Sadly not, `GQT` is an argument of `GTT` so it won't be possible.

I have a error message with the `GTT` that I don't quite understand, I'll try again an post the msg...

(when I googled it, the easy work around was to use a `shared_ptr` :P )

pshriwise (approved):
Thanks again @bam241 for taking this on! It's a much-needed update. Hopefully we can start to convert some of our other classes to these more modern practices in the future.

bam241:
@pshriwise @gonuke @makeclean 
Looking back at this, we have a return method for the `MBI_shared_ptr`. 
it will be a `nullptr` if the MBI instance has been passed by the user as a raw_ptr and populated otherwise.

Do we want to throw a warning if someone if trying to get it and it is not defined, or is returning a `nullptr` enough ?

pshriwise:
I'd prefer there be some kind of warning/error here. If we're returning an empty shared pointer and know it, it seems as though we should give some kind of indication about that. Honestly, I think we should throw a runtime error, but I'd take a warning too :)

src/dagmc/DagMC.hpp
```diff
@@ -382,16 +387,25 @@ class DagMC {
 
   /** Get the instance of MOAB used by functions in this file. */
   Interface* moab_instance() {return MBI;}
+  std::shared_ptr<Interface> moab_instance_sptr() {
+    if (nullptr == MBI_shared_ptr)
+      std::runtime_error("MBI instance is not defined as a shared pointer !");
+    return MBI_shared_ptr;
+  }
```
bam241, line 394: Added a `runtime_error` when trying to return the `MBI` instance as a `shared_ptr` when it has been provided as a direct pointer. (It has been implemented in the Header, as it is fairly simple, I can move it to the `cpp` file if you prefer)

bam241 (commented):
@svalinn/dagmc I am now satisfied with this.

I updated the news to reflect our discussion.

pshriwise (approved):

bam241:
A remaining identified issue (that existed before) is if returning `GTT` instance while having the `MBI` instance instantiated internally, if one delete the `DagMC` object, the `MBI` pointer will be deallocated... leaving the `GTT` pointing to a void pointer...

This would also occur before this change tho... (Does not mean we should not fix it, but might be outside of the scope of this PR...)

pshriwise:
Agreed @bam241. Let's make an issue for that and move forward for now. I'll merge this today if there are no other comments/concerns.

pshriwise:
Thanks again for tackling this @bam241!

Issue #673, opened 2020-04-24
Shared library testing on CI and Geant4


pshriwise:
So I noticed recently that:

  1. We don't do testing on executables linked to our shared libraries in CI 
  2. When shared testing is enabled, the dagsolid_unit_tests test fails to find some of the libraries it needs and fails on Ubuntu 18.04 only

After some digging, I found that this is because in our 16.04 builds the `RPATH` property of shared libraries and executables is set whereas in 18.04 the `RUNPATH` property is set. This difference is expected as the linking program in Ubuntu, `ld`, moved to using `RUNPATH` in 18.04. (Some links on the difference between these properties and their effects on how libraries are dynamically loaded are at the bottom of this long-winded issue.)

The important difference here is that the `RUNPATH` property won't dynamically load transitive libraries while `RPATH` will. None of the Geant4 libraries have their own `RPATH`/`RUNPATH` properties set -- at least for the installation we rely on in CI anyway. As a result, the libraries, executables, and tests related to Geant4 fail find their dependencies at runtime. Specifically, the problem we run into first is that we link to the `libG5geometry.so` library which in turn links to the `libG4graphics_reps.so` library. These two are in the same location, but because of the `RUNPATH` limitation `libG4graphics_rps.so` isn't found.

I see a couple of options to solve this problem:

__Option 1:__

The Geant4 team seems to be aware that a user's environment needs to be modified to work correctly with their installation and provide a `geant4.sh` in the installation `bin` directory to fix that issue. If we want to count on Geant4 user experience, we can expect that they'll know to source this file before using any of the executables/libraries we produce and things should work just fine. Then we can source this file as part of our test environment or add the Geant4 `lib` directory to the `LD_LIBRARY_PATH` for the test and call it good.

__Option 2:__

Ensure that `RPATH` is set instead of `RUNPATH`. This can be done by adding `--disable-new-dtags` to the CMake shared linker flags in the `/src/geant4/` directory which should affect only the targets build in that space.

I'm leaning toward option 1 and have solution in place, but I'd thought I'd explain the problem here and see what others think.

Info on `RPATH` vs. `RUNPATH`:

  - https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/RPATH-handling
  - https://medium.com/nataraj-raghavendra/rpath-vs-runpath-883029b17c45

ljacobson64:
I suspect this issue can be solved by adding

[...]

PR #674, opened 2020-04-24
Shared library/executable testing in CI and update to the Geant4 RPATH

pshriwise:
## Description
This PR adds shared library testing to our CI. The CI tests will fail at first for reasons described in #673, but once the update to `build_geant4.sh` is applied we should see them succeed. I'll wait to update the docker image until someone has had a look and give me the :+1: to do so.

To enable the shared build testing, I've moved a lot of the actual building of DAGMC into `travis.tests.sh`, so our files 

One other small change I've added is the `CTEST_OUTPUT_ON_FAILURE` flag to provide more information in the CI log if a test does fail.

## Motivation and Context
We should be testing both static and shared builds of DAGMC on CI. We have plenty of spare time and resources on Travis.

### DIFF
--- a/CI/docker/build_geant4.sh
+++ b/CI/docker/build_geant4.sh
@@ -21,6 +21,7 @@ cmake ../geant4.${geant4_version} -DBUILD_STATIC_LIBS=ON \
              -DCMAKE_BUILD_TYPE=Release \
              -DCMAKE_C_COMPILER=${CC} \
              -DCMAKE_CXX_COMPILER=${CXX} \
+             -DCMAKE_INSTALL_RPATH=${geant4_install_dir}/lib \
              -DCMAKE_INSTALL_PREFIX=${geant4_install_dir}
 make -j${jobs}
 make install
--- a/CI/env.sh
+++ b/CI/env.sh
@@ -8,6 +8,8 @@ export geant4_shasum=4b05b4f7d50945459f8dbe287333b9efb772bd23d50920630d5454ec570
 export build_dir=/root/build/${COMPILER}
 export install_dir=/root/opt/${COMPILER}
 
+export CTEST_OUTPUT_ON_FAILURE=1
+
 if [ "${HDF5_VERSION}" == "system" ]; then
   export hdf5_build_dir=
   export hdf5_install_dir=/usr/lib/x86_64-linux-gnu/hdf5/serial
--- a/CI/env.sh
+++ b/CI/env.sh
@@ -30,6 +30,13 @@ export moab_install_dir=${install_dir}/moab-${MOAB_VERSION}-hdf5-${HDF5_VERSION}
 export dagmc_build_dir=${build_dir}/DAGMC-moab-${MOAB_VERSION}-hdf5-${HDF5_VERSION}
 export dagmc_install_dir=${install_dir}/DAGMC-moab-${MOAB_VERSION}-hdf5-${HDF5_VERSION}
 
+export dagmc_build_dir_shared=${dagmc_build_dir}-shared
+export dagmc_install_dir_shared=${dagmc_install_dir}-shared
+
+export dagmc_build_dir_static=${dagmc_build_dir}-static
+export dagmc_install_dir_static=${dagmc_install_dir}-static
+
+
 if [ "$COMPILER" == "gcc" ]; then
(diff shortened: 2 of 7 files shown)


pshriwise:
If there are no objections, I'm going to go head and update the CI Docker image tomorrow evening with the changes in `build_geant4.sh` from this PR.

gonuke:
Do you know why it hasn't even tried to run tests on this branch?

gonuke:
I think this should probably be coordinated with #668 - I'm not sure which should come first?

pshriwise:
@gonuke It looks like they did run, but the results aren't appearing here for some reason. :thinking: 

https://travis-ci.org/github/svalinn/DAGMC/builds/679251870

I'll chat with @bam241 tomorrow about how best to proceed with our two CI-related PRs.

pshriwise:
@bam241 and I settled on getting #668 in first. I'll update this one on top of those changes and update CI when appropriate.

In the meantime I'll see what I can do about getting the CI test results to appear here.

pshriwise:
Thanks again for tackling this @bam241!

gonuke:
This looks OK to me. We'll need to push a new docker image once this is merged.

bam241:
with this change you basically remove everything form the `install.sh` and move it into the `test.sh`
which almost remove the need of the `instal`l step in our `travis` workflow.

I am really not sure this is the best solution but:
Would it be possible to have both static and shared coexist in the same container, and just swap environment variable, so installs stays in the `install.sh` script and test in the `test.sh` ?

bam241:
> with this change you basically remove everything form the `install.sh` and move it into the `test.sh`
> which almost remove the need of the `instal`l step in our `travis` workflow.
> 
> I am really not sure this is the best solution but:
> Would it be possible to have both static and shared coexist in the same container, and just swap environment variable, so installs stays in the `install.sh` script and test in the `test.sh` ?

we could also separate STATIC/SHARED build in our travis jobs, but I feel like we already have too many...

pshriwise:
Fair point @bam241. I've updated the `install.sh` script to install and test both the static and shared versions of DAGMC in different locations. Was this about what you had in mind?

bam241:
yes it was an option.
we could had more granularity in the travis building process also... 
have job for static and job for shared (if they don't need to coexist....)

pshriwise:
Very true. I guess I agreed with your sentiment that we have enough CI jobs as it is :)

bam241:
we lost CI here :(

bam241:
CI is back... this is strange

pshriwise:
Should be running now.

pshriwise:
I'm thinking the next step here is to update the CI Docker image, make sure it passes, and then merge. 

Does that sound good @bam241 @gonuke?

gonuke:
I've lost the train of thought here a little, but is the current failing test because we don't have a shared object build of G4 on the image?  Why does that only affect U18.04/gcc?

pshriwise:
Correct. That only affects 18.04 for the reasons described in #673.

ljacobson64:
There is a shared object build of Geant4 in the docker image. Even though the [cmake command](https://github.com/svalinn/DAGMC/blob/develop/CI/docker/build_geant4.sh#L19) contains `-DBUILD_STATIC_LIBS=ON` and does not contain anything about building shared libraries, shared libraries are still built by default. I believe the tests would pass if Pat's change [here](https://github.com/svalinn/DAGMC/pull/674/files#diff-53211a5f6056e7acab03b692b7088c3cR24) were incorporated into the docker image.

bam241:
I think, the new docker container should be pushed, to make sure the failing test is passing
and then merge this PR once everything is working well.

Other PRs will probably not be sensitive to those docker changes anyway....

pshriwise:
So shall I go ahead and push the new image via `update_docker.sh` then?

bam241:
> So shall I go ahead and push the new image via `update_docker.sh` then?

I vote yes !

pshriwise:
Shockingly, everything seems to have gone according to plan. Let me know if there are any other adjustments I should make.

bam241 (approved):
this looks great !

will merge this afternoon unless someone comments.

gonuke:
Thanks @pshriwise

PR #676, opened 2020-05-13
Addressing most of the comments about src/dagmc files

bam241:
This addresses most of the comments about dagmc sources from #660 .

I was not sure how to fix some of them so I ping @pshriwise @gonuke @makeclean on them to get more information.

I also included changes suggested by my linter (mostly line <80 chars...)

### DIFF
--- a/src/dagmc/DagMC.cpp
+++ b/src/dagmc/DagMC.cpp
@@ -399,7 +399,7 @@ ErrorCode DagMC::build_indices(Range& surfs, Range& vols) {
   EntityHandle tmp_offset = 0;
 
   if (surfs.size() == 0 || vols.size() == 0) {
-    std::cout << "Volumes or Surfaces founds" << std::endl;
+    std::cout << "Volumes or Surfaces not found" << std::endl;
     return  MB_ENTITY_NOT_FOUND;
   }
   setOffset = std::min(*surfs.begin(), *vols.begin());
--- a/src/dagmc/DagMC.cpp
+++ b/src/dagmc/DagMC.cpp
@@ -94,38 +94,43 @@ float DagMC::version(std::string* version_string) {
 
 unsigned int DagMC::interface_revision() {
   unsigned int result = 0;
+  std::string key_str = "$Rev:";
+  int max_length = 10;
+
   std::string interface_string = DAGMC_INTERFACE_REVISION;
-  if (interface_string.rfind("$Rev:") != std::string::npos) {
+  if (interface_string.rfind(key_str) != std::string::npos) {
     // start looking for the revision number after "$Rev: "
     char* endptr = NULL;
     errno = 0;
-    result = strtol(interface_string.c_str() + 5, &endptr, 10);
-    if (endptr == interface_string.c_str() + 5 ||
-        ((result == LONG_MAX || result == LONG_MIN) && errno == ERANGE)) {
+    result = strtol(interface_string.c_str() + key_str.size(), &endptr,
+                    max_length);
+    // check if the string has been fully parsed and the results is within
+    // normal range
+    if (endptr == interface_string.c_str() + key_str.size()  // parsing end
+        || ((result == LONG_MAX || result == LONG_MIN)
+            && errno == ERANGE)) {  // checking range
       std::cerr << "Not able to parse revision number" << std::endl;
       exit(1);
     }
     return result;
   }
-  /* SECTION I: Geometry Initialization and problem setup */
+  return result;
 }
 
 // the standard DAGMC load file method
 ErrorCode DagMC::load_file(const char* cfile) {
   ErrorCode rval;
+  std::string filename(cfile);
   std::cout << "Loading file " << cfile << std::endl;
   // load options
   char options[120] = {0};
-  char file_ext[4] = "" ; // file extension
+  std::string file_ext = "" ; // file extension
 
   // get the last 4 chars of file .i.e .h5m .sat etc
-  if (sizeof(cfile) / sizeof(char) > 4) {
-    memcpy(file_ext, &cfile[strlen(cfile) - 4], 4);
-  } else {
-    std::cerr << "DagMC warning: unhandled file "
-              << "loading options." << std::endl;
+  int file_extension_size = 4;
+  if (filename.size() > file_extension_size) {
+    file_ext = filename.substr(filename.size() - file_extension_size);
   }
-
   EntityHandle file_set;
   rval = MBI->create_meshset(MESHSET_SET, file_set);
   if (MB_SUCCESS != rval)
@@ -393,19 +398,12 @@ ErrorCode DagMC::build_indices(Range& surfs, Range& vols) {
   // surf/vol offsets are just first handles
   EntityHandle tmp_offset = 0;
 
-  if (surfs.size() != 0 && vols.size() != 0) {
-    setOffset = std::min(*surfs.begin(), *vols.begin());
-    tmp_offset = std::max(surfs.back(), vols.back());
-  } else if (0 == surfs.size()) {
-    setOffset = *surfs.begin();
-    tmp_offset = surfs.back();
-  } else if (0 == vols.size()) {
-    setOffset = *vols.begin();
-    tmp_offset = vols.back();
-  } else {
+  if (surfs.size() == 0 || vols.size() == 0) {
     std::cout << "Volumes or Surfaces founds" << std::endl;
     return  MB_ENTITY_NOT_FOUND;
   }
+  setOffset = std::min(*surfs.begin(), *vols.begin());
+  tmp_offset = std::max(surfs.back(), vols.back());
 
   // set size
   entIndices.resize(tmp_offset - setOffset + 1);
--- a/src/dagmc/DagMC.cpp
+++ b/src/dagmc/DagMC.cpp
@@ -117,6 +117,8 @@ unsigned int DagMC::interface_revision() {
   return result;
 }
 
+/* SECTION I: Geometry Initialization and problem setup */
+
 // the standard DAGMC load file method
 ErrorCode DagMC::load_file(const char* cfile) {
   ErrorCode rval;
@@ -142,7 +144,7 @@ ErrorCode DagMC::load_file(const char* cfile) {
     // Some options were unhandled; this is common for loading h5m files.
     // Print a warning if an option was unhandled for a file that does not end in '.h5m'
     std::string filename(cfile);
-    if (std::string(file_ext) != ".h5m") {
+    if (file_ext != ".h5m") {
       std::cerr << "DagMC warning: unhandled file loading options." << std::endl;
     }
   } else if (MB_SUCCESS != rval) {
@@ -395,15 +397,14 @@ int DagMC::get_entity_id(EntityHandle this_ent) {
 ErrorCode DagMC::build_indices(Range& surfs, Range& vols) {
   ErrorCode rval = MB_SUCCESS;
 
-  // surf/vol offsets are just first handles
-  EntityHandle tmp_offset = 0;
 
   if (surfs.size() == 0 || vols.size() == 0) {
     std::cout << "Volumes or Surfaces not found" << std::endl;
     return  MB_ENTITY_NOT_FOUND;
   }
   setOffset = std::min(*surfs.begin(), *vols.begin());
-  tmp_offset = std::max(surfs.back(), vols.back());
+  // surf/vol offsets are just first handles
+  EntityHandle tmp_offset = std::max(surfs.back(), vols.back());
 
   // set size
   entIndices.resize(tmp_offset - setOffset + 1);
--- a/src/dagmc/DagMC.cpp
+++ b/src/dagmc/DagMC.cpp
@@ -7,6 +7,7 @@
 #include <limits>
 #include <algorithm>
 #include <set>
+#include <climits>
 
 #include <ctype.h>
 #include <string.h>
@@ -20,6 +21,7 @@
 
 #define MB_OBB_TREE_TAG_NAME "OBB_TREE"
 #define FACETING_TOL_TAG_NAME "FACETING_TOL"
+static const int null_delimiter_length = 1;
 
 namespace moab {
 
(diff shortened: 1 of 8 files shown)


bam241:
I am thinking that addressing the linter suggestion in this PR might be a bit much...
Let me know I can simplify this if necessary....

src/dagmc/DagMC.cpp
```diff
@@ -92,16 +94,22 @@ float DagMC::version(std::string* version_string) {
 
 unsigned int DagMC::interface_revision() {
   unsigned int result = 0;
-  const char* interface_string = DAGMC_INTERFACE_REVISION;
-  if (strlen(interface_string) >= 5) {
+  std::string interface_string = DAGMC_INTERFACE_REVISION;
+  if (interface_string.rfind("$Rev:") != std::string::npos) {
```
gonuke: Does anyone know if this actually works!?!?  This looks like a tag expansion that was part of CVS/SVN, but I don't think we ever get "$Rev:" in the file anymore.  Perhaps we need a new solution for this entirely?

gonuke: If it is working, maybe we should define `$Rev:` as a string variable, and refer to it's length where we otherwise use 5 as a magic number below.

I'm not sure why 10 as a magic number below?

src/dagmc/DagMC.cpp
```diff
@@ -92,16 +94,22 @@ float DagMC::version(std::string* version_string) {
...
+    result = strtol(interface_string.c_str() + 5, &endptr, 10);
+    if (endptr == interface_string.c_str() + 5 ||
+        ((result == LONG_MAX || result == LONG_MIN) && errno == ERANGE)) {
+      std::cerr << "Not able to parse revision number" << std::endl;
+      exit(1);
+    }
+    return result;
   }
-  return result;
+  /* SECTION I: Geometry Initialization and problem setup */
```
gonuke: Why did this comment get move into this function?

src/dagmc/DagMC.cpp
```diff
@@ -111,7 +119,12 @@ ErrorCode DagMC::load_file(const char* cfile) {
   char file_ext[4] = "" ; // file extension
 
   // get the last 4 chars of file .i.e .h5m .sat etc
-  memcpy(file_ext, &cfile[strlen(cfile) - 4], 4);
+  if (sizeof(cfile) / sizeof(char) > 4) {
```
gonuke: add variable `int file_extension_length = 4` instead of magic number

src/dagmc/DagMC.cpp
```diff
@@ -111,7 +119,12 @@ ErrorCode DagMC::load_file(const char* cfile) {
...
+    memcpy(file_ext, &cfile[strlen(cfile) - 4], 4);
+  } else {
+    std::cerr << "DagMC warning: unhandled file "
```
gonuke: It might still be better for this error to be generated later, after any error messages that might get generated by `load_file`.  It also looks like they create a `std::string` version of `cfile` - maybe we should do that earlier and take advantage of `std::string` functions throughout.

src/dagmc/DagMC.cpp
```diff
@@ -379,10 +391,21 @@ ErrorCode DagMC::build_indices(Range& surfs, Range& vols) {
   ErrorCode rval = MB_SUCCESS;
 
   // surf/vol offsets are just first handles
-  setOffset = std::min(*surfs.begin(), *vols.begin());
-
-  // max
-  EntityHandle tmp_offset = std::max(surfs.back(), vols.back());
+  EntityHandle tmp_offset = 0;
+
+  if (surfs.size() != 0 && vols.size() != 0) {
```
gonuke: Looking back at the Pull-Request comment....  I am not sure if these checks are necessary.  All valid DagMC geometries must have at least one surface and at least one volume.  I don't know if there is a prior check to make sure this is true.  If there is, then no check is needed here.  If there is not, then perhaps we should check that both are non-zero length and give a failing error if either one is, and then use the (nearly) original code....

src/dagmc/DagMC.hpp
```diff
@@ -359,9 +364,12 @@ class DagMC {
   ErrorCode unpack_packed_string(Tag tag, EntityHandle eh,
                                  std::vector< std::string >& values);
 
-  std::vector<EntityHandle>& surf_handles() {return entHandles[2];}
-  std::vector<EntityHandle>& vol_handles() {return entHandles[3];}
-  std::vector<EntityHandle>& group_handles() {return entHandles[4];}
+  std::vector<EntityHandle>& surf_handles()
+  {return entHandles[surfs_handle_idx];}
```
gonuke: Not sure I like this style...  is this what Astyle requires here? I'd prefer some kind of indent unless our style guide disallows it???

src/dagmc/DagMC.hpp
```diff
@@ -446,7 +454,7 @@ inline int DagMC::index_by_handle(EntityHandle handle) {
 }
 
 inline unsigned int DagMC::num_entities(int dimension) {
-  assert(0 <= dimension && 3 >= dimension);
+  assert(0 <= dimension && groups_handle_idx >= dimension);
```
gonuke: Since you've setup these variables, you might want to add:
`vertex_handle_idx = 0` (to use here) and
`curve_handle_idx = 1`  (for completeness)

src/dagmc/dagmcmetadata.cpp
```diff
@@ -135,81 +138,82 @@ void dagmcMetaData::parse_material_data() {
...
     std::string grp_name = "";
 
     // determine if we have a density property
-    if (!density_props[0].empty()) {
-      grp_name = "mat:" + material_props[0] + "/rho:" + density_props[0];
-      volume_density_data_eh[eh] = density_props[0];
-    } else {
-      grp_name = "mat:" + material_props[0];
-      volume_density_data_eh[eh] = "";
+    if (density_props.size() == 1) {
```
gonuke, line 175: I wonder if it's possible for this to be valid if >1???

src/dagmc/dagmcmetadata.cpp
```diff
@@ -250,7 +254,15 @@ void dagmcMetaData::parse_importance_data() {
       imp_particles.insert(pair.first);
       // insert into map too
       if (importance_map[eh].count(pair.first) == 0) {
-        importance_map[eh][pair.first] = atof(pair.second.c_str());
+        char* e;
+        errno = 0;
+        double imp = std::strtod(pair.second.c_str(), &e);
+        if (*e != '\0' || errno != 0) {
+          std::cerr << "Can't parse " << pair.second << " as a float."
```
gonuke: Provide the user a little more insight about where this came up.
```suggestion
          std::cerr << "Can't parse importance " << pair.second << " as a float."
```

src/dagmc/dagmcmetadata.cpp
```diff
@@ -250,7 +254,15 @@ void dagmcMetaData::parse_importance_data() {
       imp_particles.insert(pair.first);
       // insert into map too
       if (importance_map[eh].count(pair.first) == 0) {
-        importance_map[eh][pair.first] = atof(pair.second.c_str());
+        char* e;
```
gonuke: Wow - this is verbose.  Why can't C++ make this easier???  Does `std::stod` do any better?

src/dagmc/tests/dagmc_pointinvol_test.cpp
```diff
@@ -57,7 +57,7 @@ int dagmc_point_in_vol_dir(double origin[3], double dir[3], int vol_idx) {
   double next_surf_dist;
   EntityHandle next_surf;
 
-  // normalise the vector
+  // normalise the direction vector
```
gonuke:
```suggestion
  // direction vectors are always interpreted as unit vectors - make sure this one is normalized
```

gonuke (changes requested):
Some thoughts based on context and background.

gonuke:
Not sure why this is failing??  Might need a little more verbosity to figure it out.

src/dagmc/DagMC.cpp
bam241: just wondering if only one of them is non zero would the `min()`/`max()` still work ?

src/dagmc/DagMC.cpp
bam241: sounds about right for `$rev`

I don't know what this 10 is for....

bam241: this comes from `src/dagmc/DagMCVersion.hpp.in`

bam241: and did not get populate in my `DagMCVersion.hpp`:
`#define DAGMC_INTERFACE_REVISION "$Rev$"` <- I still have this like in `bld/src/dagmc/DagMCVersion.hpp`

src/dagmc/dagmcmetadata.cpp
bam241: I do not know.

Would anyone would have with some more experience about this ?

makeclean: > 1 is caught earlier on, line 102

src/dagmc/dagmcmetadata.cpp
makeclean: It probably can, IIRC we hadnt shifted off the C++03 standard for MOAB compatability, so going back to C99 was the only option

src/dagmc/dagmcmetadata.cpp
bam241: ok I remember now,
162 ensure that `density_props .size() <= 1`
this ensure we have at lease 1 element before calling `density_props[0]`

src/dagmc/DagMC.hpp
bam241: `astyle` wouldn't let me indent it.
So I opened the brackets.

ljacobson64: I just ran astyle on

```
std::vector<EntityHandle>& surf_handles() {return entHandles[2];}
std::vector<EntityHandle>& vol_handles() {return entHandles[3];}
std::vector<EntityHandle>& group_handles() {return entHandles[4];}
```

 and it gave me

```
std::vector<EntityHandle>& surf_handles() {
    return entHandles[2];
}
std::vector<EntityHandle>& vol_handles() {
    return entHandles[3];
}
std::vector<EntityHandle>& group_handles() {
    return entHandles[4];
}
```

ljacobson64: Not to turn this into a discussion about changing our preferred code formatter, but `clang-format` gives

```
std::vector<EntityHandle> &surf_handles() { return entHandles[2]; }
std::vector<EntityHandle> &vol_handles() { return entHandles[3]; }
std::vector<EntityHandle> &group_handles() { return entHandles[4]; }
```

which looks the best in my opinion

bam241: Yep this is what I ended up implementing :)

as it is in the header file, I was trying to keep it as short as possible.... (this is why I tried the open/close brackets in the same line.

bam241: As I replaced the magic number (`2`/`3`/`4`) by variable, lines are now `> 80` chars with the inline formatting

src/dagmc/DagMC.cpp
```diff
@@ -92,27 +94,43 @@ float DagMC::version(std::string* version_string) {
...
+            && errno == ERANGE)) {  // checking range
+      std::cerr << "Not able to parse revision number" << std::endl;
+      exit(1);
+    }
+    return result;
   }
   return result;
 }
 
-/* SECTION I: Geometry Initialization and problem setup */
```
gonuke: Did we take out the other comments that for these `Section` headers?

src/dagmc/DagMC.cpp
```diff
@@ -379,10 +396,14 @@ ErrorCode DagMC::build_indices(Range& surfs, Range& vols) {
   ErrorCode rval = MB_SUCCESS;
 
   // surf/vol offsets are just first handles
-  setOffset = std::min(*surfs.begin(), *vols.begin());
+  EntityHandle tmp_offset = 0;
 
-  // max
-  EntityHandle tmp_offset = std::max(surfs.back(), vols.back());
+  if (surfs.size() == 0 || vols.size() == 0) {
+    std::cout << "Volumes or Surfaces founds" << std::endl;
```
``` suggestion
    std::cout << "Volumes or Surfaces not found" << std::endl;
```

gonuke (changes requested):
Maybe a couple of things lost in the revisions...

gonuke (changes requested):
Maybe a couple of things lost in the revisions...

src/dagmc/DagMC.cpp
bam241: I am not sure what I did with this...
Some kind of a weird combination of typos...

I'll restore it.

src/dagmc/DagMC.cpp
```diff
@@ -124,7 +144,7 @@ ErrorCode DagMC::load_file(const char* cfile) {
     // Some options were unhandled; this is common for loading h5m files.
     // Print a warning if an option was unhandled for a file that does not end in '.h5m'
     std::string filename(cfile);
-    if (filename.length() < 4 || filename.substr(filename.length() - 4) != ".h5m") {
+    if (std::string(file_ext) != ".h5m") {
```
gonuke: `file_ext` is already a `std::string` - no need to caste it here

src/dagmc/DagMC.cpp
```diff
@@ -379,10 +398,14 @@ ErrorCode DagMC::build_indices(Range& surfs, Range& vols) {
   ErrorCode rval = MB_SUCCESS;
 
   // surf/vol offsets are just first handles
-  setOffset = std::min(*surfs.begin(), *vols.begin());
+  EntityHandle tmp_offset = 0;
```
gonuke: Combine this with line 408

gonuke (changes requested):
Two small requests & rebase for merge today

bam241:
@gonuke I think this is ready.
I rebased, addressed your 2 remaining comments and cleaned the commit history (squash).

gonuke (approved):

gonuke:
Thanks @bam241 - finally merging!!!

PR #679, opened 2020-05-28
clang-format example using Google style

bam241:
just a test on 2 files to see what would happened


src/dagmc/DagMC.hpp
```diff
@@ -86,17 +89,18 @@ class DagMC {
...
-   * things happen that one should be aware of.
+   * Note: When loading a prexisting file with an OBB_TREE tag, a number of
+   *unspoken things happen that one should be aware of.
    *
-   * 1) The file is loaded and when we query the meshset, we find entities with the OBB_TREE tag
-   * 2) The OBBTreeTool assumes that any children of the entity being queried in a ray intersect sets
-   *     operation are fair game, the surface meshsets have triangles as members, but OBBs as children
-   *     but no querying is done, just assumptions that the tags exist.
+   * 1) The file is loaded and when we query the meshset, we find entities with
+   *the OBB_TREE tag 2) The OBBTreeTool assumes that any children of the entity
```
bam241, line 106: there is an issue here that would need to be fixed manually

ljacobson64:
I personally really like `clang-format`

bam241:
there it is.... this is a lot....

I used default Google style in clang-format which only sort #include by alphabetical order, does't do block.... we can configure include blocks if we'd like.
I also updated the housekeeping script to use clang-format instead of `astyle`....

bam241:
@gonuke @pshriwise @makeclean @ljacobson64 this is rdy, maybe not to be merged, but at least to discuss it :)

gonuke:
I'm willing to make this change, but I'm concerned about the need for a manual fix.  That's not practical given the way we use it.

bam241:
I am hoping that the manual fixes are just for the first time.

Those files that I fixed, are now passing `clang-format` without change, just clang-format seems to able to reformat them properly the first time.

ljacobson64:
I have also noticed that in very rare situations, you have to run clang-format twice in order to get the code to a state where it won't change again upon a further run of clang-format. This is not ideal, but again, very rare: out of ~89000 lines of C++ code in DAGMC, this only occurs in 3 lines. I don't think this is a big enough problem to worry about.

gonuke:
It's only a problem if it happens every time that CI is launched.  That is, even if its only a couple of lines, if it happens every time it will be a pain.

ljacobson64:
It should never be a problem in CI as long we run it twice the first time we run clang-format. From then on, it should only ever need to be run once.

bam241:
@gonuke, @ljacobson64 is correct,

in few cases clang-format is not good enough to properly fix the formatting, but one can fix it manually in a clang-format compliant way.

this is what I did here, we are passing the new `clang-format` test with some file I had to format manually.

if manual formatting does not work it is possible add flag to tell clang-format to ignore this part of the file. see src/dagmc/DagMCVersion.hpp.in

ljacobson64:
@bam241 I'm confused about what you mean by needing to format files manually

bam241:
so sometimes basically with running clang-format the end results is messy

see the example I mentioned on `dagmc.hpp` up here. I had to manually correct the formatting to get it a nice and a clang-format compliant form.

bam241:
To many commits have past, I'll redo this based on the actual develop state.

But we have to move quickly, as a lot is going to be changing a little....

src/overlap_check/test/overlap_check_test.hpp
```diff
@@ -1,16 +1,19 @@
 
+#include <memory>
+
 #include "gtest/gtest.h"
+#include "moab/Core.hpp"
```
bam241, line 6: Reordering includes in the cpp files forced me to move some of the includes into the header file.

I think in any case it was bad practice to have some of the required header includes in the cpp file (and as they got loaded before the header, it was ok...)

src/overlap_check/test/overlap_check_test.cpp
```diff
@@ -1,17 +1,15 @@
...
-#include <algorithm>
-#include <memory>
-#include "moab/Core.hpp"
+#include <vector>
 
 #include "overlap.hpp"
 
-#include "overlap_check_test.hpp"
-
 void OverlapTest::SetUp() {
```
bam241, line 13: reordering includes broke compilation...

`memory` and `moab/Core` were included before `overlap_check_test.hpp` so compilation was ok, events they were needed by `overlap_check_test.hpp` and not included there...
I moved them in `overlap_check_test.hpp`....

bam241 (commented):
The only manual change I had to make was in
`src/overlap_check/test/overlap_check_test.cpp `
`src/overlap_check/test/overlap_check_test.hpp `

bam241:
As so many file have to change, any PR merged before this, would force me to update this. So please review promptly :)

gonuke (approved):
My review is only based on the files that implement the change and not the style changes themselves.  I assume we are comfortable with those style changes, and feel we can rely on future PRs to change/update individual aspects of style if we deem it necessary.

bam241:
There is few files that might need to be tweaked within the style

This assumes we are ok using `google` style from clang-format without any fancy option.

I don't necessary want to use `google` style, but I strongly believe we should pick a standard style and use it as is... It would make developers life so much easier...

pshriwise:
I've become a fan of the [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines). They're written by C++ language developers who have a deep understanding of the language and its roadmap. We don't currently enforce this as part of our PRs (yet), but we do have a [`.clang-format` file](https://github.com/openmc-dev/openmc/blob/develop/.clang-format) in the repo which applies these guidelines.

helen-brooks:
No problems from my end.

bam241:
> I've become a fan of the [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines). They're written by C++ language developers who have a deep understanding of the language and its roadmap. We don't currently enforce this as part of our PRs (yet), but we do have a [`.clang-format` file](https://github.com/openmc-dev/openmc/blob/develop/.clang-format) in the repo which applies these guidelines.

what I am fearing with a so complicated and customized set of .clang-format file (from openMC) is that we make it more complicated for dev to use.

A simple .clang-format following a default style would be more practical and simpler to enforce...

gonuke:
We should probably announce this changes since it may be a big task for anyone rebasing this into their branch

PR #686, opened 2020-06-30
Replace LAPACK with Eigen3 for KDE tallies

gonuke:
## Description
This change removes the LAPACK dependency that was embedded in KDE Tallies.  This dependency was only used for solving Ax=b for 3x3 and 4x4 systems.  It has been replaced by Eigen3.

## Motivation and Context
Because LAPACK is a compiled library dependency, there may be conflicts between the compiler used to build LAPACK and the compiler being used to build DAGMC.  This is especially true on shared HPC systems where only certain combinations are available.  Since Eigen3 is a C++ header only library, this potential for conflict is eliminated.

Also, since LAPACK requires FORTRAN, building against LAPACK may impose additional requirements on the environment.

## Changes
This change is an infrastructure change - no bug is fixed and no feature is added.  A small change was made to documentation to mention Eigen3 as a dependency. (Note: Lapack was never explicitly listed as a dependency in the past.)

## Behavior
The behavior has not changed.


gonuke:
Note: There is not a lot of institutional knowledge about this part of the code base.  Unit tests played an important role since they initially failed, but now pass.

ljacobson64:
LGTM. Thanks @gonuke!

pshriwise (approved):
Thanks @gonuke! Nice to see this transition happening. We've been discussing it for some time.

Looks like docs will need to be rebuilt after those goes in.

src/tally/KDEKernel.cpp
```diff
@@ -49,9 +51,11 @@ double KDEKernel::boundary_correction(const double* u,
 
   // solve for the boundary correction factor
   if (num_corrections == 1) {
+
```
bam241: are those blank lines required by style ?

src/tally/KDEKernel.cpp
```diff
@@ -49,9 +51,11 @@ double KDEKernel::boundary_correction(const double* u,
...
     double correction_factor = (ai_u[2] - ai_u[1] * u[0]);
     correction_factor /= ai_u[0] * ai_u[2] - ai_u[1] * ai_u[1];
     return correction_factor;
+
```
bam241: same here.

bam241 (approved):
LGTM!
(1 minor style comment)

thanks @gonuke

src/tally/KDEKernel.cpp
gonuke: Not sure, but `astyle` doesn’t complain, so I hold my breath and cross my fingers....

bam241:
```suggestion
```

src/tally/KDEKernel.cpp
bam241:
```suggestion
```

src/tally/KDEKernel.cpp
bam241: we really need to move away from `astyle` :)

gonuke:
seems like we lost CI again, but it passed here: https://travis-ci.org/github/svalinn/DAGMC/builds/705468798

bam241:
LGTM!
Merging.
thx @gonuke !

PR #684, opened 2020-06-14
Fix eigen3 bld

gonuke:
Fix the MOAB build in the docker image to use Eigen3 and no FORTRAN support

Issue #606, opened 2019-02-20
Cannot build static executables with system HDF5


ljacobson64:
It is not possible to build static executables using MOAB built with system HDF5. The error messages that occur look like this:

```
/usr/lib/x86_64-linux-gnu/libsz.a(sz_compat.o): In function `SZ_BufftoBuffCompress':
(.text+0x149): undefined reference to `aec_buffer_encode'
(.text+0x339): undefined reference to `aec_buffer_encode'
```

The issue is that system HDF5 links to another library called `sz`, while custom HDF5 builds do not (at least by default). CMake's `FindHDF5` macro correctly identifies whether `sz` needs to be added to the `HDF5_LIBRARIES` variable, but it does not pick up on the fact that `sz` itself is linked to another library: `aec`. This is the problem.

The reason CI doesn't fail because of this is that we don't use system HDF5 on CI. We build our own HDF5.

I tested what happens if you build your own HDF5 but enable `sz`, and then build MOAB on that HDF5, and then build DAGMC on that MOAB. The same issue appears: it doesn't know that it needs to look for `aec`.

I don't know the best way to fix this. We could manually add `aec` to our `HDF5_LIBRARIES` variable, but this doesn't seem like a good idea though since not all systems will need it. Perhaps we could check what's in `HDF5_LIBRARIES`, and if it finds `sz`, manually add `aec` too?

PR #687, opened 2020-06-30
Moab cmake

gonuke:
This changes the build process for MOAB in the docker image to use CMake instead of Autotools.

This build system is more robust across different platforms and will be where ongoing improvements are focused.

depends on PR #684


pshriwise:
In the past DAGMC has required that both static and shared libraries are included in the MOAB install but to my knowledge CMake can only build and install one set of those at a time. From these changes it looks like we'll no longer have a set of static libraries in the Travis docker image. (It seems the CMake default for MOAB is to build shared object libraries.)

gonuke:
Good point.  I've been advocating a move to shared libraries only but I can't recall if there is a real reason to keep static libraries?  @ljacobson64 @makeclean ?

gonuke:
This might break CI since it tries to build both static and shared versions...???

pshriwise:
Yeah I'd guess that it would. An additional build and install with `BUILD_SHARED_LIBS=OFF` has done the trick for me in the past.

gonuke:
I added a static lib build for now - we can push off the debate for static + shared to another time....

bam241:
LGTM! 
Did someone test ro rebuild the docker image using this ?

gonuke:
I haven’t tested a new image... I suppose the best plan is to test the CI building & test process on a new image built with this

bam241:
I'll do that later today

bam241:
ubuntu 16.04 seems to be working.
testing 18.04 right now.

bam241:
having issue with 18.04, all the test failed with:
```
est project /root/build/gcc/DAGMC-moab-5.1.0-hdf5-1.10.4-shared/bld
      Start  1: dagmc_unit_tests
 1/24 Test  #1: dagmc_unit_tests ........................***Failed    0.00 sec
/root/build/gcc/DAGMC-moab-5.1.0-hdf5-1.10.4-shared/bld/src/dagmc/tests/dagmc_unit_tests: error while loading shared libraries: libhdf5.so.103: cannot open shared object file: No such file or directory
```

pshriwise:
If it isn't too much trouble, could we add 19.10 into the mix as well? I vaguely recall having some problems that were unique to building MOAB with CMake in that version...

gonuke:
Shouldn't we leap to 20.04 then?

ljacobson64:
I think we should only have LTS releases (16.04, 18.04, 20.04) in CI

pshriwise:
Mmm yeah that would make more sense.

gonuke:
So back to the original error....  

It seems that the custom HDF5 shared lib is not being found, so it's not in the library path or rpath is not right.... 

Also, do we only test in CI against the custom HDF5 and never against the system HDF5?  Or am I reading something wrong...

ljacobson64:
We only test with the custom HDF5.

https://github.com/svalinn/DAGMC/issues/606

bam241:
@gonuke I still have the issue, but not on build, on test only.

I am wondering is this could be related to the issue Helen mention this morning ?!

pshriwise:
@bam241 did you try running the installed test executeables? Same issue?

bam241:
@pshriwise I did not try, I use a script that mimics what CI is supposed to do.
(I did not investigate this to much either :))

gonuke:
I suspect a combination of things here, but don't have time to test:

- autotools is setting the rpath when it builds MOAB so that it knows where to find HDF5 and CMake is not
- the LD_LIBRARY_PATH is being cleared by ctest (per Helen's discovery in the CMakeLists.txt file)

This would only fail when running ctest after building MOAB with CMake.

ljacobson64:
If the problem is that the CMake build isn't adding the rpath, this can be fixed without needing to change MOAB at all by adding `-DCMAKE_INSTALL_RPATH=${hdf5_install_dir}/lib:${moab_install_dir}/lib` to the MOAB cmake command

gonuke:
I added the RPATH line that @ljacobson64 suggested - maybe that will help

ljacobson64:
Looks like that fix worked -- the tests are passing on Travis

gonuke:
CI is not testing this since this is just a change in the Dockerfiles that build the images

ljacobson64:
Right, never mind

shimwell:
Thanks for making this, I think it will be most useful for an open OpenMC issue https://github.com/openmc-dev/openmc/issues/1709 where we are trying to make the automated OpenMC docker image optionally include DAGMC

bam241:
what is the status there ?

gonuke:
This has been rebased against changes proposed in #735

ljacobson64 (approved):
This LGTM, assuming the tests pass after the docker image is rebuilt

gonuke:
I'm rebuilding docker images now (1/2 done) and will test later and update

gonuke:
All tests passed on new docker images.  This CI should be operating on those same docker images now.

gonuke:
I guess they didn't pass :(

CI/docker/build_moab.sh
```diff
@@ -14,19 +14,28 @@ rm -rf ${moab_build_dir}/bld ${moab_install_dir}
...
-make -j${jobs}
+cd bld
+cmake ../moab -DBUILD_SHARED_LIBS=OFF \
+              -DENABLE_HDF5=ON -DHDF5_ROOT=${hdf5_install_dir} \
+              -DENABLE_BLASLAPACK=OFF \
+              -DENABLE_FORTRAN=OFF \
+              -DCMAKE_INSTALL_PREFIX=${moab_install_dir} \
+              -DCMAKE_C_COMPILER=${CC} \
+              -DCMAKE_CXX_COMPILER=${CXX} \
+              -DCMAKE_INSTALL_RPATH=${hdf5_install_dir}:${moab_install_dir}
```
ljacobson64: I think that replacing
```
-DCMAKE_INSTALL_RPATH=${hdf5_install_dir}:${moab_install_dir}
```
with
```
-DCMAKE_INSTALL_RPATH=${hdf5_install_dir}/lib:${moab_install_dir}/lib
```
here will fix the current CI issue.

CI/docker/build_moab.sh
```diff
@@ -14,19 +14,28 @@ rm -rf ${moab_build_dir}/bld ${moab_install_dir}
...
+rm -rf *
+cmake ../moab -DBUILD_SHARED_LIBS=ON \
+              -DENABLE_HDF5=ON -DHDF5_ROOT=${hdf5_install_dir} \
+              -DENABLE_PYMOAB=ON \
+              -DENABLE_BLASLAPACK=OFF \
+              -DENABLE_FORTRAN=OFF \
+              -DCMAKE_INSTALL_PREFIX=${moab_install_dir} \
+              -DCMAKE_C_COMPILER=${CC} \
+              -DCMAKE_CXX_COMPILER=${CXX} \
+              -DCMAKE_INSTALL_RPATH=${hdf5_install_dir}:${moab_install_dir}
```
ljacobson64: Same comment here.

gonuke:
I'll try that a little later today

CI/docker/build_moab.sh
ljacobson64: https://github.com/svalinn/DAGMC/pull/687#issuecomment-656489283

gonuke: It looks like I might have dropped the `/lib` when adapting this in  the last update

gonuke:
Rebuilding docker images this morning (hopefully) to push this afternoon for final testing :crossed_fingers:

gonuke:
Relaunched with new docker images.  Everything passed except the elusive overlap_check tests that appear to be a MOAB problem?

ljacobson64:
Weird... and the tests pass with clang but not with gcc

gonuke:
This problem has been popping up more & more often... ☹️

bam241:
is that not a problem related to MOAB ?! (I vaguely remember @pshriwise tell me this....)

gonuke:
Yes - I think he's working on it in MOAB

ljacobson64:
Seems odd that switching moab's build system from autotools to cmake would cause this issue to pop up

bam241:
I believe this problem predate the proposed change to switch to MOAB build to CMake.

bam241:
the root of the error might be explained there: https://stackoverflow.com/questions/30887143/make-j-8-g-internal-compiler-error-killed-program-cc1plus

gonuke:
Should we combine this with #726 while we're waiting?

pshriwise:
It looks like this now needs to be rebased to ensure we're using the updated docker images with a recent version of MOAB. Is that right?

pshriwise:
@ljacobson64 I know it's been a while, but is this looking good to you still?

gonuke:
I suppose someone should rebuild the CI images and confirm that they work

gonuke:
Or we can wait for the MOAB release for this one....

gonuke:
This was replaced by #771 in our new build infrastructure.

PR #688, opened 2020-06-30
Making density specification optional in the metadata manager.

pshriwise:
## Description

This provides an option to allow the `dagmcMetaData` class to read material assignments without requiring that a density specification is present with the default being the original behavior of the class.

## Motivation and Context

My take on this is that this requirement was originally adopted for DAG-MCNP, where cell material assignments include specification of the density in that region. In other codes the density is supplied as part of the material definition (e.g. OpenMC) so having a density present in the DAGMC assignments creates something of a contradiction between the specified DAGMC density and the density specified as part of the native material definitions.

This change provides the option to omit or ignore the density specification on the DAGMC model material assignments, but still use the `dagmcMetaData` class to parsing of the model metadata rather than rely on other methods (like using the somewhat outdated metadata options that are part of the `DagMC`  class itself).

### DIFF
--- a/src/dagmc/dagmcmetadata.cpp
+++ b/src/dagmc/dagmcmetadata.cpp
@@ -7,9 +7,12 @@ const char graveyard_str[] = "Graveyard";
 const char vacuum_str[] = "Vacuum";
 
 // constructor for metadata class
-dagmcMetaData::dagmcMetaData(moab::DagMC* dag_ptr, bool verbosity) {
+dagmcMetaData::dagmcMetaData(moab::DagMC* dag_ptr,
+                             bool verbosity,
+                             bool ensure_density_present) {
   DAG = dag_ptr; // dagmc pointer
   verbose = verbosity;
+  ensure_density = ensure_density_present;
   // these are the keywords that dagmc will understand
   // from groups if you need to process more
   // they should be added here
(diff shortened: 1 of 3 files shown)


gonuke:
Thanks @pshriwise - seems like there is virtually no cost to including this.

I might prefer `require_density` vs `ensure_density` but that's a quibble....

pshriwise:
Happy to make that change, @gonuke. It'll be in shortly.

gonuke (approved):
Thanks @pshriwise - LGTM

src/dagmc/dagmcmetadata.cpp
```diff
@@ -7,9 +7,12 @@ const char graveyard_str[] = "Graveyard";
 const char vacuum_str[] = "Vacuum";
 
 // constructor for metadata class
-dagmcMetaData::dagmcMetaData(moab::DagMC* dag_ptr, bool verbosity) {
+dagmcMetaData::dagmcMetaData(moab::DagMC* dag_ptr,
+                             bool verbosity,
+                             bool require_density_present) {
```
bam241, line 12: I think modern C++ advice to use member initialization rather than assignment in constructors.

as you add a new member maybe it is a good place to start ?

bam241 (approved):
This looks good to me.

Minor suggestion in order to move toward [modern C++ practices](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-initialize).... 
(I don't think we should hold the PR for this tho...)

src/dagmc/dagmcmetadata.cpp
pshriwise: I had the same thought and opted to go for the minimal set of changes for now. There are a few other things I'd like to update as well. If it's alright I'll plan on submitting those in a separate PR as you suggested.

bam241: sounds good for me
Again this is definitely **NOT** a blocking point.

I just saw it and it popped in my mind that we should start eventually to do that....

gonuke:
Thanks @pshriwise

PR #689, opened 2020-07-02
Metadata class updates.

pshriwise:
## Description
Updates to the `dagmcMetaData` class. 

This addresses a comment from @bam241 in #688 about some potential improvements to style in the `dagmcMetaData` class.

It addresses an output type in the UWUW class as well.

### DIFF
--- a/src/dagmc/dagmcmetadata.hpp
+++ b/src/dagmc/dagmcmetadata.hpp
@@ -6,27 +6,38 @@
 
 class dagmcMetaData {
  public:
+  // Constructor
   dagmcMetaData(moab::DagMC* DAGptr,
                 bool verbosity = false,
                 bool require_density_present = true);
-  ~dagmcMetaData();
+
+  // Destructor
+  ~dagmcMetaData() = default;
 
   // load the dagmc properties into maps
   void load_property_data();
 
   // get a given property on a volume
   std::string get_volume_property(std::string property, moab::EntityHandle vol);
+  // get a property for the specified volume, treats the vol parameter as
+  // an index by default and as an ID if idx is false.
   std::string get_volume_property(std::string property, int vol, bool idx = true);
 
   // get a given property on a surface
   std::string get_surface_property(std::string property, moab::EntityHandle surface);
+  // get a property for the specified surface, treats the surface parameter as
+  // an index by default and as an ID if idx is false.
   std::string get_surface_property(std::string property, int surface, bool idx = true);
 
+  // unpack the packed string of the form delimeter<data>delimiter<data>delimiter into
+  // a vector of the form data[0],data[1] etc
   std::vector<std::string> unpack_string(std::string to_unpack, std::string delimiters = "|");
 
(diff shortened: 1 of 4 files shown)


gonuke:
Thanks @pshriwise.  

I seem to remember a discussion about the pros/cons of using `auto`, but can't remember the outcome.  Have we already committed to that path?

pshriwise:
My opinion is that it's handy to use in the right situations, but should be used cautiously. In certain situations it makes the code less verbose and more readable. Here I felt using it to replace the very long type definitions was worth it. I do try to avoid using it in places where the compiler will infer a type as that can lead to unexpected problems however. Using it as the return type in function signatures is one such example.

IIRC @makeclean may have strong opinions about `auto`. I'm happy to defer to any strong objections about its use.

makeclean:
My only objection is the obfuscation it can cause with multi-header types, but that lack of clarity is fixed by an ide with type discovery. So I don't know that I'm opposed totally, it's very handy but if youre not using an editor with type discovery then complex types will be painful. Also happy to defer

gonuke:
Apparently this failed the `astyle` check, although not reported here.

pshriwise:
I applied the style guide command, but it made some changes there weren't necessarily.... stylish. I'm curious to see what CI does with it.

gonuke:
This one passed CI (but I'm still not sure why it's not showing the results of tests here...) @bam241 ?

src/dagmc/dagmcmetadata.hpp
```diff
@@ -47,13 +58,20 @@ class dagmcMetaData {
...
   void finalise_counters();
 
-  std::map<moab::EntityHandle, std::vector<std::string> > get_property_assignments(std::string property,
-      int dimension,
-      std::string delimiters,
-      bool remove_duplicates = true);
+  // Parse property for entities with the specified dimension and delimiters.
+  // Optionally remove duplicate property values if necessary.
+  std::map<moab::EntityHandle, std::vector<std::string>>
+                                                      get_property_assignments(std::string property,
```
pshriwise, line 64: Not sure why the style guide considers this to be better...

gonuke: agreed!  @bam241 has suggested that we switch to clang-format.... maybe soon...

bam241:
> This one passed CI (but I'm still not sure why it's not showing the results of tests here...) @bam241 ?

there is a failure somewhere in the link between Travis-CI and GitHub, restarting the tests corresponding the the PR on CI ends usually does the trick :)

pshriwise:
@makeclean agreed. We probably don't want to go crazy with it. Using it for native types like `double`, `int`, and even `std::vector` probably isn't worth it. It can really help with these nested template types though I think. If you're ok with using it sparingly I'll leave these changes as they are. 

Seems as though CI has reconnected here too, so that's good.

gonuke:
Thanks @pshriwise - I'll merge this and we'll officially have `auto` in core DAGMC code....

PR #690, opened 2020-07-07
Surf props

pshriwise:
## Description
This PR includes additional options for surface boundary conditions to the `dagmcMetaData` class. It also ignores capitalization when comparing strings for boundary condition values as well as the graveyard and vacuum identifiers.

A test case for the lowercase values as well as a new boundary condition has been added in the `TestBoundaryAssigns` test.

## Motivation and Context
This allows for support for other commonly used boundary conditions in Monte Carlo codes.


src/dagmc/dagmcmetadata.cpp
```diff
@@ -337,10 +345,17 @@ void dagmcMetaData::parse_boundary_data() {
     // 2d entities have been tagged with the boundary condition property
     // ie. both surfaces and its members triangles,
 
-    if (boundary_assignment[0].find("Reflecting") != std::string::npos)
+
+    std::string bc_string = to_lower(boundary_assignment[0]);
+
+    if (bc_string.find("reflecting") != std::string::npos)
```
gonuke: Does it make sense to unmagic these strings like we've done with `graveyard_str` and `vacuum_str`?

gonuke (changes requested):
Thanks @pshriwise - nice additions.

There may be robustness questions with a DAGMC periodic boundary, but look forward to seeing an implementation.

src/dagmc/dagmcmetadata.cpp
pshriwise: Yeah I'd say so. Can do!

gonuke (approved):

gonuke:
LGTM - Thanks @pshriwise

PR #691, opened 2020-07-07
Addressing PullRequest comments related to Geant4 

bam241:
This addresses most of the comment from PullRequest #660.

Most of the comments are related to commented pieces of code.

As most of them are in the `N01Example` I which of those are there as example of what could be used in it.
I tried to sort the dirty ones (old version not fully cleaned up) to the one showing some additional capabilities...

Might be wrong tho, and additional look might be a good idea. (@makeclean ?)


gonuke:
astyle failure :(

bam241:
@gonuke @pshriwise styled. this should be ok when the tests passe

gonuke (approved):
LGTM @bam241

PR #692, opened 2020-07-08
test Circle CI file

bam241:
This is an attempt to run CI using in CircleCI


bam241:
@gonuke @pshriwise @ljacobson64  I think this is ready to be reviewed.

I used a `test_` prefix on the docker image name, so this would need to be changed before we move forward if we decide to...

bam241:
I did there:
- revamp CI to use Circle in place of Travis
- revamp Docker container generation to take advantage of the layering capabilities of the Dockerfile. With this we now have 15 docker containers where we one had 2 before.
- various simplification possible because now each setup has its own docker container.

this should also simplify the addition of new configurations (i.e ubuntu 20.04, other gcc version, other hdf5 version etc...)

CI/env.sh
```diff
@@ -2,47 +2,26 @@
...
 export dagmc_build_dir_shared=${dagmc_build_dir}-shared
 export dagmc_install_dir_shared=${dagmc_install_dir}-shared
 
 export dagmc_build_dir_static=${dagmc_build_dir}-static
 export dagmc_install_dir_static=${dagmc_install_dir}-static
 
 
+# Compiler related variables
+
+export  FC=`which gfortran`
```
gonuke, line 18: Do we actually use this anywhere in CI?

bam241:
I think, some optimisation on the `TRAVIS` variables are possible.

I build a "hack" to populate the same env variable that travis are populating in order to have all the scripts working almost as his. this would probably need to be optimized.... 

but didn't want to spent the time before we actually agree on this.

CI/env.sh
```diff
@@ -2,47 +2,26 @@
 
 export jobs=`grep -c processor /proc/cpuinfo`
```
gonuke, line 3: Is this used?  And why override for GCC below?

gonuke (changes requested):
This looks pretty good to me.  A few questions/comments...

Do you think there would be a way to use Dockerhub to generate all the different variations?

CI/env.sh
bam241: gcc compilation crashes if I used all the processors available....

did not dig much more

bam241:
> Do you think there would be a way to use Dockerhub to generate all the different variations?

Probably yes.

It depends on how much time it needs to compile geant4.

but I think we could trigger (in dockerhub) the first build on this repo, and trigger all the other on parents  container modifications.

CI/env.sh
bam241: I don't know.
I could try to remove it....

bam241: it is used in our various `build_XXXX.sh` scripts

CI/env.sh
bam241: it is used in our various `build_XXXX.sh` scripts

gonuke: Got it.  I guess I was sloppy in looking for where it was used by just searching in the PR on github and didn't think about the other files - oops.

CI/env.sh
gonuke: Perhaps it is also used elsewhere in the scripts - I only searched the PR as shown by GH.  However, I don't think we build any fortran in CI or docker???

bam241: I don't think either.

I just put it there, because it looked better, without thinking more about it...

bam241:
our contribution docs and our build status is based on Travis CI, 

While contribution docs could be updated here build status will probably need to wait for the first PR relying on CI to be merged (this one)

CI/env.sh
bam241: we do install `gfortran` in docker

gonuke: I think the only place we use FORTRAN is for building Fluka and MCNP.  We don't do either of those in CI (right now).  That said, it might be wise to keep it around in case developers want to use our images as part of their own workflow for building/testing.

news/PR-0692.rst
```diff
@@ -0,0 +1,11 @@
+**Added:** None
```
gonuke, line 1: This news file probably needs some contents

script
```diff
@@ -0,0 +1,16 @@
+docker build -t svalinn/test_dagmc-ci-ubuntu-16.04-clang-g4-hdf5_1.10.4-moab_master --build-arg UBUNTU_VERSION=16.04 --build-arg COMPILER=clang --build-arg HDF5=1.10.4 --build-arg MOAB=master -f CI/Dockerfile_3_moab .
```
gonuke: Is this meant to be part of the PR?  Doesn't it duplicate what's in `update_docker.sh`?

gonuke:
I think it's OK to wait on the contributing docs until it's easier to create authentic screen shots with Circle CI in use.  I think it's fine to wait on build status as well.

script
bam241: I don't think that was not supposed to be in there.

gonuke: was your double negative in the above sentence intentional or French?

bam241:
**WARNING**
@gonuke 
I am using a prefix `test_` for all the docker containers using for CircleCI in this PR to avoid collision with the existing containers.
One need to update the name of the containers and fix the circle.yml config file before merging.

Ping me with you feel like this is ready.

script
bam241: Neither I edited my sentence and forgot the second negative...

gonuke (approved):

gonuke:
I think this is ready, as far as I'm concerned.  Maybe
1. let the CI pass with the current setup
2. rename all the files
3. push those changes
4. let CI pass again
5. I'll merge

bam241:
CI [passed](https://app.circleci.com/pipelines/github/svalinn/DAGMC/98/workflows/2b08e8f8-c733-494f-963a-61e867ce3f1d)

I'll start the update of the docker container.

bam241:
@gonuke I think this is good to go.
CI passed on the renamed Circle Container.

gonuke:
Thanks @bam241

PR #693, opened 2020-07-15
Support for ray tracing with Embree via double-down

pshriwise:
## Description
This adds an experimental feature to DAGMC which allows Embree to be used as back-end ray tracer for DAGMC via the new (and also somewhat experimental) [double-down](https://github.com/pshriwise/double-down) project.

I'm creating a draft here for comments on the approach, to check it on CI, and for others to try it out if they are interested.

## Motivation and Context
Provides significant serial and parallel efficiency.

## Changes
Creates a type alias for the `moab::GeomQueryTool` object. The double down project implements a very similar interface.


gonuke:
I think I follow what's going on and the concept seems OK to me.  While it's a little tedious to change it everywhere, I wonder if we want to use a name other than `GQT` so that we don't assume that its a `GeometryQueryTool` instance.

Also not sure why all the new scripts in `CI/docker`?

pshriwise:
Good point, @gonuke. I'm not sure why all those updates are there. I should've checked the diff more carefully I suppose. I'll see what I can do about them. 

I'm happy to update the `GQT` name to something new. Any suggestions? If sticking with our acronym scheme, I was thinking maybe `RT` (`RayTracer`) or `RTI` (`RayTracingInterface`). In a vacuum, I'd probably go with `ray_tracer`.

Update:
Looks like some of the CI script relocations weren't tracked in the rebase I did. I'll take care of those updates manually.

gonuke:
I'm OK with `ray_tracer`.  The current syntax standard is motivated by just how long the class names are, but I think `ray_tracer` is short enough.

gonuke:
Where will `RTI.hpp` and `MOABRay.hpp` live?

pshriwise:
Those are both part of double-down and will be provided as part of the `dd` target when finding the double down installation in CMake.

pshriwise:
I'm going to upgrade this to a real PR. It's passing checks now and it's had a once-over from @gonuke (thanks!)

CI/travis/build_double_down.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
gonuke, line 1: Should we put this into some/all of our docker images?  I don't know how long it takes to build this, but it would be great it we didn't add to the build time for DAGMC for every PR.

gonuke (commented):
It all looks good to me - just one strategic question about CI

CI/travis/build_double_down.sh
pshriwise: Good point! I'll get that added to the image.

gonuke: You may want to confer with @bam241 about how best to add it to the matrix within Travis or wait until he gets CircleCI going?

pshriwise: Indeed. @bam241 what do you think?

bam241: Either way works for me.

it will probably be easier for you to wait for the CircleCI to be merge.
(It is almost ready. I think if it got some more reviews it could be merged today)

pshriwise:
Ok so after some discussion it looks like this PR will now depend on #692 where we can incorporate the Embree and double-down build in our CI images.

gonuke:
Yup - and #692 is merged now.  Let me know if you need any help weaving the double down into the new CI/docker infrastructure

bam241:
@pshriwise let me know if you need a hand with the new CI/docker infrastructure :)

pshriwise:
Ok so it looks like there are a few failing builds on the `develop` branch in CI. In Travis, these builds were allowed failures. I did a little searching but didn't find anything about allowed failures for jobs in Circle. Is that possible @bam241?

I ask because any builds with Double Down should probably be allowed failures as well.

pshriwise:
Ran a build in CI with the double down turned on for commit [8051e9b](https://app.circleci.com/pipelines/github/pshriwise/DAGMC/20/workflows/9db663a3-74c7-495d-b634-cf411a2731c8), but I've moved it into the merge workflow now as it's an experimental build. I'm assuming we don't want it to hold up future PR's if it fails.

pshriwise:
One more note, I added a flag to do a dry run of all the docker builds without pushing them. I found this useful in making sure I wasn't partially replacing the CI images.

CI/Dockerfile_4_embree
```diff
@@ -0,0 +1,10 @@
+ARG UBUNTU_VERSION=18.04
```
gonuke: I think this file is now unused/superfluous?

CI/travis/install.sh
```diff
@@ -19,5 +19,10 @@ if [ "${PULL_REQUEST}" == "false" ]; then
   cd -
 fi
 
+# Build the double_down project if needed
+if [ "${DOUBLE_DOWN}" == "ON" ]; then
```
gonuke, line 23: Isn't double down already installed always now?

.circleci/config.yml
```diff
@@ -41,13 +44,21 @@ jobs:
...
           name: Testing DAGMC
           command: CI/travis/tests.sh
-  
+      # do not report failure for experimental double down builds
+      - when:
+          name: Experimental build check
+          condition: <<parameters.double_down>>
+          ignore_previous_status: true
+          steps:
+            - run: echo "DOUBLE DOWN BUILD FAILED"
```
gonuke: Won’t this always report failure?  It ignores the previous status, which I assume prevents the test from reporting failure.  However, it doesn’t use the previous status to decide whether to report this failure.  Do we need to add something to the conditional?

gonuke (changes requested):
I think this looks ok except that it may report a constantly failing status for DD

gonuke:
I see the check mark next to 8051e9b, but I can’t find a report on CircleCI to see what it reports.  Maybe we should merge this with it in the PR workflow, and then another PR to move it to the merge workflow?

CI/Dockerfile_4_embree
pshriwise: This whole file is, yes. Meant to remove it after my first attempt at adding this to CI. Thanks!

CI/travis/install.sh
pshriwise: Embree is always present in our docker files, but double down will be built each time as we're building using it's `develop` branch. It's build time is very short relative to MOAB & DAGMC.

.circleci/config.yml
pshriwise: That whole block isn't really useful. I'll remove it.

Based on previous failed attempts to get that build working, that block will always report the same thing but the build will still fail. The `ignore_previous_status` line just ensures that this step in the job always runs and the `echo` command always returning a status of 0 should always result in this step passing -- even if the `echo`'d message doesn't make a lot of sense.

CI/travis/install.sh
gonuke: Got it.  Yes, I forgot this was now on merge, rather than on PR.

.circleci/config.yml
pshriwise: https://github.com/svalinn/DAGMC/pull/693#issuecomment-667475806

Still curious about this @bam241. Will we be able to have a set of builds on merge with allowed failures that won't cause our `develop` branch show as failing CI?

pshriwise:
> I see the check mark next to 8051e9b, but I can’t find a report on CircleCI to see what it reports. Maybe we should merge this with it in the PR workflow, and then another PR to move it to the merge workflow?

Yeah that build seems to be viewable only on my CircleCI account (as opposed to Svalinn) for some reason?? Not sure why. I moved the build back to the PR section for now so it should be viewable. Then I can make one final commit after approval moving it back to our set of merge jobs.

.circleci/config.yml
```diff
@@ -72,7 +76,15 @@ workflows:
           filters:
             branches:
               ignore: develop
-  
+      - build_and_test:
+          name: "Double Down"
+          img: "18.04"
+          compiler: "gcc"
+          hdf5: "1.10.4"
+          moab: "master"
```
gonuke: This forces a build of MOAB `master` branch rather than using an image with a prebuilt MOAB.  Is that intentional?

.circleci/config.yml
gonuke: Not sure there is a clean way to allow failures in Circle right now

.circleci/config.yml
pshriwise: It is. There are changes to `RayHistory` in master that are necessary for double down.

gonuke (approved):

gonuke:
It's running the test in CI (???) - do you need the filters as well?

gonuke:
Now a test fails!?!?

pshriwise:
Works with a restart. But it was a legitimate test failure, not just a timeout. :disappointed:  That'll have to be addressed in another PR.

gonuke:
Merging before the test fails again :) - thanks @pshriwise

PR #694, opened 2020-07-23
Use cmake variable for Eigen3 location and include that dir

gonuke:
The original code over-specified the location of Eigen3 because it did not use the include dir found by CMake. This worked on a default install, but not on a custom install. This version should work in both cases.

Also remove the extra include in the `cpp` file.

### DIFF
--- a/news/PR-0694.rst
+++ b/news/PR-0694.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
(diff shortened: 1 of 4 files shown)


bam241:
@gonuke have you push your commit with the news?

gonuke:
News has now been pushed

bam241 (approved):
LGTM! 
thx @gonuke !
Merging...

PR #695, opened 2020-07-30
Modify MCNP6 patch files

ljacobson64:
Add DAGMC guard around changed if statement in history_neutral_high.F90
Add DAGMC guard around check for goto statement in charged_particle_history.F90
Add Franz Gallmeier patch

### DIFF
--- a/news/PR-0695.rst
+++ b/news/PR-0695.rst
@@ -1,0 +1,16 @@
+**Added:** None
+
+**Changed:**
+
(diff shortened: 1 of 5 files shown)


gonuke:
Thanks for the quick fix @ljacobson64 - since CI doesn't test this, I'm relying on a visual review.  (I am not going to build every version of MCNP to confirm these.)

Issue #699, opened 2020-08-25
uwuw_preproc fails to read new pyne material library


zxkjack123:
**Describe the Bug**<br/>
DAGMC `uwuw_preproc` fails to read new pyne material library. 

**To Reproduce**<br/>
I updated PyNE and DAGMC to the newest version (updated time 2020-08-25), and created a simple geometry with only materials of "Water" and "Vacuum" -> `imprint body all` -> `merge all` -> `export dagmc geom.h5m` -> `uwuw_preproc`. Then the failure read of pyne material library.

**Expected Behavior**<br/>
The pyne material library should be load and the materials should be rewrite to dagmc geometry.

**Screenshots or Code Snippets**<br/>
The python script for write PyNE h5 material library:
```
#!/usr/bin/python
from pyne.material import Material
from pyne.material_library import MaterialLibrary
print("Welcome!")
# define mat water 
watervec={10010:2,80160:1} # simple water
water = Material()
water.density = 1.0
water.metadata["name"] = "Water"
water.from_atom_frac(watervec)
#
lib = {"Water": water}
mat_lib=MaterialLibrary(lib)
mat_lib.write_hdf5("mvgeom_material_lib.h5", h5_overwrite=True)
print("All done!")
```

Output of "h5ls -r mvgeom_material_lib.h5"
```
/                        Group
/material_library        Group
/material_library/materials Group
/material_library/materials/composition Dataset {1/Inf}
/material_library/materials/composition_metadata Dataset {1/Inf}
/material_library/materials/nuclidelist Dataset {2}
```

Output of "h5dump mvgeom_material_lib.h5 ":
```
HDF5 "mvgeom_material_lib.h5" {
GROUP "/" {
   GROUP "material_library" {
      GROUP "materials" {
         DATASET "composition" {
            DATATYPE  H5T_COMPOUND {
               H5T_IEEE_F64LE "mass";
               H5T_IEEE_F64LE "density";
               H5T_IEEE_F64LE "atoms_per_molecule";
               H5T_ARRAY { [2] H5T_IEEE_F64LE } "comp";
            }
            DATASPACE  SIMPLE { ( 1 ) / ( H5S_UNLIMITED ) }
            DATA {
            (0): {
                  18.0106,
                  1,
                  3,
                  [ 0.111915, 0.888085 ]
               }
            }
         }
         DATASET "composition_metadata" {
            DATATYPE  H5T_VLEN { H5T_STD_I8LE}
[...]

PR #658, opened 2020-01-14
Update version to 3.2.0

ngranda:
This PR:
- bumps the dagmc version
- creates a changelog file with the consolidated news file

PR #700, opened 2020-08-27
updating PyNE version to v0.7.3

bam241:
As we have anew PyNE release that breaks `MaterialLibrary` one need to update the amalgamated version of PyNE shipped with DAGMC


bam241:
I am not sure I understand why this is failing for clang:
```
In file included from /root/build_dir/DAGMC/src/pyne/pyne.cpp:20:
/root/build_dir/DAGMC/src/pyne/pyne.h:5911:21: warning: 'read_e_bounds' has C-linkage specified, but returns user-defined type 'std::vector<double>' which is incompatible with C [-Wreturn-type-c-linkage]
std::vector<double> read_e_bounds(std::string e_bounds_file);
                    ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: warning: 'join_to_string<int>' has C-linkage specified, but returns user-defined type 'std::string' (aka 'basic_string<char>') which is incompatible with C [-Wreturn-type-c-linkage]
std::string pyne::join_to_string(std::vector<T> vect, std::string delimiter) {
                  ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:424:28: note: in instantiation of function template specialization 'pyne::join_to_string<int>' requested here
template std::string pyne::join_to_string(std::vector<int> vect, std::string delimiter);
                           ^
In file included from /root/build_dir/DAGMC/src/pyne/pyne.cpp:20:
/root/build_dir/DAGMC/src/pyne/pyne.h:5911:21: warning: 'read_e_bounds' has C-linkage specified, but returns user-defined type 'std::vector<double>' which is incompatible with C [-Wreturn-type-c-linkage]
std::vector<double> read_e_bounds(std::string e_bounds_file);
                    ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: warning: 'join_to_string<double>' has C-linkage specified, but returns user-defined type 'std::string' (aka 'basic_string<char>') which is incompatible with C [-Wreturn-type-c-linkage]
std::string pyne::join_to_string(std::vector<T> vect, std::string delimiter) {
                  ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:425:28: note: in instantiation of function template specialization 'pyne::join_to_string<double>' requested here
template std::string pyne::join_to_string(std::vector<double> vect,
                           ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: error: definition with same mangled name as another definition
std::string pyne::join_to_string(std::vector<T> vect, std::string delimiter) {
                  ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: note: previous definition is here
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: warning: 'join_to_string<std::__cxx11::basic_string<char> >' has C-linkage specified, but returns user-defined type 'std::string' (aka 'basic_string<char>') which is incompatible with C [-Wreturn-type-c-linkage]
std::string pyne::join_to_string(std::vector<T> vect, std::string delimiter) {
                  ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:427:28: note: in instantiation of function template specialization 'pyne::join_to_string<std::__cxx11::basic_string<char> >' requested here
template std::string pyne::join_to_string(std::vector<std::string> vect, std::string delimiter);
                           ^
In file included from /root/build_dir/DAGMC/src/mcnp/mcnp_funcs.cpp:1:
In file included from /root/build_dir/DAGMC/src/mcnp/mcnp_funcs.h:13:
In file included from /root/build_dir/DAGMC/src/mcnp/../uwuw/uwuw.hpp:1:
/root/build_dir/DAGMC/src/pyne/pyne.h:5911:21: warning: 'read_e_bounds' has C-linkage specified, but returns user-defined type 'std::vector<double>' which is incompatible with C [-Wreturn-type-c-linkage]
std::vector<double> read_e_bounds(std::string e_bounds_file);
                    ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: warning: 'join_to_string<int>' has C-linkage specified, but returns user-defined type 'std::string' (aka 'basic_string<char>') which is incompatible with C [-Wreturn-type-c-linkage]
std::string pyne::join_to_string(std::vector<T> vect, std::string delimiter) {
                  ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:424:28: note: in instantiation of function template specialization 'pyne::join_to_string<int>' requested here
template std::string pyne::join_to_string(std::vector<int> vect, std::string delimiter);
                           ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: warning: 'join_to_string<double>' has C-linkage specified, but returns user-defined type 'std::string' (aka 'basic_string<char>') which is incompatible with C [-Wreturn-type-c-linkage]
std::string pyne::join_to_string(std::vector<T> vect, std::string delimiter) {
                  ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:425:28: note: in instantiation of function template specialization 'pyne::join_to_string<double>' requested here
template std::string pyne::join_to_string(std::vector<double> vect,
                           ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: error: definition with same mangled name as another definition
std::string pyne::join_to_string(std::vector<T> vect, std::string delimiter) {
                  ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: note: previous definition is here
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: warning: 'join_to_string<std::__cxx11::basic_string<char> >' has C-linkage specified, but returns user-defined type 'std::string' (aka 'basic_string<char>') which is incompatible with C [-Wreturn-type-c-linkage]
std::string pyne::join_to_string(std::vector<T> vect, std::string delimiter) {
                  ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:427:28: note: in instantiation of function template specialization 'pyne::join_to_string<std::__cxx11::basic_string<char> >' requested here
template std::string pyne::join_to_string(std::vector<std::string> vect, std::string delimiter);
```

Any thoughts ? @pshriwise @gonuke

bam241:
I tried to remove the explicit instantiation but did not fix the problem...

pshriwise:
For the " has C-linkage specified" errors, the `read_e_bounds` function is in an `extern C` block but returns a C++ standard library type in the PyNE header file. That function either needs to return a C-type or be moved outside of that block.

pshriwise:
Oop, the "C-linkage" messages are just warnings here (I had more restrictive flags on, so they were errors for me). The other problem looks like it's due to the partial template specifications of `join_to_string`. Thinking...

bam241:
when removing the explicit instantiation of `join_to_string`
I get:
```
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: error: definition with same mangled name as another definition
std::string pyne::join_to_string(std::vector<T> vect, std::string delimiter) {
                  ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: note: previous definition is here
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: error: definition with same mangled name as another definition
std::string pyne::join_to_string(std::vector<T> vect, std::string delimiter) {
                  ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: note: previous definition is here
4 warnings and 2 errors generated.
make[2]: *** [src/pyne/CMakeFiles/pyne_dagmc-static.dir/pyne.cpp.o] Error 1
src/pyne/CMakeFiles/pyne_dagmc-static.dir/build.make:62: recipe for target 'src/pyne/CMakeFiles/pyne_dagmc-static.dir/pyne.cpp.o' failed
CMakeFiles/Makefile2:595: recipe for target 'src/pyne/CMakeFiles/pyne_dagmc-static.dir/all' failed
make[1]: *** [src/pyne/CMakeFiles/pyne_dagmc-static.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
1 warning generated.
[  4%] Built target mcnp_funcs
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: error: definition with same mangled name as another definition
std::string pyne::join_to_string(std::vector<T> vect, std::string delimiter) {
                  ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: note: previous definition is here
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: error: definition with same mangled name as another definition
std::string pyne::join_to_string(std::vector<T> vect, std::string delimiter) {
                  ^
/root/build_dir/DAGMC/src/pyne/pyne.cpp:411:19: note: previous definition is here
```

bam241:
this should be working now.

Please review carefully I am not sure I modify the usage of MaterialLibrary the way it was intended....

src/geant4/app/src/DagSolidMaterial.cc
```diff
@@ -93,39 +90,41 @@ std::map<int, G4Element*> get_g4elements(std::map<int, G4Isotope*> isotope_map)
...
 
   std::string name;
   int ncomponents;
 
   // loop over the PyNE material library instanciating as needed
-  for (it = material_library.begin() ; it != material_library.end() ; ++it) {
+  for (it = pyne_mat_map.begin() ; it != pyne_mat_map.end() ; ++it) {
     // get the material object
-    pyne::Material mat = it->second;
+    std::shared_ptr<pyne::Material> mat = it->second;
```
gonuke: Does this need to be a `shared_ptr`?  It seems most of the other changes are a consequence of this one.  Maybe keep that separate from this PR?

src/mcnp/mcnp_funcs.cpp
```diff
@@ -123,15 +123,16 @@ void dagmcwritefacets_(char* ffile, int* flen) { // facet file
   // out to the fcad file
   if (workflow_data->material_library.size() != 0) {
     // get the mat lib
-    std::map<std::string, pyne::Material> mat_lib = workflow_data->material_library;
-    std::map<std::string, pyne::Material> ::iterator it;
+    pyne::MaterialLibrary mat_lib = workflow_data->material_library;
```
gonuke: I know the original code did this, but does it need to be a copy?  Could this just be a reference?

src/mcnp/mcnp_funcs.cpp
```diff
@@ -222,13 +223,13 @@ void write_cell_cards(std::ostringstream& lcadfile, const char* mcnp_version_maj
       std::string mat_name = DMD->volume_material_property_data_eh[entity];
       // if we not vacuum or graveyard
       if (mat_name.find(vacuum_str) == std::string::npos && mat_name.find(graveyard_str) == std::string::npos) {
-        if (workflow_data->material_library.count(mat_name) == 0) {
+        pyne::Material material = workflow_data->material_library.get_material(mat_name);
```
gonuke: I am not a fan of this approach.  See my comment in `UWUW` about defining `workflow->material_library` as a `pyne::mat_map`....

src/mcnp/mcnp_funcs.cpp
```diff
@@ -317,16 +318,19 @@ void write_surface_cards(std::ostringstream& lcadfile) {
 
 // write out all the tally data from the uwuw file
 void write_material_data(std::ostringstream& lcadfile) {
-  std::map<std::string, pyne::Material> material_library = workflow_data->material_library;
+  pyne::MaterialLibrary material_library = workflow_data->material_library;
```
gonuke: create a reference rather than a copy?

src/uwuw/tests/uwuw_unit_tests_preprocessor.cpp
```diff
@@ -192,14 +192,14 @@ TEST_F(UWUWTest, materialMetadata) {
...
+  pyne::MaterialLibrary mat_lib = uwuw->material_library;
 
   // pull out the only material
-  pyne::Material mat = mat_lib["mat:CentreStack"];
+  pyne::Material mat = mat_lib.get_material("mat:CentreStack");
 
   EXPECT_EQ(mat.metadata["name"].asString(), "mat:CentreStack");
   EXPECT_EQ(mat.metadata["fluka_name"].asString(), "CENTREST");
-  EXPECT_EQ(mat.metadata["mat_number"].asInt(), 1);
+  EXPECT_EQ(mat.metadata["mat_number"].asInt(), 8);
```
gonuke: Do we know why this `mat_number` changed?  It probably doesn't matter, but it would be good to know why.

src/uwuw/uwuw_preprocessor.cpp
```diff
@@ -38,25 +40,26 @@ uwuw_preprocessor::~uwuw_preprocessor() {
...
-      std::cout << "writing material, " << mat.metadata["fluka_name"].asString();
+  if (verbose) {
+    pyne::mat_map mat_lib_obj = uwuw_material_library.get_mat_library();
+    pyne::mat_map::iterator it;
+
+    // loop over the processed material library and write each one to the file
+    for (it = mat_lib_obj.begin(); it != mat_lib_obj.end(); ++it) {
+      // the current material
+      std::shared_ptr<pyne::Material> mat = it->second;
+      std::cout << "writing material, " << mat->metadata["name"].asString();
```
gonuke: Originally, this output occurred immediately before writing each material.  I wonder if we should reword it slightly since the writing of the HDF5 material now occurs after all this output happens?  If someone uses this for debugging, there will be a gap in logical time between these messages and the actual writing???  not sure...

src/uwuw/uwuw_preprocessor.cpp
```diff
@@ -157,15 +160,16 @@ void uwuw_preprocessor::process_materials() {
 
     // check for missing materials
     if (found_grave == std::string::npos && found_vacuum == std::string::npos) {
-      if (material_library.count(mat_name) == 0) {
+      pyne::Material found_mat = material_library.get_material(mat_name);
```
gonuke: We should probably implement a `count` method or something equivalent on the PyNE side.  For now, can't we get `material_library.get_keys()` and check that?

gonuke (changes requested):
I think this all looks functionally correct.  I have a few design questions.

We should probably consider some additions to the PyNE class for convenience, including:
* `MaterialLibrary::iterator` (that is basically just a `mat_map` iterator)
* a `count` or `exists` method on `MaterialLibrary` so that we don't need to use `get_material` to test for existence
* define the `[]` operator for `MaterialLibrary` as a convenient way to access `get_material`

MigMagan:
It seems that the default hdf5 path has changed. Passing the path with the -m argument, the preprocessor finds the materials. 

```
mmagan@Kronem:~/Trelis/Tests/Example$ ~/DAGMC-BLD/DAGMC-PyNE-0.7/bin/uwuw_preproc Example.h5m -l /home/mmagan/.local/lib/python3.6/site-packages/pyne/nuc_data.h5
My lib size: 0
Loading file Example.h5m
Initializing the GeomQueryTool...
Using faceting tolerance: 0.001
Implicit Complement assumed to be Vacuum
material Lead was not found in the material library

mmagan@Kronem:~/Trelis/Tests/Example$ ~/DAGMC-BLD/DAGMC-PyNE-0.7/bin/uwuw_preproc Example.h5m -l /home/mmagan/.local/lib/python3.6/site-packages/pyne/nuc_data.h5 -m /material_library/materials
My lib size: 456
Loading file Example.h5m
Initializing the GeomQueryTool...
Using faceting tolerance: 0.001
Implicit Complement assumed to be Vacuum
Photon/Flux 3
particle: Photon
Flux
particle_name Photon
valid 1
Volume
Photon/Flux 3
particle: Photon
Flux
particle_name Photon
valid 1
Volume
Photon/Flux 3
particle: Photon
Flux
particle_name Photon
valid 1
Volume
Photon/Flux 3
particle: Photon
Flux
particle_name Photon
valid 1
Volume
```
I suggest changing the default hdf5 path:
 ```diff
--- a/src/uwuw/app/uwuw_preproc.cpp
+++ b/src/uwuw/app/uwuw_preproc.cpp
@@ -24 +24 @@ int main(int argc, char* argv[]) {
-  po.addOpt<std::string>("matlib_hdf5_path,m", "specify the HDF5 location of the material library (default "")",
+  po.addOpt<std::string>("matlib_hdf5_path,m", "specify the HDF5 location of the material library (default /material_library/materials )",
@@ -40 +40 @@ int main(int argc, char* argv[]) {
-    matlib_hdf5_path = "/materials";
+    matlib_hdf5_path = "/material_library/materials";
```

**Comment by gonuke** [comment]
Thanks @MigMagan - this is due to a change in how PyNE handles material libraries.  The patch proposed by @bam241 takes full advantage of those changes in PyNE and we'll hopefully have it converge in the next day or 2

**Comment by MigMagan** [comment]
Possibly specific to MCNP, the correct material numbers in cells and in material are not written. 

output of lcad following the guide example at https://svalinn.github.io/DAGMC/usersguide/uw2.html:
```
1 0 -11.35 imp:n=1
2 0 -11.35 imp:n=1
3 0 -12.3 imp:n=1
4 0 -12.3 imp:n=1
7 0  imp:n=0  $ graveyard
8 0  imp:n=1  $ implicit complement

(Surface numbers).

C materials from library
C name: mat:Lead
C density = 11.35000
m?
     82204 -1.5641e-01
     82206 -2.7189e+00
     82207 -2.5054e+00
     82208 -5.9692e+00
C name: mat:Lead/rho:12.3
C density = 12.30000
m?
     82204 -1.6950e-01
     82206 -2.9465e+00
     82207 -2.7151e+00
     82208 -6.4688e+00
```
Expected behavour:
```
1 1 -11.35 imp:n=1
2 1 -11.35 imp:n=1
3 2 -12.3 imp:n=1
4 2 -12.3 imp:n=1
7 0  imp:n=0  $ graveyard
8 0  imp:n=1  $ implicit complement

(Surface numbers)

C materials from library
C name: mat:Lead
C density = 11.35000
m1
     82204 -1.5641e-01
     82206 -2.7189e+00
     82207 -2.5054e+00
     82208 -5.9692e+00
C name: mat:Lead/rho:12.3
C density = 12.30000
m2
     82204 -1.6950e-01
     82206 -2.9465e+00
     82207 -2.7151e+00
     82208 -6.4688e+00
```

File: src/uwuw/uwuw_preprocessor.cpp (hunk 6 above)
Comment (zxkjack123) [inline src/uwuw/uwuw_preprocessor.cpp]:
```suggestion
      std::cout << "writing material, " << mat->metadata["name"].asString() << std::endl;
```

bam241:
@gonuke I changed the implementation and usage of material_library in the direction you suggested (I think)
it looks a bit ugly yet, because I would like to map the  `MaterialLibrary.begin()` to `MaterialLibrary.material_library.begin()` (and the `.end()`), to we would be able to iterate directly on the MaterialLibrary object (I am not sure this is exactly what you suggested, but seems somewhat similar...)

src/uwuw/tests/uwuw_unit_tests_preprocessor.cpp
bam241: it is because of the new way number get attributed when adding mat in the MaterialLibrary 
If a mat without number gets added, it gets: `mat.lib.size()+1` as a mat_number

bam241:
this is using a non-merged version of PyNE, that contains an update of the `PyNE::MaterialLibrary`
This should probably be merged after the PyNE one :)

gonuke:
> Possibly specific to MCNP, the correct material numbers in cells and in material are not written.
> 

@MigMagan  Is this still true after building a new version of `uwuw_preproc` using this patch?  The newest version of PyNE should automatically generate material numbers for every material in a material library.

gonuke:
@bam241 can you please bump this to 0.7.1 now that I've released it?

MigMagan:
> > Possibly specific to MCNP, the correct material numbers in cells and in material are not written.
> 
> @MigMagan Is this still true after building a new version of `uwuw_preproc` using this patch? The newest version of PyNE should automatically generate material numbers for every material in a material library.

I'm afraid I am seriously confused about how to use `uwuw_preproc`
The current user guide says we should use the nuc_data.h5 library that is build with PyNE. 
https://svalinn.github.io/DAGMC/usersguide/uw2.html
Doing so results in the behaviour I commented, with the material numbers missing. But using the `mat_lib.h5` file  included in the tests does work. So maybe it is just a matter that we do not use the `nuc_data.h5` library directly, and instead build a material database, either based on it or with python commands/scripts such as in Issue https://github.com/svalinn/DAGMC/issues/699 ?

Best.

bam241:
@MigMagan the issue you are describing should only occur if the materials in the MaterialLibrary don't have material number in their metadata.
(when no number is associated with a material, when calling the `mcnp()` method to the material input card a `?` will replace the number)
This should not append with the new version of PyNE (>0.7) (when building a material library the new API check for the presence of a material number for each added material, and specify one if not present. 
I am wondering if the `nuc_data.h5` you are using have been generated with a previous version of the PyNE...

gonuke:
@bam241 Are we sure the prebuilt nuc_data we provide is updated correctly?

bam241:
you are probably correct, `nuc_data` have probably not being updated since the MaterialLibrary update...

MigMagan:
@bam241 I just updated my `nuc_data.h5` to ensure I had it right, and experienced the same behaviour. It seems that in the data file, the only metadata present is the name. `prebuilt_nuc_data.h5` is 14 months old in the PyNE repo, so ithat may be the issue.

gonuke:
Thanks for the help in sorting this out @MigMagan - the next step is to do some "surgery" on the prebuilt nuc data to fix this problem

gonuke:
@MigMagan we've updated the prebuilt data in PyNE/data.  Can you confirm whether it works for you with these patches?

MigMagan:
> @MigMagan we've updated the prebuilt data in PyNE/data. Can you confirm whether it works for you with these patches?

Great, it works, `nuc_data.h5` now has material numbers and they get copied to the `lcad` file. 

There is a minor limitation; when using a material with more than one density, the material numbers are the same, and the m cards are duplicated. When the density difference is not big, or the problem is not using charged particles (see MCNP6 manual section 1.5.5, point 6), you can simply delete the duplicated material card. Otherwise, it is better to change the material number for one of the densities. This is easily handled and a warning is written, so it doesn't seem prioritary.

makeclean:
this must be a regression, since the uwuw system should automatically renumber the unique material/density pairs

gonuke:
There is no evidence that the UWUW system renumbered these unique pairs....

* the loop over unique pairs happens here: https://github.com/svalinn/DAGMC/blob/develop/src/uwuw/uwuw_preprocessor.cpp#L146 and calls `create_new_material()` to make a copy of the material and set the density correctly
* `create_new_material()` does not change the `mat_number` metadata at all, so there are no unique material numbers
* in the past, there were no material numbers assigned at all by PyNE, and each material was given a unique number, but NOT each unique material/density pair

bam241:
@gonuke this is now in.

@makeclean as @gonuke suggested we believe that before we only ensured uniqueness of the mat_number per mat (not per couple mat/density)

 this now ensure uniqueness of every mat/density couple as you suggested.

bam241:
@MigMagan I think you should not have the material number duplicating issue anymore :)

if you have time could you confirm ?

src/uwuw/uwuw.cpp
```diff
@@ -74,24 +74,10 @@ bool UWUW::check_file_exists(const std::string& filename) {
...
-  std::map<std::string, pyne::Material> library; // material library
-
-  const char* data_path = datapath.c_str();
-
-  if (!hdf5_path_exists(filename, data_path))
-    return library;
-
-  num_materials = get_length_of_table(filename, datapath);
-
-  for (int i = 0 ; i < num_materials ; i++) {
```
makeclean, line 87: this is the part of the code that gave unique material numbers that I was thinking of, appoligies - now that its done on the pairs thats probably better.

bam241: no worries, this lead to an good improvement, glad you bring that up :)

MigMagan:
> @MigMagan I think you should not have the material number duplicating issue anymore :)
> 
> if you have time could you confirm ?

Yes, confirmed working!

bam241:
this has been updated to the last `pyne/develop` version.

was wondering if we want to issue a new micro release for pyne so we can match a specific tag instead of a commit....
(specially as this required some fixes in PyNE...)

src/mcnp/mcnp_funcs.cpp
pshriwise: ping @bam241

pshriwise: I'd make that reference `const` as well if possible.

src/mcnp/mcnp_funcs.cpp
```diff
@@ -317,16 +319,18 @@ void write_surface_cards(std::ostringstream& lcadfile) {
...
-       it != material_library.end() ; ++it) {
-    pyne::Material new_material = (it->second);
-    std::string material_card = new_material.mcnp();
+
+  pyne::MaterialLibrary::iterator it;
+
+  for (it = material_library.begin(); it != material_library.end() ; ++it) {
+    std::shared_ptr<pyne::Material> new_material = (it->second);
+    std::string material_card = new_material->mcnp();
     lcadfile << material_card;
```
pshriwise:
```suggestion
    lcadfile << (it->second)->mcnp();
```

src/pyne/update_pyne.sh
```diff
@@ -2,8 +2,9 @@
 
 set -e
 
+branch="develop"
```
pshriwise: Is there a benefit to creating a variable here instead of using "develop" directly in the clone command below?

src/uwuw/uwuw.cpp
```diff
@@ -74,24 +74,10 @@ bool UWUW::check_file_exists(const std::string& filename) {
...
-    pyne::Material mat; // from file
-    mat.from_hdf5(filename, datapath, i);
-    // renumber material number by position in the library
-    mat.metadata["mat_number"] = i + 1;
-    library[mat.metadata["name"].asString()] = mat;
-  }
+pyne::MaterialLibrary UWUW::load_pyne_materials(std::string filename, std::string datapath) {
+  pyne::MaterialLibrary library(filename, datapath); // material library
 
+  std::cout << "My lib size: " << library.size() << std::endl;
```
pshriwise: Do we always want to see this output?

pshriwise: If this was for debug purposes, let's remove it.

pshriwise (changes requested):
There are some nice design updates that come with this. Thanks @bam241! 

Just a few small comments, but I think this is very close to ready.

src/geant4/app/include/ExN01DetectorConstruction.hh
```diff
@@ -33,7 +34,7 @@ class ExN01DetectorConstruction : public G4VUserDetectorConstruction {
   // particle filters for tallies
   std::map<std::string, G4SDParticleFilter*> particle_filters;
 
-  void BuildParticleFilter(std::string particle_name);
+  void BuildParticleFilter(std::vector<std::string> particle_name);
```
pshriwise, line 37: These don't seem to have anything to do with the changes to the PyNE `MaterialLibrary` API. Is it part of the Geant4 API update?

bam241: no it is part of the PyNE::Tally API change :)

src/mcnp/mcnp_funcs.cpp
bam241: do you mean ?
`lcadfile << (it->second)->mcnp();`

src/geant4/app/include/ExN01DetectorConstruction.hh
pshriwise: Ah, I see. :+1:

src/mcnp/mcnp_funcs.cpp
pshriwise: Haha yes. Sorry. Editing.

src/pyne/update_pyne.sh
bam241: I liked it because it allows to play with it easily, and try out other version when necessary.
I can flip that back, don't have a strong filling here.

pshriwise: I guess I'd prefer to keep it in the same line unless it's used in multiple places or becomes an external parameter for the script. Less to parse mentally.

pshriwise:
Was just building this branch against OpenMC and ran into the following issues:

```shell
/home/shriwise/opt/openmc/src/src/dagmc.cpp: In function ‘bool openmc::get_uwuw_materials_xml(std::string&)’:
/home/shriwise/opt/openmc/src/src/dagmc.cpp:69:21: error: passing ‘const pyne::MaterialLibrary’ as ‘this’ argument discards qualifiers [-fpermissive]
   69 |     for (auto mat : mat_lib) { ss << mat.second.openmc("atom"); }
      |                     ^~~~~~~
In file included from /home/shriwise/opt/dagmc/include/uwuw.hpp:1,
                 from /home/shriwise/opt/openmc/src/src/dagmc.cpp:15:
/home/shriwise/opt/dagmc/include/pyne.h:5550:12: note:   in call to ‘pyne::MaterialLibrary::iterator pyne::MaterialLibrary::begin()’
 5550 |   iterator begin() { return material_library.begin(); }
      |            ^~~~~
/home/shriwise/opt/openmc/src/src/dagmc.cpp:69:21: error: passing ‘const pyne::MaterialLibrary’ as ‘this’ argument discards qualifiers [-fpermissive]
   69 |     for (auto mat : mat_lib) { ss << mat.second.openmc("atom"); }
      |                     ^~~~~~~
In file included from /home/shriwise/opt/dagmc/include/uwuw.hpp:1,
                 from /home/shriwise/opt/openmc/src/src/dagmc.cpp:15:
/home/shriwise/opt/dagmc/include/pyne.h:5551:12: note:   in call to ‘pyne::MaterialLibrary::iterator pyne::MaterialLibrary::end()’
 5551 |   iterator end() { return material_library.end(); }
      |            ^~~
/home/shriwise/opt/openmc/src/src/dagmc.cpp:69:49: error: ‘class std::shared_ptr<pyne::Material>’ has no member named ‘openmc’
   69 |     for (auto mat : mat_lib) { ss << mat.second.openmc("atom"); }
      |                                                 ^~~~~~
/home/shriwise/opt/openmc/src/src/dagmc.cpp: In function ‘void openmc::load_dagmc_geometry()’:
/home/shriwise/opt/openmc/src/src/dagmc.cpp:161:44: error: ‘class pyne::MaterialLibrary’ has no member named ‘empty’; did you mean ‘emtpy’?
  161 |   bool using_uwuw = !uwuw.material_library.empty();
      |                                            ^~~~~
      |                                            emtpy
/home/shriwise/opt/openmc/src/src/dagmc.cpp:242:60: error: ‘class std::shared_ptr<pyne::Material>’ has no member named ‘metadata’
  242 |           int mat_number = uwuw.material_library[uwuw_mat].metadata["mat_number"].asInt();
      |                                                            ^~~~~~~~
```

Seems like there might be something up with the `MaterialLibrary` iterator implementation?

Also a couple of methods missing that were there before.

bam241:
the problem comes from the material lib, it is now storing shared_ptr<Material> vs Material before.

before dagmc was not leveraging the PyNE::MaterialLirbary but was storing it as a map<string, Material>

bam241:
and there might be some missing methods...

pshriwise:
> the problem comes from the material lib, it is now storing shared_ptr vs Material before.

I'd be ok with updating this method call and a move to `shared_ptr` for storing those objects makes a lot of sense. It seems like there would be some way to get the iterator to access the object itself though :thinking: 

> before dagmc was not leveraging the PyNE::MaterialLirbary but was storing it as a map<string, Material>

Fair enough, but it seems like we should still be able to iterate over a `MaterialLibrary` object if we could before...

I do see iterators in the new `MaterialLibrary`. Some of the compiler errors indicate that overloads of the iterator methods with `const` aren't available I think.

bam241:
for .openmc() and .metadata() methods I think replacing `.` by `->` should do the trick.
=> I am a little anxious to add a method to get the object, that you would have to call eventually (as I don't think it would make sense for the iterator of the MaterialLibrary to point to the object not the `shared_ptr`) (Might be wrong there...)

for the missing overloading can you point to somewhere we see those, so we can update the pyne::MaterialLibrary accordingly ?

bam241:
we can also add a empty() for convenience that return something based on the size of the MaterialLibrary

pshriwise:
> for the missing overloading can you point to somewhere we see those, so we can update the pyne::MaterialLibrary accordingly ?

https://github.com/pyne/pyne/blob/403f4807ab9ed5d2d5d78a87d808892bbb04ec3c/src/material_library.h#L150-L153

bam241:
@gonuke @pshriwise I think I dressed all your comments.

news/PR-0700.rst
```diff
@@ -0,0 +1,14 @@
+**Added:** None
+
+**Changed:** 
+- Update amalgamated PyNE version to v0.7.3
```
pshriwise, line 5: Need a blank line between the header and content here.

src/fluka/fluka_funcs.h
```diff
@@ -12,8 +12,8 @@
 #include "moab/Interface.hpp"
 #include "moab/CartVect.hpp"
 #include "DagMC.hpp"
-#include "../pyne/pyne.h"
-#include "../uwuw/uwuw.hpp"
+#include "pyne.h"
+#include "uwuw.hpp"
```
pshriwise: This is a good update. Thanks!

src/uwuw/uwuw_preprocessor.cpp
```diff
@@ -161,11 +163,17 @@ void uwuw_preprocessor::process_materials() {
...
 
       // make a new material object with the appropriate density & name
-      pyne::Material new_material = create_new_material(material_library[mat_name],
+      pyne::Material new_material = create_new_material(found_mat,
                                                         density);
-      uwuw_material_library[*s_it]  = new_material;
+
+      // Renumbering the material will ensure uniqueness of material number for each mat/density
+      // combinason in the library
```
pshriwise:
```suggestion
      // combination in the library
```

pshriwise (changes requested):
Just a few small things and then I think this is good to go!

bam241:
@pshriwise I think I addressed your last comments

bam241:
I don't know why my mod is shown in the commit but not in the PR diff :(

pshriwise:
> I don't know why my mod is shown in the commit but not in the PR diff :(

Huh, yeah I'm not sure either. But I see in the history your changes went in.

src/fluka/fluka_funcs.h
bam241: this is in develop should not be there :(

bam241:
Holly F... I made even more a mess....

bam241:
@pshriwise I think I cleaned it, cherry-pick was the answer....
could you take 5 to check I did't add any mess ? 
I'll do that after 2:30 pm as well

pshriwise (approved):
Looks good after removing those `develop` commits :)

bam241:
yes I agree.

Please move forward whenever.

the next action is to merge #658 and update the version changelog with those changes.

After that I think we can start the count down before the release.

bam241:
@gonuke is there anything still holding this ?

bam241:
@pshriwise @gonuke I think this is hanging for too long.

Maybe we can move on with this.

I'll issue a correction PR if @gonuke has some additional comments about this.

gonuke (approved):
Thanks @bam241 - sorry I lost touch with this review (and many others!)

gonuke:
Thanks @bam241

PR #701, opened 2020-09-09
Only require fortran compiler if building MCNP5/6

pshriwise:
## Description
As of #686 we no longer have a strict dependency on a Fortran compiler unless we're building MCNP executables.

## Motivation and Context
For codes requiring only the C++ based libraries we build. These changes reduce our dependency list.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 project(DAGMC)
 cmake_minimum_required(VERSION 3.1)
-enable_language(Fortran)
+enable_language(CXX)
 
 # Set DAGMC version
 set(DAGMC_MAJOR_VERSION 3)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -16,6 +16,11 @@ dagmc_setup_build()
 
 dagmc_setup_options()
 
+if(BUILD_MCNP5 OR BUILD_MCNP6)
```
gonuke, line 19: Maybe Fluka also?

gonuke (changes requested):
Thanks @pshriwise - not sure if we need Fortran for Fluka also? @makeclean ?

makeclean:
No Fortran needed for fluka

bam241 (approved):
LGTM !

pshriwise:
Testing this with an MCNP build to make sure it isn't breaking that... will confirm one way or the other shortly.

pshriwise:
Successfully built DAG-MCNP6 and tests passed. :heavy_check_mark:

gonuke (approved):

gonuke:
Thanks @pshriwise

shimwell:
This is great, just perfect for OpenMC

Would we also need another release to target this in a conda recipe

PR #704, opened 2020-10-15
Start of a metadata management script

pshriwise:
## Description
Adds a Python script based on PyMOAB for viewing and updating DAGMC metadata on a .h5m file.

## Motivation and Context
Mistakes in metadeta are often made. If our only option for updating them is in Trelis, this means re-faceting and exporting the entire model again. This script provides an alternative to that.

### DIFF
--- a/news/PR-704.rst
+++ b/news/PR-704.rst
@@ -1,0 +1,13 @@
+**Added:**
+
+* Python script for DAGMC metadata
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/scripts/dagmc_metadata
+++ b/scripts/dagmc_metadata
@@ -67,19 +67,9 @@ def print_groups(mb):
         group_id = mb.tag_get_data(tags['id'], group, flat=True)[0]
         output += "Group {}: {}\n".format(group_id, name)
 
-        volumes = []
-        surfaces = []
-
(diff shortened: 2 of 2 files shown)


gonuke:
This is a great step forward, but perhaps raises the question of some core DAGMC methods that can be reused in various places. Specifically, dagmc_viz and dagmc_stats probably have overlapping use of key infrastructure.

pshriwise:
A fair point. Is there interest in incorporating some of those into this repo or unifying them in some way?

gonuke:
Maybe we should discuss on our next call?  Probably a matter of defining some common elements to go in a `dagmc` module distributed with DAGMC (this repo) and then different front ends that depend on and extend it?

makeclean:
What does dagmc_stats do?

gonuke:
dagmc_stats is a project begun about 18 months ago to report a variety of statistics about a DAGMC model.  One of its primary motivations was to allow users to share characteristics of their model with us without sharing the model:
- num of vertices, triangles, curves, surfaces, volumes, groups
- pdf for (min, max, median, mean, histogram?)
    - num of surf/volume
    - num of tri/surf
    - valence/vert

@kkiesling has a student extending it for some of her use:
- roughness of a surface

Anything else you want?

ahnaf-tahmid-chowdhury:
@pshriwise, are you still interested working on this? I think we can move this to PyDAGMC?

gonuke:
I agree that this is superseded by pydamgc

PR #707, opened 2020-10-21
Update to Git SHA instead of SVN revision number.

pshriwise:
## Description
I noticed today that the value `"$Rev"` was present in the configured and installed `DagMCVersion.hpp` file, which didn't seem right. We were last using SVN in ~2010.

This PR updates commit revision tracking from an SVN revision number to a Git SHA. I've deprecated the `DagMC::interface_revision` method and added a `DagMC::git_sha` method. I'm fine with simply removing the `interface_revision` function, but it seemed proper to deprecate it first and since I doubt anyone is using it I don't think it's doing any harm either way


bam241 (approved):
LGTM!

src/dagmc/DagMC.cpp
```diff
@@ -112,28 +112,7 @@ float DagMC::version(std::string* version_string) {
...
-    if (endptr == interface_string.c_str() + key_str.size()  // parsing end
-        || ((result == LONG_MAX || result == LONG_MIN)
-            && errno == ERANGE)) {  // checking range
-      std::cerr << "Not able to parse revision number" << std::endl;
-      exit(1);
-    }
-    return result;
-  }
-  return result;
+  return 0;
```
bam241: maybe move this in the header ?

gonuke (approved):
LGTM!

bam241:
thx @pshriwise ! merging!

PR #708, opened 2020-10-27
Fix for setting precision and overlap tolerance.

pshriwise:
@bam241 noticed that a couple of calls went missing for setting the numerical precision and overlap tolerance in #693. This beings those calls back in one of the `DagMC` constructors and makes those calls on the DagMC object itself rather than on the ray tracer.


bam241 (approved):
LGTM! thx for addressing those so quickly!

PR #711, opened 2020-11-16
Removed the now-redundant data member defaultFacetingTolerance

helen-brooks:
As it says on the can. 

Removed a data member from the class DagMC that was not used anywhere and is presumably just a relic. No behaviour changes; purpose of PR is just housekeeping.


gonuke (approved):
Thanks @helen-brooks !  I think this is a holdover from the days when we might facet during initialization of the MC code, so is definitely due to be axed.

PR #712, opened 2020-11-18
Release 3.2.0

bam241:
This is the PR to issue the Release 3.2.0.

This should not be merged before Monday November 23.
A tag corresponding to the release 3.2.0 will then be issued.

Please pull the `3.2.0-rc2` branch or the `3.2.0-rc2` to test it and verify everything is still working for you.

A full changelog from the previous release can be found on the Release candidate page: https://github.com/svalinn/DAGMC/releases/tag/3.2.0-rc2

If no issue is found this should be merged on November 23.

PR #713, opened 2020-11-25
Updating objects in some of the fluka_funcs functions.

pshriwise:
I took a crack at updating the calls for the PyNE material library in the `fluka_funcs.*` files that were failing to compile, as reported by @ljacobson64  in #712.

The issues with the update from `Tally::particle_name` to `Tally::particle_names` in PyNE is more subtle. What I have in there is likely insufficient. Should those FLUKA routines create multiple tallies if multiple particle names exist? Should that routine fail? Any guidance here is appreciated @bam241 @makeclean @gonuke.

### DIFF
--- a/src/fluka/fluka_funcs.cpp
+++ b/src/fluka/fluka_funcs.cpp
@@ -985,7 +985,16 @@ void fludag_all_tallies(std::ostringstream& mstr, std::map<std::string, pyne::Ta
   std::list<std::string> tally_parts;
   std::string tally_id;
   for (it = tally_map.begin() ; it != tally_map.end() ; ++it) {
-    tally_id = (it->second).tally_type + "/" + (it->second).particle_names[0];
+
+    std::string particle_name = (it->second).particle_names.at(0);
+
+    if ((it->second).particle_names.size() > 1) {
+      std::cerr << "Warning: Multiple particles specified on a tally. "
+                   "Only the first particle (" << particle_name <<
+                   ") will be used." << std::endl;
+    }
+
+    tally_id = (it->second).tally_type + "/" + particle_name;
     if (std::count(tally_parts.begin(), tally_parts.end(), tally_id) == 0) {
       tally_parts.insert(tally_parts.end(), tally_id);
     }
@@ -1007,7 +1016,7 @@ void fludag_all_tallies(std::ostringstream& mstr, std::map<std::string, pyne::Ta
     ss << ".";
     tally.entity_name = ss.str();
 
-    std::string tally_id = tally.tally_type + "/" + tally.particle_names[0];
+    std::string tally_id = tally.tally_type + "/" + tally.particle_names.at(0);
 
     std::list<std::string>::iterator iter = std::find(tally_parts.begin(), tally_parts.end(), tally_id);
 
--- a/src/fluka/fluka_funcs.cpp
+++ b/src/fluka/fluka_funcs.cpp
@@ -988,8 +988,8 @@ void fludag_all_tallies(std::ostringstream& mstr, std::map<std::string, pyne::Ta
 
     if ((it->second).particle_names.size() > 1) {
       std::cerr << "Warning: Multiple particles specified on a tally. "
-                   "Only the first particle (" << particle_name <<
-                   ") will be used." << std::endl;
+                "Only the first particle (" << particle_name <<
+                ") will be used." << std::endl;
     }
 
     tally_id = (it->second).tally_type + "/" + particle_name;
--- a/src/fluka/fluka_funcs.cpp
+++ b/src/fluka/fluka_funcs.cpp
@@ -815,7 +815,8 @@ void fludag_write(std::string matfile, std::string lfname) {
 
   // ASSIGNMA Cards
   std::ostringstream assignma_str;
-  fludagwrite_assignma(assignma_str, workflow_data.material_library);
+  pyne::mat_map mat_lib = workflow_data.material_library.get_mat_library();
+  fludagwrite_assignma(assignma_str, mat_lib);
 
   // get the importances
   std::ostringstream importance_str;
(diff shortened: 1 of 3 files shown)


src/fluka/fluka_funcs.cpp
```diff
@@ -984,7 +985,7 @@ void fludag_all_tallies(std::ostringstream& mstr, std::map<std::string, pyne::Ta
   std::list<std::string> tally_parts;
   std::string tally_id;
   for (it = tally_map.begin() ; it != tally_map.end() ; ++it) {
-    tally_id = (it->second).tally_type + "/" + (it->second).particle_name;
+    tally_id = (it->second).tally_type + "/" + (it->second).particle_names[0];
```
makeclean: Originally a pyne::Tally object had a single particle that described it, I take it from the new name that they now have multiple names?

makeclean (commented):
Just one query about tally_names?

src/fluka/fluka_funcs.cpp
pshriwise: That was my understanding too, yeah. Though from these lines in pyne it looks like only single-particle tallies are intended for FLUKA, so maybe what we have here is ok?

https://github.com/pyne/pyne/blob/010c3d3bd2b1530aea4d7e00f2771a129a028673/src/tally.cpp#L600-L603

Hoping for @bam241's thoughts here as it looks like was part of that update.

src/fluka/fluka_funcs.cpp
```diff
@@ -833,7 +834,8 @@ void fludag_write(std::string matfile, std::string lfname) {
   if (workflow_data.material_library.size() != 0) {
     // write COMPOUND CARDS
     std::ostringstream material_str;
-    fludag_all_materials(material_str, workflow_data.material_library);
+    pyne::mat_map mat_lib = workflow_data.material_library.get_mat_library();
+    fludag_all_materials(material_str, mat_lib);
```
bam241: why not pass the material library directly as you can use it as a std::map

src/fluka/fluka_funcs.cpp
```diff
@@ -954,7 +955,7 @@ void fludagwrite_assignma(std::ostringstream& ostr,
       // otherwise we assume uwuw
       // if you arent graveyard or vacuum you must be in the uwuw library
       if (mat_prop.find("Graveyard") == std::string::npos && mat_prop.find("Vacuum") == std::string::npos) {
-        pyne::Material material = pyne_map[mat_prop];
+        pyne::Material material = *(pyne_map.at(mat_prop));
```
bam241: if you keep the pyne::MaterialLibrary you could use get_material method and or the [] operator

src/fluka/fluka_funcs.cpp
```diff
@@ -1027,17 +1037,18 @@ void fludag_all_tallies(std::ostringstream& mstr, std::map<std::string, pyne::Ta
...
 
   // loop over all materials, summing
-  std::map<std::string, pyne::Material>::iterator nuc;
+  pyne::mat_map::const_iterator nuc;
   for (nuc = pyne_map.begin(); nuc != pyne_map.end(); ++nuc) {
-    unique = unique + (nuc->second);
+    unique = unique + *(nuc->second);
   }
   // now collapse elements
   unique = unique.collapse_elements(exception_set);
```
bam241: I am not 100% on what is done there.
but with the Pyne::MaterialLibrary you can use `get_nuclist()` to get a set of the nuclides (if required...)

src/fluka/fluka_funcs.cpp
pshriwise: Are you saying we should leave the function arguments as they were and pass this the `mat_lib` variable?

bam241: I think we should pass directly the material_library object
```suggestion
    fludag_all_materials(material_str, workflow_data.material_library);
```

pshriwise: Ah, nevermind. I see what you're saying. Changes pending...

src/fluka/fluka_funcs.cpp
makeclean: well thats all that does, but that functionality from materialibrary must be new?

pshriwise: Yeah I'm honestly not sure about what's being done here either. In terms of the objects, it's iterating over the material objects and adding them. `nuc` is an odd variable name though. You ok with a change to `it` as well here?

makeclean: what it does is use pyne::material addition to build a material object that has every nuclide present in the entire problem.

makeclean: that 'special' unique material then does collapse_elements() to get a material that contains the special set of nuclides that FLUKA is aware of

makeclean: but that material object is never written

makeclean: nuc should be material or mat or something like that

bam241 (approved):
Thx @pshriwise for addressing those.
LGTM!

src/fluka/fluka_funcs.cpp
bam241: > makeclean 
> well thats all that does, but that functionality from materialibrary must be new?

Yes it has been added when we created the C++ API

pshriwise: Done!

pshriwise:
This looking ok now @bam241 @makeclean?

bam241:
Fine by me!
Waiting for @makeclean approval to merge !

bam241:
LGTM!
In order to keep this release active, I'll be merging!
If additional comments rise, they will be addressed in an other PR :)

bam241:
thx @pshriwise for addressing those issues.

PR #714, opened 2020-11-30
Adding methods for graveyard removal and creation

### LINKED ISSUE #649: Automatic graveyard generation
Graveyards can be problematic when working with DAGMC geometries (particularly for viz), but are necessary to robustly determine particle termination in DAGMC simulations. The graveyard volume can be automatically generated at runtime in a DAGMC simulation if a graveyard volume is not already present.

A rectangular shell prism can then be generated that is just larger than the size of the model.

A start for something like this can be found [here](https://github.com/pshriwise/openmc/blob/9ef18f79ed3176a1dd9d487f6ca240fe430c2d68/src/cad.cpp#L9).

pshriwise:
## Description
This PR adds the following public methods related to graveyard management:

  - `DagMC::remove_graveyard`
  - `DagMC::create_graveyard`
  - `DagMC::has_graveyard`
  - `DagMC::get_graveyard_group`
  - `DagMC::category_tag`

We may want to have a discussion about moving some of these capabilities into the `GeomTopoTool` class in MOAB. If that's the case, that's fine by me but I figured this PR contains the framework and a place to have that discussion :smiley:

## Motivation and Context
These capabilities were added so that models can be exported from Cubit/Trelis with or without a graveyard. The new API functions allow external codes to automatically generate a graveyard volume if one is not present on the model. It also allows an external application to remove the graveyard if needed.

In addition to reducing workflow overhead, the ability to work with models that don't contain a graveyard makes visualization much more convenient.

## Other Information
This PR has a few byproducts that come with it:

  - the `dagmcMetaData::to_lower` method now relies on a common call to the `lowercase_str` method which is provided in the newly added `util.hpp` file. The `lowercase_str` method has been added as part of the `moab` namespace to avoid populating the common namespace (though we may want to have a discussion about our overall namespace situation sometime in the future).
  - the addition of a BBOX struct to the `DagMC` class. This is purely used as a convenience for calculating the global bounding box of the problem when creating a graveyard.
 - the addition of a `DagMC::category_tag` method. This adds some MOAB-specific syntax to DAGMC, but I wasn't sure how else to handle it and it's needed to find the graveyard volume.
  - this adds a test file of a pincell, `pincell.h5m`, that was created in Trelis to ensure that the `remove_graveyard` method works without leaving orphaned entities or entity sets behind.

Resolves #649


src/dagmc/util.hpp
```diff
@@ -0,0 +1,14 @@
+#ifndef _DAGMC_UTIL
+#define _DAGMC_UTIL
+
+#include <string>
+
+namespace moab { // TODO: separate into a new namespace
```
bam241: why not creating right away a `dagmc_tool` namespace ?

src/dagmc/DagMC.cpp
```diff
@@ -212,6 +214,423 @@ ErrorCode DagMC::setup_indices() {
...
+      graveyard_group = group;
+      graveyard_count++;
+    }
+  }
+
+  // there should not be more than one graveyard group
+  if (graveyard_count > 1) {
+    MB_CHK_SET_ERR(MB_FAILURE,
+                   "More than one graveyard group is present in the model");
+  }
```
bam241: as it is a failure and we don't return how much there is, is it worth to continue looping after founding the second graveyard ?

src/dagmc/DagMC.cpp
```diff
@@ -212,6 +214,423 @@ ErrorCode DagMC::setup_indices() {
...
+
+  // get the graveyard volume
+  Range graveyard_vols;
+  rval =
+      moab_instance()->get_entities_by_handle(graveyard_group, graveyard_vols);
+  MB_CHK_SET_ERR(rval, "Failed to get the graveyard volume(s)");
+  sets_to_delete.merge(graveyard_vols);
+
+  // get the implicit complement, it's children will need updating
+  EntityHandle ic = 0;
```
bam241: not sure if `ic` is a commun shortcut for implicit complement, but maybe it is worse expending it a bit ? `icomplement` ? or ..?

src/dagmc/DagMC.cpp
```diff
@@ -212,6 +214,423 @@ ErrorCode DagMC::setup_indices() {
...
+  // remove existing graveyard
+  if (overwrite) {
+    remove_graveyard();
+  }
+
+  // if a graveyard already exists and we aren't overwriting it,
+  // report an error
+  if (has_graveyard() && !overwrite) {
+    MB_CHK_SET_ERR(MB_FAILURE, "Graveyard already exists");
+  }
```
bam241, line 403: maybe this can be combined together ?
```
if (overwrite) {}
else if(has_graveyard()) {}
```

src/dagmc/DagMC.cpp
```diff
@@ -212,6 +214,423 @@ ErrorCode DagMC::setup_indices() {
...
+                 "Failed to add the graveyard volume to the graveyard group");
+
+  /// SURFACE CREATION ///
+
+  // expand the box a bit
+  for (int i = 0; i < 3; i++) {
+    double bump = 10 * numerical_precision();
+    box.upper[i] += bump;
+    box.lower[i] -= bump;
+  }
```
bam241: maybe the 2 loop to increase the size of of the box could be part of a dedicated method ?

src/dagmc/DagMC.cpp
```diff
@@ -212,6 +214,423 @@ ErrorCode DagMC::setup_indices() {
...
+      "Failed to build accel. data structure for the new implicit complement");
+
+  // re-initialize indices
+  rval = setup_indices();
+  MB_CHK_SET_ERR(rval, "Failed to setup indices after graveyard creation");
+
+  return rval;
+}
+
+ErrorCode DagMC::box_to_surf(double llc[3], double urc[3],
```
bam241: `const &` for the `double[3]` arguments ?

src/dagmc/DagMC.cpp
bam241: mayb as part of the `BOX` struc ?

src/dagmc/util.hpp
```diff
@@ -0,0 +1,14 @@
+#ifndef _DAGMC_UTIL
+#define _DAGMC_UTIL
+
+#include <string>
+
+namespace moab {  // TODO: separate into a new namespace
```
bam241: (Already mentionned it... but adding it here for simplicity) why not already introduce a `dagmc_tool` namespace ?

src/dagmc/dagmcmetadata.cpp
```diff
@@ -339,9 +341,7 @@ void dagmcMetaData::parse_tally_volume_data() {
 }
 
 std::string dagmcMetaData::to_lower(std::string input) {
-  for (int i = 0; i < input.size(); i++) {
-    input[i] = std::tolower(input[i]);
-  }
+  moab::lowercase_str(input);
```
bam241: I don't like the use of the `moab` namespace here because one can believe this comes from MOAB where it is actually a dagmc method.

bam241 (changes requested):
Looking good.
I really like the exaustive tests implemented !
Some few comments on implementation details, nothing major tho !
thx @pshriwise !

src/dagmc/DagMC.cpp
pshriwise: - [x] `ic` is one we've used frequently in the past, but there's no reason not to be more explicit here. I'll update that

src/dagmc/DagMC.cpp
pshriwise: I don't follow... this error should only be raised in the case that a graveyard already exists and the overwrite flag is set to false. Separating the two would have a different meaning than the one that exists here. 

That being said, the check for `!overwrite` isn't really necessary, as the block above removes the graveyard if `overwrite` is `true`, meaning that at this point if a graveyard already exists we can raise an error.

src/dagmc/DagMC.cpp
pshriwise: - [x] Good suggestion. Better to put something like that in the struct for use elsewhere.

src/dagmc/DagMC.cpp
pshriwise: If it were a data structure that could get really large, then yes we'd want to do that. I don't think it's worth it here though.

pshriwise: - [x] Making them const is probably good for peace of mind though :)

src/dagmc/dagmcmetadata.cpp
pshriwise: - [x] I don't disagree, but this happens a lot in DAGMC. In fact, the entire `DagMC` class is in the `moab` namespace. I was mostly trying to follow suit here. As an incremental step, I'll update this to a `dagmc_util` namespace. There might be an even better approach here though. I'll think on it a bit.

src/dagmc/DagMC.cpp
pshriwise: This is after the loop, which does go over all groups to count the number of graveyards found. A good point that we don't return how many graveyard groups were found though -- I've added that to the error message here.

pshriwise:
I think I've touched on all of your comments at this point @bam241. Thanks for your review! I think it's ready for another round.

src/dagmc/DagMC.cpp
bam241: just line 392 you had a 
`if(overwrite){}`
and l 398
`if (has_greaveyard() && !overwrite){}`

my suggestion was to combine the two into a if() else if()

but your new implementation resolves it :)

bam241 (approved):
Thank you for this @pshriwise !
This looks good to me.

I'll merge by the end of the week if no-one argues :)

pshriwise:
Please wait to merge. This doesn't quite work with double-down yet -- it just isn't reflected in the checks here b/c we don't test against double-down for PRs.

bam241:
@pshriwise for safety I converted this into a draft. so it won't be merged...

gonuke:
@pshriwise - we're gearing up for a release and you tagged this for a 3.2.1 release (we'll probably jump to 3.3.0).

Do you want to work this in?

pshriwise:
Yeah, I'd like to if possible. I'll rebase it and push now.

doc/CHANGELOG.rst
```diff
@@ -7,16 +7,23 @@ DAGMC Changelog
...
    * CMake option to checkout PyNE submodule automatically (#734)
    * GitHub Action to build and upload Docker images. (#746, #748, #754, #757, #758, #759, #765, #767)
    * Enforcing usage of Python3 for PyNE amalgamation. (#773)
+   * DagMC methods for creation and removal of the graveyard volume (#714)
+     * DagMC::remove_graveyard
+     * DagMC::create_graveyard
+     * DagMC::has_graveyard
+     * DagMC::get_graveyard_group
+     * DagMC::category_tag
+     * DagMC::box_to_surf (private method)
```
gonuke: Not sure we need all this detail in the CHANGELOG

gonuke (changes requested):
oops - I thought I'd submitted this review already...

shimwell:
Just resolved a few simple conflicts

gonuke:
I'm a little confused by the last commit... Does @Shimwell have write permission on @pshriwise fork?

pshriwise:
> I'm a little confused by the last commit... Does @Shimwell have write permission on @pshriwise fork?

I don't think so? @Shimwell submitted a PR to update this branch handling some merge conflicts. I didn't see it in time and ended up doing my own rebase (Sorry, Jon).

shimwell:
Ah yes sorry just trying to help by clicking the solve conflicts button on this thread and resolving the conflicts (which I guess submits a PR against the fork branch).Sorry for the confusion.

gonuke:
Thanks @pshriwise

Issue #721, opened 2021-01-18
CMake Package Version File


pshriwise:
Install a `DAGMCConfigVersion.cmake` file according to CMake's [standard](https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/Packaging#package-version-files). This will allow the DAGMC version to be checked as part of CMake's `find_package` in external codes.

pshriwise:
Addressed in #722

PR #722, opened 2021-01-18
Adding CMake Version File

pshriwise:
## Description

Addressing #721 here. This is fairly simple -- just adds a version file to simplify any external version checking from other packages.


ljacobson64:
Just tested this with `find_package(DAGMC REQUIRED VERSION 3.3.0)` and cmake did not give an error

pshriwise:
According to the [`find_package` docs](https://cmake.org/cmake/help/v3.12/command/find_package.html), it would be 

`find_package(DAGMC 3.3.0 REQUIRED)`

to ensure a specific version is found. 

I tried the same as you expecting some kind of warning from CMake about the `VERSION`, but I didn't see anything. Maybe that generally has some other purpose in CMake functions?

ljacobson64:
Ah, so my syntax was wrong. I tested it with the right syntax and it behaves appropriately now.

ljacobson64 (approved):

pshriwise:
Are we happy with these changes as they are?

PR #728, opened 2021-02-24
add compatibility flag required to build PyNE against hdf5@1.12+

bam241:
deprecation in hdf5 1.12 change the interfaces.
this allows to use the old interface with `hdf5 1.12+` on which PyNE relies.
see pyne/pyne#1220 for more details

### DIFF
--- a/src/pyne/CMakeLists.txt
+++ b/src/pyne/CMakeLists.txt
@@ -12,4 +12,8 @@ if (CMAKE_Fortran_COMPILER_ID STREQUAL "Intel")
   set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O0 -g")
 endif ()
 
+if(NOT (${HDF5_VERSION} VERSION_LESS 1.12.0))
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DH5Oget_info_by_idx_vers=1 -DH5O_info_t_vers=1 -DH5Oget_info_by_name_vers=1")
+ endif()
+
 dagmc_install_library(pyne_dagmc)
--- a/news/PR-0728.rst
+++ b/news/PR-0728.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+- adding special build flag to enable old hdf5 interface for PyNE when compiling
+  against hdf5@1.12+
+
+**Security:** None


bam241:
@gonuke this should fix the hdf5@1.12 but again I am unsure this is the right way to do this.

gonuke:
Thanks @bam241 - this does seem like a bandaid, but we'll have to take it for now

gonuke:
This passed tests, but there was a failure on merge
https://app.circleci.com/pipelines/github/svalinn/DAGMC/271/workflows/52e99c23-b970-4a48-a8ec-ee60ee2d51b1/jobs/1060
Any ideas @bam241

bam241:
this is very strange, this is one of the 4/5 tests we actually run on PR....

I restarted the one that failed...

bam241:
the second run worked.

I don't like that I feel like we are experiencing those more and more...
I am not sure how to debug this, if this is a bug...

pshriwise:
It is! I've noticed this too, but haven't had time to look into it. Seeing this thread reminded me to have another look. I'm fairly certain that this has to do with a threadsafe vulnerability in MOAB. I'll make an issue in MOAB's repo and drop the link here for the explanation.

pshriwise:
Ok here's the relevant MOAB issue: https://bitbucket.org/fathomteam/moab/issues/135/thread-safety-issue-in-typesequencemanager

I'm not totally sure why we don't see this problem in our MC applications -- we're able to get the same result with 1 as with N threads. My only guess there is that there's enough time between the calls due to other operations. The calls in the `overlap_check` program are also extremely coherent in terms of the location and ray direction, making it more likely that they'll touch the same data in MOAB and expose these kinds of issues.

src/pyne/CMakeLists.txt
```diff
@@ -12,4 +12,8 @@ if (CMAKE_Fortran_COMPILER_ID STREQUAL "Intel")
   set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O0 -g")
 endif ()
 
+if(NOT (${HDF5_VERSION} VERSION_LESS 1.12.0))
```
ahnaf-tahmid-chowdhury, line 15: It seems conda version doesn't set HDF5_VERSION.

```console
Writing atomic_data.cpp
renamed 'pyne.cpp.new' -> 'pyne.cpp'
'/DAGMC/src/pyne/pyne/share/source.F90' -> '/DAGMC/src/pyne/../mcnp/mcnp5/pyne_source/source.F90'
'/DAGMC/src/pyne/pyne/share/source_mcnp6.F90' -> '/DAGMC/src/pyne/../mcnp/mcnp6/pyne_source/source.F90'

CMake Error at src/pyne/CMakeLists.txt:32 (if):
  if given arguments:

    "NOT" "(" "VERSION_LESS" "1.12.0" ")"

  Unknown arguments specified
```

PR #731, opened 2021-03-18
adding pyne as submodule and corresponding cmake change

bam241:
Change pyne from amalgamate only to submodule with amalgamation at compilation


src/pyne/CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
 message("")
 
-set(SRC_FILES pyne.cpp)
-set(PUB_HEADERS pyne.h)
+execute_process(COMMAND amalgamate_pyne.sh)
+
+file(GLOB SRC_FILES "*.cpp")
+file(GLOB PUB_HEADERS "*.hpp")
```
gonuke:
```suggestion
file(GLOB PUB_HEADERS "*.h")
```

gonuke: Why not keep the same `set` command - the file names are the same, right?

bam241: the goal was to have it the same way it is implemented else where in `DAGMC`, not sure it makes a lot of sense for pyne...

src/pyne/CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
 message("")
 
-set(SRC_FILES pyne.cpp)
-set(PUB_HEADERS pyne.h)
+execute_process(COMMAND amalgamate_pyne.sh)
```
gonuke: Based on the failing tests, it seems that this may still not be run correctly???

bam241: the problem is deeper than that.

basically the `execute_process` if run is parallel, so the file are not there when needed to configure the lib in CMAKE...

pshriwise: Hrmm yeah I think we'll want to use the `add_custom_command`/`add_custom_target` commands here so we can add this step as a dependency of the PyNE library build.

bam241: from what I read those are run at build time ...

gonuke: If we add a result code, can this force it to not be parallel?
https://stackoverflow.com/questions/52076948/check-exit-status-of-execute-process-in-cmake

bam241:
@pshriwise @gonuke this seems to be working.

what PyNE commit do we want to use for the submodule ?

pshriwise:
I just want to note that this adds `Python` as a build requirement for us. It's ubiquitous enough that I'm fine with that, but we may want to check that this works with both Python 2 and 3 just in case.

src/pyne/CMakeLists.txt
```diff
@@ -1,7 +1,16 @@
...
 
-set(SRC_FILES pyne.cpp)
-set(PUB_HEADERS pyne.h)
+
+MESSAGE("${CMAKE_CURRENT_SOURCE_DIR}/amalgamate_pyne.sh")
+execute_process(COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/amalgamate_pyne.sh ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR} RESULT_VARIABLE STATUS  OUTPUT_VARIABLE OUTPUT2)
+if(STATUS AND NOT STATUS EQUAL 0)
+  message(STATUS "PyNE AMALGAMATION FAILED: ${STATUS}")
+else()
+  message(STATUS "PyNE AMALGAMATION SUCCESSFULL ${OUTPUT2}")
```
pshriwise:
```suggestion
  message(STATUS "PyNE AMALGAMATION SUCCESSFUL ${OUTPUT2}")
```

src/pyne/CMakeLists.txt
```diff
@@ -1,7 +1,16 @@
 message("")
 
-set(SRC_FILES pyne.cpp)
-set(PUB_HEADERS pyne.h)
+
+MESSAGE("${CMAKE_CURRENT_SOURCE_DIR}/amalgamate_pyne.sh")
+execute_process(COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/amalgamate_pyne.sh ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR} RESULT_VARIABLE STATUS  OUTPUT_VARIABLE OUTPUT2)
```
pshriwise: Can we break up this long line a bit?

pshriwise (changes requested):
This looks good overall to me. Just a couple of small comments. 

One thing that would be nice is to automate the submodule checkout as part of our CMake command for convenience. It would make life easier for new users.

.circleci/config.yml
```diff
@@ -42,6 +42,11 @@ jobs:
     working_directory: /root/build_dir/DAGMC
     steps:
       - checkout
+      - run:
+          name: "Pull Submodules"
+          command: |
+            git submodule init
```
pshriwise: We should include these instructions in our installation docs as well.

gonuke:
Discussion point: this implementation places the amalgamated files in the `CMAKE_CURRENT_BINARY_DIR` rather than `CMAKE_SRC_DIR`.  On one hand, it would be cleaner to have them in `CMAKE_SRC_DIR` where other source files are.  On the other hand, as files that are auto-generated, there is an argument for them to be in `CMAKE_CURRENT_BINARY_DIR` (like `DagMCVersion.hpp`).

I have a week preference for `CMAKE_SRC_DIR` because it will probably result in less confusion going forward. These files could be added to the `.gitignore` file as well to prevent them being caught by git.

Other opinions?

pshriwise:
> I have a week preference for `CMAKE_SRC_DIR` because it will probably result in less confusion going forward. These files could be added to the `.gitignore` file as well to prevent them being caught by git.
> 
> Other opinions?

No strong opinions here either. It does seem a little more natural that the generated source files stay with the other source files. I'm on board if @bam241 is.

bam241:
> > I have a week preference for `CMAKE_SRC_DIR` because it will probably result in less confusion going forward. These files could be added to the `.gitignore` file as well to prevent them being caught by git.
> > Other opinions?
> 
> No strong opinions here either. It does seem a little more natural that the generated source files stay with the other source files. I'm on board if @bam241 is.

I have the same feeling.

The only thing that bother me is the tacky way in the other lib (like mcnp, geant4 and fluka)

How about creating a variable to share with the other packages ? so it would look less tacky ?

Again I don't have any strong feeling, and tweaking the implementation at this point is a matter of minutes...

bam241:
I deleted my previous comment.

I like @gonuke suggestion, @pshriwise seems to be onboard as well

so I feel like we are kinda all liking the idea... let go that way !

bam241:
@pshriwise @gonuke this should be good to go, (adding Pyne.h/.ccp in `CMAKE_SRC_DIR` with updated gitignore)

we need to choose which commit from pYNE to use (right now last `pyne/develop`)

gonuke (approved):
LGTM - thanks @bam241

pshriwise:
> we need to choose which commit from pYNE to use (right now last `pyne/develop`)

My first thought was "yeah, why not have all the latest updates in PyNE?". I also had the thought that we should update to a specific release of PyNE for a release of DAGMC, but this would tie DAGMC's release timeline to PyNE's. 

I'll throw out three paths forward:

  1. Use `pyne/develop` and update to a specific version of PyNE upon DAGMC release, being ok with having to wait for the next PyNE release to release DAGMC.
  2. Use `pyne/develop` always, being ok with releasing DAGMC with a development version of amalgamated PyNE.
  3. Always use a specific version of PyNE in DAGMC, only updating the PyNE submodule from one release to another.
  
  After writing all three options out I like the third the best. We have influence over PyNE's release timeline, so if we do want to make updates to the version we can make that happen, but it won't impede DAGMC's release timeline as option 1 would. It also avoids the potential situation where we forget to update PyNE to a specific version in the rush to do a DAGMC release.
  
  In the end, any of these three are workable to me. I just think that option 3 is the most sustainable. Thoughts @bam241 @gonuke?

gonuke:
I agree with 3, including the possibility of micro releases of DAGMC that are just to update the PyNE hash

bam241:
I agree with both you statement, I just updated the `PyNE` submodule to tag `0.7.3`

pshriwise (approved):
Thanks for the good discussion and for taking care of this @bam241! Will merge by EOD.

PR #734, opened 2021-03-24
Checkout PyNE submodule as part of CMake

pshriwise:
## Description
Adds an option (default ON) to checkout the PyNE submodule as part of the CMake configuration step.

## Motivation and Context
The addition of this submodule in DAGMC makes updates of PyNE in DAGMC more elegant, but it also requires an extra manual step in DAGMC installation (`git submodule update --init`). There are many build scripts/CI scripts out there that don't include this step and will break as a result.

## Changes
This PR incorporates the submodule initialization and update into the CMake step, meaning that no external build scripts will need to change due to the use of PyNE as a submodule. 

## Behavior
As mentioned above, a section has been added to `CMakeLists.txt` that will checkout the PyNE submodule automatically. If `git` is found on the system, the option `GIT_SUBMODULE` is added with a default value of `ON`. This option can be turned off for development with new versions of PyNE in DAGMC (another benefit of using the submodule as opposed to the previous approach). If `git` is not available, then this option will not appear in the CMake configuration and we will expect that the PyNE submodule is already present in the DAGMC source.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ jobs:
         type: string
     docker:
       - image: svalinn/dagmc-ci-ubuntu-18.04-housekeeping:latest
-        auth: 
+        auth:
           username: $DOCKERHUB_USER
           password: $DOCKERHUB_PASS
     working_directory: /root/build_dir/DAGMC
@@ -36,16 +36,12 @@ jobs:
         default: "OFF"
     docker:
       - image: svalinn/dagmc-ci-ubuntu-<< parameters.img >>-<< parameters.compiler >>-ext-hdf5_<< parameters.hdf5 >>-moab_<< parameters.moab >>:latest
-        auth: 
+        auth:
           username: $DOCKERHUB_USER
           password: $DOCKERHUB_PASS
     working_directory: /root/build_dir/DAGMC
     steps:
       - checkout
-      - run:
-          name: "Pull Submodules"
-          command: |
-            git submodule update --init 
       - run:
           name: Setup environment
           command: |
@@ -66,7 +62,7 @@ workflows:
   version: 2
   pull_request: # Run only for PullRequest
     jobs:
-      
+
       - house_keeping:
           context: dockerhub
           matrix:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@ set(DAGMC_VERSION ${DAGMC_MAJOR_VERSION}.${DAGMC_MINOR_VERSION}.${DAGMC_PATCH_VE
 
 # Set git SHA1 hash as a compile definition
 find_package(Git)
-if(GIT_FOUND)
+if(GIT_FOUND AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.git")
   execute_process(COMMAND ${GIT_EXECUTABLE} rev-parse HEAD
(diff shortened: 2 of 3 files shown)


gonuke:
Did the command-line change for newer version of git?  The 18.04 image is using git 2.17.1 and the 16.04 image is using git 2.7.4.  The former fails on the submodule and the latter is successful.

pshriwise:
> Did the command-line change for newer version of git? The 18.04 image is using git 2.17.1 and the 16.04 image is using git 2.7.4. The former fails on the submodule and the latter is successful.

Huh, no I don't think it has. I'll dig into it right now.

bam241:
I think you need a `--recursive`:

https://cliutils.gitlab.io/modern-cmake/chapters/projects/submodule.html

pshriwise:
> I think you need a `--recursive`:
> 
> https://cliutils.gitlab.io/modern-cmake/chapters/projects/submodule.html

`--recursive` is already part of the command in `CMakeLists.txt`. (Though it doesn't appear in the error message. I'll update that.) At any rate, this shouldn't have an effect though as there are no submodules in PyNE to checkout.

Thanks for putting that reference in here btw. That's where this pattern came from.

bam241:
I believe, the git error is: `error: pathspec '36' did not match any file(s) known to git.`
Do you have any idea how git try to get `36` from our repo or PyNE ?

gonuke:
I saw ‘4’ as the failing pathspec before on the first failure!?!

gonuke:
The strange thing is that it works fine on 16.04

pshriwise:
> I saw ‘4’ as the failing pathspec before on the first failure!?!

Right! It's pretty strange. It seems like something extra is getting added to the command.

pshriwise:
Installed a newer version of CMake in CI to get the exact command being run. Not super helpful...

```
'bash' '-c' '/usr/bin/git submodule update --init --recursive'
error: pathspec '36' did not match any file(s) known to git.
```

pshriwise:
Wow. I can hardly believe what the problem was. Apparently having the `jobs` environment variable messes with `git submodule` commands (maybe others?) on Ubuntu >= 18.04. I tried the following on my DAGMC repo locally (20.04):

`jobs=4 git submodule update --init --recursive`

And it fails with the same error. The `pathspec` showing up in CI was the value of the jobs variable. I tried searching to see if this is intended behavior for `git`, but I couldn't find anything. I wonder if it's worth reporting...

The fix I've added in 3f0e4c7 is just temporary. I think a better long-term solution is to update the `jobs` variable we set to `ci_jobs`. This means updating the docker files I think. I can get that process started tomorrow if you're on board with that plan, @gonuke @bam241.

pshriwise:
Though now that I say that... if we find that updating the docker images will take some time I wouldn't mind getting this solution in sooner as it's holding up some downstream progress elsewhere.

gonuke:
Wild!!!

check out this file (search for "jobs") from the official git repo:

https://github.com/git/git/blob/3a0b884caba2752da0af626fb2de7d597c844e8b/git-submodule.sh

gonuke:
I think we should also remove the `git submodule` command from `.circleci/config.yml`

gonuke:
I'm working on rebuilding docker images now, by the way.

gonuke:
Thanks @pshriwise

PR #735, opened 2021-03-25
Change name of var for make file parallelization

gonuke:
## Description
Change the name of the variable used in CI to manage `make` parallelization from `jobs` to `ci_jobs`.

## Motivation and Context
The variable `jobs` is used in undocumented/poorly documented ways by the `git submodule` command.  We have used a variable with that name in our CI scripts to pass the number of processors to `make -j` for build parallelization.  Changing the name of this variable should allow `git submodule` to work as expected. (as introduced in #734)

Reference: https://github.com/git/git/blame/3a0b884caba2752da0af626fb2de7d597c844e8b/git-submodule.sh
In the last 3-5 years, `git submodule` started using this env variable to expose parallelism to the user.  E.g. Line 547 was added to this file about 3 years ago. Line 50 was added in the above about 12 months ago, indicating a similar problem identified within the git project.  Presumably, newer versions of `git` would not see this problem, just as older versions of `git` did not.

## Changes
This is best described as a bug fix.

## Behavior
Prior to this change, calls to `git submodule` would fail with an unknown pathspec with the same value as the `jobs` variable.

After this change, calls to `git submodule` should work correctly.

### DIFF
--- a/CI/env.sh
+++ b/CI/env.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-export jobs=`grep -c processor /proc/cpuinfo`
+export ci_jobs=`grep -c processor /proc/cpuinfo`
 
 export CTEST_OUTPUT_ON_FAILURE=1
 
@@ -20,7 +20,7 @@ export  FC=`which gfortran`
 if [ "$COMPILER" == "gcc" ]; then
   export  CC=`which gcc`
   export CXX=`which g++`
-  export jobs="4"
+  export ci_jobs="4"
 elif [ "$COMPILER" == "clang" ]; then
   export  CC=`which clang`
   export CXX=`which clang++`
--- a/news/PR-0735.rst
+++ b/news/PR-0735.rst
@@ -2,11 +2,14 @@
 
 **Changed:** None
 
+- updated location & checksum of GEANT4 tarball
+ 
 **Deprecated:** None
 
 **Removed:** None
 
 **Fixed:** 
+
 - Renamed `jobs` variable CI build system to avoid undocumented conflict with `git submodule`
 
 **Security:** None
--- a/news/PR-0735.rst
+++ b/news/PR-0735.rst
@@ -1,0 +1,12 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** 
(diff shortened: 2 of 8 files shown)


news/PR-0735.rst
```diff
@@ -0,0 +1,12 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** 
+- Renamed `jobs` variable CI build system to avoid undocumented conflict with `git submodule`
```
pshriwise, line 13: For rendering


```suggestion

- Renamed `jobs` variable CI build system to avoid undocumented conflict with `git submodule`
```

gonuke:
Update: I was able to successfully build the images for U16.04, but had to give up while building U18.04 b/c I needed my computer for other things - will resume later today.

gonuke:
I've successfully built all the docker images once, but rebasing has resulted in needing to rebuild them all.  This will take O(16h) which I'll start tonight and hopefully update the images tomorrow morning to demonstrate passing tests on 18.04.

CI/env.sh
```diff
@@ -20,7 +20,7 @@ export  FC=`which gfortran`
 if [ "$COMPILER" == "gcc" ]; then
   export  CC=`which gcc`
   export CXX=`which g++`
-  export jobs="4"
+  export ci_jobs="4"
```
ljacobson64, line 23: Doesn't this line overwrite the previous declaration of `ci_jobs`?

ljacobson64:
Funnily enough, I noticed this exact same problem a while back in my personal build scripts. I ended up changing all instances of `jobs` to `num_cpus`.

CI/env.sh
gonuke: It does, but it's on purpose.  On our CI machines, gcc manages to consume all the memory and the job crashes/fails.  We need to limit to 4 CPUs

ljacobson64: That makes sense

gonuke:
These passed tests with the new CI images and without the fix that @pshriwise put in - so I think that means everything is right in the world, once again

ljacobson64 (approved):

pshriwise (approved):
This looks ready to me. Thank you for addressing this so quickly, @gonuke!

PR #738, opened 2021-04-19
Enable double-down in CI for pull requests.

pshriwise:
Seeing as this option is being increasingly used outside of DAGMC, I'd like to add it to CI run with pull requests to make sure things aren't breaking as development continues.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -113,7 +113,7 @@ workflows:
               only: develop
       - build_and_test:
           context: dockerhub
-          name: "Double Down"
+          name: "Double-Down"
           img: "18.04"
           compiler: "gcc"
           hdf5: "1.10.4"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,6 +86,18 @@ workflows:
             branches:
               ignore: develop
 
+      - build_and_test:
+          context: dockerhub
+          name: "Double Down"
+          img: "18.04"
+          compiler: "gcc"
+          hdf5: "1.10.4"
+          moab: "master"
+          double_down: "ON"
+          filters:
+            branches:
+              ignore: develop
+
   merge:  # Run only on develop change -> when merging a PullRequest
     jobs:
       - build_and_test:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,7 +88,7 @@ workflows:
 
       - build_and_test:
           context: dockerhub
-          name: "Double Down"
+          name: "Double-Down"
           img: "18.04"
           compiler: "gcc"
           hdf5: "1.10.4"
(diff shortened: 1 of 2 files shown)


pshriwise:
I'm seeing failures in CI, but they don't seem to be related to the changes here. Am I interpreting that right @bam241?

bam241:
I am 90% sure that comes from `make -j${ci_procs}` for some reason gcc is not stable when using to many proc to compile....

pshriwise:
Ah I see. Should I try reducing the number of `make` jobs here or is that being addressed in another PR?

gonuke:
This is clashing a little with the fixes I have in #687, but I thought the fix for GCC already existed in the `env.sh` that should help this PR as well

gonuke:
P.S. #687 is basically waiting on the race condition fix for MOAB because that's causing the tests to fail

pshriwise:
Happy to let #687 go in first. I can handle any conflicts that come up. Added some labels here to make sure we wait a bit on this. 

> P.S. #687 is basically waiting on the race condition fix for MOAB because that's causing the tests to fail

That PR should go in soon -- the ball is in the MOAB dev team's court to add a change to my PR before we merge it. It won't fix our broken builds for MAOB 5.1.0 though. Do we have a plan for how to deal with that going forward? Update to the most recent release of MOAB with that fix and rely on it for CI? Should we also disable the `overlap_check` program or limit it to a single thread depending on the MOAB version?

bam241:
> This is clashing a little with the fixes I have in #687, but I thought the fix for GCC already existed in the `env.sh` that should help this PR as well

we changed the variable name to `${ci_proc}`  in order to avoid conflict with the CMake var `${proc}`. I am unsure if this solved anything with regard to gcc errors....
I usually experience this kind of gcc error with I compile with a very high `-jX` or with only the `-j`

bam241:
an explanation of the problem might be there: https://stackoverflow.com/questions/30887143/make-j-8-g-internal-compiler-error-killed-program-cc1plus

gonuke:
But we also have a check that sets ${ci_proc} to be only 4 for bcc

gonuke:
Should this be matrixed in to the other builds?

pshriwise:
> Should this be matrixed in to the other builds?

If it won't affect the CI runtime too much, I'd be happy to add it to the job matrix. What do you think @bam241?

bam241:
sounds like a good plan

gonuke:
On second thoughts... maybe leave it out for now.  We probably need to rethink CI as the number of dimensions of the matrix expand.  We probably don't need to fill the matrix anyway, and we might want to move from 16.04 to 18.04.

gonuke (approved):
Let's merge this and worry about the CI matrix at another time.

PR #742, opened 2021-06-16
Regrouping all versions and SHA in a single file, simplifying version for MOAB

bam241:
This simplifies how we deal with version for MOAB and HDF5 (in CI)
and regrouped all versions and SHA storing into a single file. this should simplify changing and updating versions....


bam241:
wondering if `versions.sh` should also include links to repos and version to download.... (and be renamed....)

CI/update_docker.sh
```diff
@@ -30,22 +29,18 @@ for ubuntu_version in ${ubuntu_versions}; do
                                   --build-arg COMPILER=${compiler} \
                  -f CI/Dockerfile_1_external_deps .
     if [ "$PUSH" = true ]; then docker push ${image_name}; fi
-    for hdf5 in $hdf5_versions; do
```
gonuke, line 33: Don't we still want/need HDF5 variations?

bam241: we use only one everywhere.... so we might as well remove them for simplicity...
if we need more we can always add that later....

gonuke:
I thought we tested the default version as well as the 1.10.4 version

CI/update_docker.sh
gonuke: I thought we tested the default version as well as the 1.10.4 version

CI/update_docker.sh
```diff
@@ -17,8 +17,7 @@ done
 
 ubuntu_versions="16.04 18.04"
 compilers="gcc clang"
-hdf5_versions="1.10.4"
```
bam241, line 20: see here @gonuke
 we only builded for 1.10.4
and in CI we only test for 1.10.4 as well (we had the capability to test more tho)

CI/update_docker.sh
gonuke: In /CI/Dockerfile we invoke this with HDF5=system

bam241: it is an available option, but I just checked we don't use it anymore in CI

gonuke: Can we keep the infrastructure for doing this, even if we aren't using it?  It will be easier to have loops of length 1 than to recreate this if we want it again.

gonuke:
I think this is deprecated by our move to Github Actions, but some of the concepts may still apply???

gonuke:
This is now way out of date... is it worth updating it? starting over? abandoning?

gonuke:
I think this PR has been superseded by other recent changes in the build system and docs.  Thanks @bam241

gonuke:
> I think this PR has been superseded by other recent changes in the build system and docs. Thanks @bam241

Please close this PR if you agree that it's no longer relevant, @bam241.  Thanks!

PR #743, opened 2021-06-19
Fix for overlap check test

pshriwise:
## Description
Threadsafe fix for the overlap check test. The `overlap_map` shared variable was being modified in an unprotected block and could potentially be modified by multiple threads at once.

### DIFF
--- a/news/PR-0743.rst
+++ b/news/PR-0743.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:**
+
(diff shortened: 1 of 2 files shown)


pshriwise:
Did my best to rule out any further race conditions by testing 500 times w/ 10, 20, & 30 threads.

gonuke (approved):
Thanks @pshriwise

bam241:
@pshriwise thx for the fix !

PR #746, opened 2021-07-02
added gh action to build dockerfiles on ghcr

shimwell:
## Description
This adds a GitHub action that builds the multistage dockerfiles in the DAGMC CI folder.
They are also uploaded to GitHub packages / container registry by the same GitHub action
I have proptyped this on my own branch and there we can see:
- the actions are running successfully https://github.com/Shimwell/DAGMC/actions
- the docker images are being uploaded to the ghcr https://github.com/Shimwell?tab=packages

## Motivation and Context
The automatic build of the docker images with this new method is less than 2 hours which I guess is quicker than running the update_docker.sh script locally. This is due to the use of matrix strategy in GitHub actions that runs jobs in parallel

## Changes
Feature / automation

## Behavior
This automates the building and uploaded of the dockerfiles which I believe is currently performed locally using the update_docker.sh script. 

## News file
All pull requests are required to include an updated news file with the PR.
The file should be located at the `/news/PR-<#>.rst` where `<#>` should be replaced with the appropriate PR number.
Please see [here](https://github.com/svalinn/DAGMC/blob/develop/news/TEMPLATE.rst) for the news file template.


.github/workflows/docker_publish.yml
```diff
@@ -0,0 +1,101 @@
+on:
+  push:
+    branches: ['gh_actions_and_ghcr_for_docker']  # todo replace with main once we know this is working
```
gonuke: Can we add the filter so that this only runs when certain files change?

gonuke (changes requested):
This looks great @Shimwell - thanks for dragging us into a new era!

Just one comment about when this runs....

shimwell:
Just to add it looks like the docker images built and are available here https://github.com/orgs/svalinn/packages

.github/workflows/docker_publish.yml
shimwell: I am not sure about triggering on changes to certain files, I shall look into that

I can however get it to trigger on a GitHub Release if that helps

gonuke: I found this somewhere and need to find the link again

gonuke: https://github.community/t/is-it-possible-to-run-the-job-only-when-a-specific-file-changes/115484

shimwell: That is a nice GitHub action you found there. 

I can give this a go.

Which files would we want to monitor for changes. I guess just the source code folder and the Docker files themselves

gonuke: For this action, it's ONLY the Docker files - not the source code.  There is no reason to rebuild the images for CI (which don't include DAGMC itself) when the source code changes.   I suppose we can add a different action that adds DAGMC to one of these images on release.

shimwell: In that case it might be worth moving the Dockerfiles to a new repo as it appears they are somewhat disconnected from the DAGMC source code itself?

Perhaps the exisiting ```dockerfiles``` repo could be renamed ```dockerfiles-pymoab``` and there could be a new one called ```dockerfiles-dagmc-dependencies```

gonuke: But they are specifically the images we rely on for CI, so that's why it makes sense for them to stay here.  We do have the other dockerfiles repo, and that could be a place to develop more generic dockerfiles-dagmc-dependencies images.  For example, the specific hash of MOAB that we have in here right now is to advance CI and not because we would want to publish a dependency based on it.

shimwell: I've added the conditional file change trigger to the github action, thanks for finding that.

I've tried to checked it works by pushing with no changes and again pushing with superficial changes to the Dockerfiles

As you can see in the Github Actions appeared to skip both times, so I've got something a bit wrong but I think we are close

.github/workflows/docker_publish.yml
```diff
@@ -0,0 +1,116 @@
...
+        uses: docker/setup-qemu-action@v1
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Verify Changed files
```
gonuke: I think this action is more complex than necessary.... and I read that it's for detecting changes during the action/workflow

It should be possible to add some version of `paths:` to the top level job control in the first few lines.

gonuke (changes requested):
I think this is always skipping because you're using an action that detects changes that occur *during* the action

shimwell:
Thanks for working through this Paul, I've been experimenting on my own fork with this new simple trigger (push on path) https://github.com/Shimwell/DAGMC/pull/1

It does appear much simpler than the previous mess I was trying, thanks for pointing me in the right direction

gonuke:
I've been thinking about whether we need more testing, but I think I'll dive in and merge it since we aren't using these for CI yet anyway

gonuke (approved):
Thanks @Shimwell  - let's give it a try

PR #748, opened 2021-07-08
moved to sequential docker building

shimwell:
## Description
This is a minor change to the docker CI image building process. It uses "needs" keyword in the steps for docker images building to ensure the images are getting the most up to date version of the previous image.

This has a nice upgrade path as I think it is compatible with docker layer caching which would speed up the overall process.

Also if we like a badge can be added to the README ?
[.github/workflows/docker_publish.yml](https://github.com/svalinn/DAGMC/actions/workflows/docker_publish.yml)

## Motivation and Context
Ensures images build on the latest previous base images
Cleaner presentation of build process with [connections between images identified on the GH actions](https://github.com/svalinn/DAGMC/actions/runs/1011041967)

## Changes
What kind of change does this PR introduce? (Bug fix, feature, documentation update...)
update to current GH action building the dockerfiles

## Behavior
What is the current behavior? What is the new behavior?
Current behavior is one large matrix of builds, new process is sequential matrixes of builds with "needs" keyword 

## News file 
done

PR #752, opened 2021-07-09
Switching from CI to GitHub

bam241:
moving away from CircleCI


bam241:
@gonuke @pshriwise @Shimwell I think this is ready for a first round of review.

this should port our CircleCI capability to GithubAction

.github/workflows/build_test.yml
```diff
@@ -0,0 +1,112 @@
...
+  push:
+    branches:
+      - develop
+      - master
+
+jobs:
+  Housekeeping:
+    runs-on: ubuntu-latest
+    container:
+      image: svalinn/dagmc-ci-ubuntu-18.04-housekeeping:latest
```
shimwell: Should this point to the ghcr image instead of the dockerhub one?
 
```suggestion
      image: ghcr.io/svalinn/dagmc-ci-ubuntu-18.04-housekeeping:latest
```

.github/workflows/build_test.yml
```diff
@@ -0,0 +1,112 @@
...
+          echo "HDF5_VERSION=${{ matrix.hdf5_versions }}" >> $GITHUB_ENV
+          echo "REPO_SLUG=${GITHUB_REPOSITORY}" >> $GITHUB_ENV
+          echo "PULL_REQUEST=$(echo  $GITHUB_REF | cut -d"/" -f3)" >> $GITHUB_ENV
+          echo "DOUBLE_DOWN="OFF"" >> $GITHUB_ENV
+          ln -s $GITHUB_WORKSPACE /root/build_dir/DAGMC
+         
+      - name: Building DAGMC
+        run: |
+          cd /root/build_dir/DAGMC
+          CI/circleci/install.sh
```
shimwell: Perhaps we should remove the ```circleci``` folder or rename it

.github/workflows/build_test.yml
gonuke: One step at a time?? or take the plunge??  I guess we have CI to test it, so I agree we go for it

.github/workflows/build_test.yml
```diff
@@ -0,0 +1,112 @@
...
+      - develop
+      - master
+
+jobs:
+  Housekeeping:
+    runs-on: ubuntu-latest
+    container:
+      image: svalinn/dagmc-ci-ubuntu-18.04-housekeeping:latest
+    steps:
+      - name: Install latest git
```
gonuke: ?? we have to install git ??

Or is it just the newer version we need - maybe @Shimwell 's ghcr.io images have it already?

.github/workflows/build_test.yml
```diff
@@ -0,0 +1,112 @@
...
+        isPR:
+          - ${{ contains(github.event_name, 'pull_request') }}
+        exclude:
+          - isPR: true
+            moab_versions: develop
+          - isPR: true
+            moab_versions: master           
+
+    container:
+      image: svalinn/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler }}-ext-hdf5_${{ matrix.hdf5_versions }}-moab_${{ matrix.moab_versions }}:latest
```
gonuke: (because I don't know) Do we need the `runs-on` entry above if we are specifying these images?

.github/workflows/build_test.yml
```diff
@@ -0,0 +1,112 @@
...
+     
+      - name: Housekeeping
+        run: |
+          cd $GITHUB_WORKSPACE
+          CI/circleci/housekeeping.sh
+
+
+  BuildTest:
+    needs: Housekeeping
+    runs-on: ubuntu-latest
```
gonuke, line 39: (b/c I don't know:) do we need this `runs-on` if we specify the images below?

.github/workflows/build_test.yml
```diff
@@ -0,0 +1,112 @@
...
+          echo "COMPILER=${{ matrix.compiler }}" >> $GITHUB_ENV
+          echo "HDF5_VERSION=${{ matrix.hdf5_versions }}" >> $GITHUB_ENV
+          echo "REPO_SLUG=${GITHUB_REPOSITORY}" >> $GITHUB_ENV
+          echo "PULL_REQUEST=$(echo  $GITHUB_REF | cut -d"/" -f3)" >> $GITHUB_ENV
+          echo "DOUBLE_DOWN="OFF"" >> $GITHUB_ENV
+          ln -s $GITHUB_WORKSPACE /root/build_dir/DAGMC
+         
+      - name: Building DAGMC
+        run: |
+          cd /root/build_dir/DAGMC
```
gonuke: consistency: above, you used `$GITHUB_WORKSPACE` and here you used `/root/build_dir/DAGMC`

.github/workflows/build_test.yml
```diff
@@ -0,0 +1,112 @@
...
+
+      - name: Setup
+        run: |
+          echo "MOAB_VERSION=${{ matrix.moab_versions }}" >> $GITHUB_ENV
+          echo "COMPILER=${{ matrix.compiler }}" >> $GITHUB_ENV
+          echo "HDF5_VERSION=${{ matrix.hdf5_versions }}" >> $GITHUB_ENV
+          echo "REPO_SLUG=${GITHUB_REPOSITORY}" >> $GITHUB_ENV
+          echo "PULL_REQUEST=$(echo  $GITHUB_REF | cut -d"/" -f3)" >> $GITHUB_ENV
+          echo "DOUBLE_DOWN="OFF"" >> $GITHUB_ENV
+          ln -s $GITHUB_WORKSPACE /root/build_dir/DAGMC
```
gonuke, line 88: b/c I don't know: does this symlink need to be repeated or is it part of the image inherited from housekeeping?

gonuke (changes requested):
Thanks @bam241!

A few comments, mostly as I try to learn the syntax and develop a mental model

.github/workflows/build_test.yml
shimwell: Oh I think I made bad choice here https://github.com/svalinn/DAGMC/pull/748/commits/906afc41ba0d6270f636225d9a27b68e6764698c. 

I was building both housekeeping images (16.04 and 18.04) then I reduced it down to just one and I chose 16.04 for some unknown reason. So the 18.04 image is there but it is just not kept up to date like the 16.04 house keeping image

.github/workflows/build_test.yml
shimwell: The dockerhub and ghcr docker images appears to have the same git version.

```bash
sudo docker run -it svalinn/dagmc-ci-ubuntu-18.04-housekeeping:latest
git --version
 >> version 2.17.1
sudo docker run -it ghcr.io/svalinn/dagmc-ci-ubuntu-18.04-housekeeping:latest
git --version
 >> version 2.17.1
```

bam241: They do. 
But we need 2.18 minimum so GitHub action use it for the checkout.
Otherwise it falls back on the "REST" api and just download a folder that does not contain git versioning info 

We need those for submodule and housekeeping

shimwell: Perhaps the ```actions/checkout@v2``` can also do submodules checkout but they are just off by default. I think it can be switched on (in the usage section of docs -> https://github.com/actions/checkout)

.github/workflows/build_test.yml
```diff
@@ -0,0 +1,112 @@
...
+    steps:
+      - name: Install latest git
+        run: |
+          apt-get install -y software-properties-common
+          add-apt-repository ppa:git-core/ppa
+          apt update
+          apt install -y git
+
+      - name: Checkout repository
+        uses: actions/checkout@v2
```
shimwell: This might be worth a try as if it works the parts were updated git installs are used could be removed

```suggestion
      - name: Checkout repository
        uses: actions/checkout@v2
        with:
          submodules: 'true'
```

bam241: already tried this does not work as git version is < 2.18

.github/workflows/build_test.yml
bam241: yes that the host on which we run the docker image

.github/workflows/build_test.yml
bam241: no we can't
`runs-on` need to be linux, Mac or windows.

... and only linux allows to get access to docker as sub-systems (I think)

.github/workflows/build_test.yml
bam241: it is not. the root problem here is the checkout from GithubAction do not use at the same location than in CircleCI.

I added that to have something working we might want to tweak our scripts a little so clean that up....

.github/workflows/build_test.yml
bam241: see here: https://github.com/svalinn/DAGMC/runs/3031336027?check_suite_focus=true

```
Run actions/checkout@v2
2  with:
3    submodules: recursive
4    repository: svalinn/DAGMC
5    token: ***
6    ssh-strict: true
7    persist-credentials: true
8    clean: true
9    fetch-depth: 1
10    lfs: false
11  env:
12    hdf5_versions: 1.10.4
13    hdf5_build_dir: hdf5_build_dir
14 /usr/bin/docker exec  47d233d87e73e379239b48b5ef537b79ae073afeb55c7f2006b218366d7d4468 sh -c "cat /etc/*release | grep ^ID"
15 Syncing repository: svalinn/DAGMC
16 Getting Git version info
20 Deleting the contents of '/__w/DAGMC/DAGMC'
21 The repository will be downloaded using the GitHub REST API
22 To create a local Git repository instead, add Git 2.18 or higher to the PATH
23 Error: Input 'submodules' not supported when falling back to download using the GitHub REST API. To create a local Git repository instead, add Git 2.18 or higher to the PATH.
```

.github/workflows/build_test.yml
bam241: renamed it `scripts` as it live in the CI folder already

.github/workflows/build_test.yml
gonuke: Can we add the git update to the Dockerfiles, then?

bam241: yes I believe that's the way to go.

I added it there to make sure we all agreed on this solution :)

gonuke: See #753

gonuke:
Do we know why these are failing?

shimwell:
> Do we know why these are failing?

Looks like we are trying to pull an image like this
docker pull ghcr.io/svalinn/dagmc-ci-ubuntu-18.04-housekeeping:stable

I think it might be named slightly differently 
docker pull ghcr.io/svalinn/dagmc-ci-ubuntu-18.04-housekeeping:latest

.github/workflows/build_test.yml
```diff
@@ -0,0 +1,98 @@
...
+  push:
+    branches:
+      - develop
+      - master
+
+jobs:
+  Housekeeping:
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/svalinn/dagmc-ci-ubuntu-18.04-housekeeping:stable
```
shimwell, line 19:
```suggestion
      image: ghcr.io/svalinn/dagmc-ci-ubuntu-18.04-housekeeping:latest
```

bam241: We want stable. They are building right now.

bam241:
The way I shaped this is the following:

docker building ci will use the latest tag as temporary tag to try compiling and the tests for dagmc. (This will run on the account of the user issuing the PR)

On merge, this will build with latest tag, run the test against the newly latest image.
If test are successful the image will be tagged as stable. And we will use those for ci.

bam241:
@gonuke @Shimwell I just realized that `latest` might be a confusion tag for temporary image shall I use something else ?

bam241:
> @gonuke @Shimwell I just realized that `latest` might be a confusion tag for temporary image shall I use something else ?

I'll do that in a dedicated PR. (opened an issue #762 )
If you are fine with the current state of this, we can merge this and retire our CircleCI hooks....

.github/workflows/build_test.yml
shimwell: Just that latest is also the default. So if you do a docker pull without anything at the end it will grab the latest one as latest is also default.

gonuke:
Thanks for the work on this transition @bam241

gonuke:
This failed on merge - we need the same fix as #757

gonuke:
The logic will get complicated if we take the same approach a s #757 because in some events we want `BuildTest` to depend on  `Housekeeping` and in other events we do not.

Modularity is the answer - make our own `BuildTest` action and call it in  different workflows. (do we want to go down this road right now???)

shimwell:
> @gonuke @Shimwell I just realized that `latest` might be a confusion tag for temporary image shall I use something else ?

Yes please.

PR #753, opened 2021-07-12
Add newer (v2.18) git to CI images

gonuke:
## Description
Install newer git during creation of CI images.

## Motivation and Context
Continuous integration relies on having a full peer of the git repo.  This was standard during CI on CircleCI, but the Github Action to fetch the repo requires a newer version of git (v2.18+) to accomplish this.

## Changes
This changes the packages installed when making the base CI images.

### DIFF
--- a/CI/Dockerfile_0_base
+++ b/CI/Dockerfile_0_base
@@ -25,9 +25,9 @@ RUN apt-get -y update; \
                        libpython3-dev \
                        wget \
                        software-properties-common;
-    add-apt-repository ppa:git-core/ppa
-    apt-get update
-    apt-get install -y git
+    add-apt-repository ppa:git-core/ppa; \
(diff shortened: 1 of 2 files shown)


gonuke:
I think these still need to be tested manually, is that right?

gonuke:
These are being tested on my repo... not sure if this is intended, but maybe a good outcome???

gonuke:
It seems that the test fail on my repo because I don't have push permission to the GHCR (?)

bam241:
This looks good .

I also think that pushing is not working. (we should probably add branch push restriction to avoid random run like this...)

I'll issue a PR to have the build run, and maybe tested on PR


if that crashes after merge, I'll take full responsibility, and will work on a fix !

Thx @gonuke, Merging

PR #754, opened 2021-07-12
integration of test for dockerfile update

bam241:
duplicating test for docker file update.


bam241:
also added a contiditon push based on PR status

bam241:
@Shimwell do you know if my implementation will use cashed img ?

shimwell:
> @Shimwell do you know if my implementation will use cashed img ?

I can see 12 jobs all running in parallel as the first stage. I don't think these will make use of cached images.

Perhaps the ```needs``` argument can be used in a few places to make sure the base images have been built first

I think the ```test_docker``` section will make use of previously build cached images that were build in the first section, but it should all appear in the logs

.github/workflows/docker_publish.yml
```diff
@@ -50,54 +60,113 @@ jobs:
         with:
           file: CI/Dockerfile_0_base
           context: .
-          push: true
+          push: ${{ matrix.isMerged }}
           build-args: |
             UBUNTU_VERSION=${{ matrix.ubuntu_versions }}
-          tags: ghcr.io/svalinn/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}
+          tags: ghcr.io/svalinn/test_dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}
```
gonuke: This will need to include some other unique string (sha?) to avoid two simultaneous PR's from overwriting each other

gonuke (changes requested):

.github/workflows/docker_publish.yml
```diff
@@ -44,6 +44,14 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+      
+      - name: Environment setup
+        run: |
+          if [ "${{ matrix.isMerged }}" == "true" ]; then
```
gonuke: Maybe I don't understand how the `isMerged` works, but this seems like the logic is reversed.  Don't we want the sha for PRs and without a sha for merged?

CI/Dockerfile_1_external_deps
```diff
@@ -1,5 +1,6 @@
-ARG UBUNTU_VERSION=18.04
-FROM ghcr.io/svalinn/dagmc-ci-ubuntu-${UBUNTU_VERSION}
+ARG UBUNTU_VERSION=18.04ARG
+ROOT_NAME=dagmc
```
gonuke:
```suggestion
ARG ROOT_NAME=dagmc
```

gonuke (changes requested):
Oops

bam241:
closing and reopening from upstream

bam241:
this is actually looking at my run on push, not the one on PR... very strange !

bam241:
here is the place where the test gets run on my user name: https://github.com/bam241/DAGMC/actions/workflows/docker_publish.yml

CI/Dockerfile_3_moab
```diff
@@ -1,7 +1,8 @@
 ARG UBUNTU_VERSION=18.04
 ARG COMPILER=gcc
 ARG HDF5=1.10.4
-FROM ghcr.io/svalinn/dagmc-ci-ubuntu-${UBUNTU_VERSION}-${COMPILER}-ext-hdf5_${HDF5}
+AARG OWNER=svalinn
```
gonuke:
```suggestion
ARG OWNER=svalinn
```

gonuke (changes requested):
This should fix the most recent error

bam241:
@gonuke thx for the suggestion ! it seems tone working now !


One need to make sure the retagging is working well, then I disable retag on personal branch and this should be good to go.

bam241:
not that I rearranged @Shimwell slicing to avoid uploading image we don't use or don't need to exchange between jobs...

.github/workflows/docker_publish.yml
```diff
@@ -148,46 +174,121 @@ jobs:
...
+
+      - name: Install DAGMC
+        run: |
+          cd $GITHUB_WORKSPACE
+          /bin/bash "CI/scripts/install.sh"
+
+      - name: Run DAGMC Test
+        run: |
+          cd $GITHUB_WORKSPACE
+          /bin/bash "/scripts/tests.sh"
```
gonuke: This is the wrong path.... maybe
```suggestion
          /bin/bash "CI/scripts/tests.sh"
```

gonuke (changes requested):
Another typo?

gonuke:
If we hard-coded the matrix, could we get more parallelism on the expensive builds?

bam241:
@gonuke I am happy with his.

package are tagged as stable when test pass (set on merge) https://github.com/bam241/DAGMC/pkgs/container/dagmc-ci-ubuntu-18.04-gcc-ext-hdf5_1.10.4-moab_develop

I also limited the retagging to svalinn only to limit the amount of pollution we are producing in users forks....

gonuke (approved):

bam241:
> If we hard-coded the matrix, could we get more parallelism on the expensive builds?

we already have parallelism for the 4 expensive builds

bam241:
see https://github.com/bam241/DAGMC/actions/runs/1042760146 
the `Matrix: build-deps-img` box

gonuke:
> > If we hard-coded the matrix, could we get more parallelism on the expensive builds?
> 
> we already have parallelism for the 4 expensive builds

OK! I must have been looking at something old that looked slower.

gonuke:
I'll wait for CircleCI (should be OK) and then merge

gonuke:
Thanks for tackling this @bam241

PR #761, opened 2021-07-19
test should always been run...

bam241:
@gonuke sorry it took an other round of fix...
but test should always been run to ensure everything is ok before tagging as stable...


gonuke:
Merging this to test merge behavior

PR #763, opened 2021-07-20
adding guard to run housekeeping on PR only

bam241:
Housekeeping should run on PR only 
(our detection script will not work on merge/push)


gonuke:
Thanks @bam241

PR #765, opened 2021-07-22
change tmp tag of docker images during testing

gonuke:
## Description
The docker images were being built during CI testing with tag `latest` which is actually a common/default way to reference the latest stable version.  This uses a different tag to avoid any ambiguiy.

## Motivation and Context
Avoid confusion in use of docker images.

## Changes
Change in workflow to name packages with `ci_testing`


gonuke:
testing under way [here](https://github.com/gonuke/DAGMC/actions/runs/1056259059)

bam241:
LGTM!

thx @gonuke

PR #769, opened 2021-07-22
Changelog not news

gonuke:
## Description
Transition from requiring NEWS files to updates of the CHANGELOG

## Motivation and Context
It takes time to combine all the NEWS files at the time of release and this can be distributed among authors of each change.

## Changes
New Github workflow to test for changes in CHANGELOG instead of existing of NEWS file.


gonuke:
Will consolidate NEWS before making this a real PR

bam241:
we also need another script for syntax.

gonuke:
Need to restore housekeeping because it does style test as well

bam241:
> we also need another script for syntax.

and maybe this could be added in the build and test workflow ?

gonuke:
> > we also need another script for syntax.
> 
> and maybe this could be added in the build and test workflow ?

Maybe as a separate PR

gonuke:
New docker images are being built [here](https://github.com/gonuke/DAGMC/actions/runs/1057725515) (updated)

bam241:
looking good merging after those test passes :)

bam241:
thx @gonuke 
merging

PR #771, opened 2021-07-22
switch to cmake and use version variable

gonuke:
This replaces #687 to move to CMake.

This also reverts some changes to get a smaller/faster clone of MOAB for a specific branch/tag instead of a sha.

### DIFF
--- a/CI/docker/build_moab.sh
+++ b/CI/docker/build_moab.sh
@@ -7,28 +7,33 @@ source ${docker_env}
 if [ ${MOAB_VERSION} == "master" ] || [ ${MOAB_VERSION} == "develop" ]; then
   branch=${MOAB_VERSION}
 else
-#  branch=Version${MOAB_VERSION}
   branch=${MOAB_VERSION}
 fi
 
 rm -rf ${moab_build_dir}/bld ${moab_install_dir}
 mkdir -p ${moab_build_dir}/bld
 cd ${moab_build_dir}
-#git clone --depth 1 https://bitbucket.org/fathomteam/moab -b ${branch}
-git clone https://bitbucket.org/fathomteam/moab 
-cd moab
-git checkout ${branch}
-autoreconf -fi
-cd ../bld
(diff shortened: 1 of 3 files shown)


gonuke:
This did NOT launch docker image builds in my repo?  I'm confused - did we break something?

gonuke:
New docker images being build [here](https://github.com/gonuke/DAGMC/actions/runs/1057855095)

CI/docker/build_moab.sh
```diff
@@ -7,28 +7,33 @@ source ${docker_env}
...
   branch=${MOAB_VERSION}
 fi
 
 rm -rf ${moab_build_dir}/bld ${moab_install_dir}
 mkdir -p ${moab_build_dir}/bld
 cd ${moab_build_dir}
-#git clone --depth 1 https://bitbucket.org/fathomteam/moab -b ${branch}
-git clone https://bitbucket.org/fathomteam/moab 
-cd moab
-git checkout ${branch}
```
gonuke, line 20: I realize now that this pattern was more flexible to different ways to specify a version.  Should we keep it this way?  It makes the git clone a little bulkier but should otherwise be equivalent. (@bam241 ?)

bam241: the previous one had to be tweaked to be used against specific commit.
I think your implementation is better as it is simpler.

gonuke: But if we ever need to go back to a specific commit, we'll have to change it all back.  If we leave it the previous way, we just need to change the VERSION string... I could go either way, but wanted to bring up that point.

bam241: I think it is better to make it more complicated to switch to single commit.

This is off normal implementation so that should remain "uneasy"

gonuke: Then I think this is ready - all the tests passed on my account

bam241:
This looks good to me,
thx @gonuke

Merging

PR #772, opened 2021-07-23
improve path pattern to trigger docker build workflow

bam241:
this should left out all the pieces we used for building and testing

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -3,7 +3,7 @@ name: Build & Publish docker image for DAGMC-CI
 on:
   push:
     paths:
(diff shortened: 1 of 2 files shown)


gonuke:
needs a rebase...

Did you cancel the build's for a reason - or do we just have to assume it will work since this PR will always trigger a build anyway?

bam241:
I did cancelled it because nothing but the trigger of the workflow changed.

this one triggering the build workflow is expected as we modified `.github/workflows/docker_publish.yml `

gonuke:
Thanks @bam241 - we'll keep an eye on this with future PR's to see that it's behaving as expected

PR #774, opened 2021-07-23
limits some workflows to pull_request trigger only

bam241:
limiting housekeeping workflow to pull_request only


gonuke:
Thanks @bam241

PR #779, opened 2021-08-16
Direct bounding box calls to relevant methods when using Double-Down

pshriwise:
@Shimwell noted that bounding boxes coming out of DAGMC when using Double-Down aren't correct. I thought I had included wrappers for the relevant methods in DAGMC, but it appears that slipped through the cracks. This updates our `getobb` calls to the correct methods on the `ray_tracer` object in Double-Down. Unfortunately, this means a couple more `#ifdef`'s for now as that method relies on the `RayTracingInterface` in Double-Down rather than the `GeomTopoTool`.

### DIFF
--- a/src/dagmc/DagMC.cpp
+++ b/src/dagmc/DagMC.cpp
@@ -369,7 +369,7 @@ ErrorCode DagMC::next_vol(EntityHandle surface, EntityHandle old_volume,
   return rval;
 }
 
-/* SECTION III */
+/* SECTION III: Indexing & Cross-referencing */
 
 EntityHandle DagMC::entity_by_id(int dimension, int id) {
   return GTT->entity_by_id(dimension, id);
@@ -445,7 +445,7 @@ ErrorCode DagMC::build_indices(Range& surfs, Range& vols) {
   return MB_SUCCESS;
 }
 
-/* SECTION IV */
+/* SECTION IV: : Handling DagMC settings */
 
 double DagMC::overlap_thickness() {
   return ray_tracer->get_overlap_thickness();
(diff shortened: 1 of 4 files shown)


pshriwise:
Realizing that we could've caught this earlier with a bounding box check, so I added one to the `dagmc_simple_test`. I confirmed that it fails before https://github.com/svalinn/DAGMC/pull/779/commits/c82c164d9c175a961470b3f3d11593f96d430c7f and passes after with Double-Down enabled.

shimwell:
Thanks @pshriwise much appreciated. I can check this out and report back

shimwell:
I've run the openmc bounding box task that was previously failing and it now works, nice work @pshriwise

src/dagmc/DagMC.cpp
```diff
@@ -445,7 +445,7 @@ ErrorCode DagMC::build_indices(Range& surfs, Range& vols) {
   return MB_SUCCESS;
 }
 
-/* SECTION IV */
+/* SECTION IV: : Handling DagMC settings */
```
```suggestion
/* SECTION IV: Handling DagMC settings */
```

gonuke (changes requested):
Thanks @pshriwise - feels almost petty to point out this inconsequential typo in a comment...

gonuke:
We can remove the `#ifdef` in the future with a `get_obb()` method in GQT that just wraps its GTT method of the same name, right?

src/dagmc/DagMC.cpp
pshriwise: Haha thanks for catching it!

pshriwise:
> We can remove the `#ifdef` in the future with a `get_obb()` method in GQT that just wraps its GTT method of the same name, right?

Yep, that should be totally possible. I've thought about merging those two classes (perhaps in DAGMC to start) so we have a single object for the ray tracing and geometry, but I haven't had time to take a crack at it.

gonuke (approved):

PR #780, opened 2021-08-20
Mac ci

bam241:
first attempt to setup MACOS CI

### DIFF
--- a/.github/workflows/mac_build_test.yml
+++ b/.github/workflows/mac_build_test.yml
@@ -23,27 +23,22 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Python 3.9
-        uses: actions/setup-python@v2
-        with:
-          python-version: 3.9
+
       - name: Initial setup
         shell: bash -l {0}
         run: |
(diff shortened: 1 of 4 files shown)


gonuke:
Maybe start with the build strategy in the condo-forge Moab-feedstock? (Just a thought)

.github/workflows/mac_build_test.yml
```diff
@@ -0,0 +1,91 @@
...
+      - develop
+      - master
+  release:
+    types: # This configuration does not affect the page_build event above
+      - created
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  main:
```
gonuke:
```suggestion
  BuildTest:
```

gonuke (changes requested):
Thanks for sorting this out @bam241 - a little change for consitency

gonuke:
Thanks for the quick work here @bam241. I'm merging this first because the rebase will probably be simpler this way around

pshriwise:
Thanks for tracking down the cause of the problem here @bam241! Sorry for not including that header initially.

PR #773, opened 2021-07-23
Attempt to ensure we are using python3 for amalgamation

bam241:
This attempt to enforce python3 usage when amalgamating PyNE

PR #792, opened 2022-02-07
Force python3 for the atomicgen.py script as well

ebknudsen:

## Description
Different conventions for calling python were used in this amalgamation. 

## Motivation and Context
If e.g. only python3 is installed on a machine the build script will fail.

## Changes
Bug fix

## Behavior
Current behavior is to call default python then 2 lines later call python3 explicitly. New behavior is to always use python3 explicitly.


ebknudsen:
I believe the changelog is up to date. I simply added this pr number to an existing line (#773, #792)

gonuke:
> I believe the changelog is up to date. I simply added this pr number to an existing line (#773, #792)

I agree - I have no idea why it skipped the changelog test 🤷 ?

PR #798, opened 2022-02-19
update version number and re-release

gonuke:
This only changes the version number in `CMakeLists.txt` file.  I propose we force this through to update the stack without a complete cycle of CI.


gonuke:
for example, I didn't touch the `ChangeLog`

shimwell:
Thanks for this change, sorry for missing it.

We even have this as a step in the release doc https://github.com/svalinn/DAGMC/blob/develop/release.rst

shimwell (approved):

pshriwise:
> Thanks for this change, sorry for missing it.

Same same. Thanks @gonuke!

pshriwise (approved):

PR #800, opened 2022-02-20
Remove superfluous headers from DAGMC headers

### LINKED ISSUE #799: Remove need for downstream codes to have Eigen3 installed
**Describe the Bug**<br/>
Currently, any downstream code that links to DAGMC must have Eigen3 installed even if it has no direct dependence on Eigen3.  Eigen3 is a source-code-header only library, so should be compiled into DAGMC libraries and not necessary to build dependent projects.  This problem exists because the DAGMC header is "polluted" with too many other headers.

**To Reproduce**<br/>
Try to build a downstream code, e.g. OpenMC (see discussion in #739), without having Eigen3 installed.

**Expected Behavior**<br/>
Build a downstream code that has no independent/direct dependency on Eigen3 without having Eigen3 installed.

gonuke:
## Description
Move headers from DAGMC's headers to the implementation files.  Upstream headers only need to be included in header files if details of the classes that those upstream headers define are accessed within the header.  If only the existence of those classes is necessary, then forward declaration of the classes is sufficient.

## Motivation and Context
Since DAGMC's headers may be included by downstream codes, it is good practice to avoid "polluting" our headers with other headers that may, unkown to us, further include other headers and so on.  This 

Fixes #799 

## Changes
Fix of an annoying but non-pathologial bug

### DIFF
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -47,9 +47,6 @@ v3.2.2
    * Removing build of static libs as a default option (#802)
    * Adding PYTHONPATH to linux CI images and creating pymoab installation directories (#802)
 
-v3.2.1
-====================
-
 **Added:**
 
    * Added link to latest Conda package in GitHub README.md (#801)
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -35,6 +35,8 @@ Next version
    * Tweak conda environment for Windows build to avoid conflicting gtest headers (#888)
    * Restrict cython version for MOAB (#893)
    * Various documentation updates (#869)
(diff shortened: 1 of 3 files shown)


gonuke:
Converted this to draft because it will be more challenging than originally expected.

ahnaf-tahmid-chowdhury:
I think we can move this branch to upstream and give this a update?

gonuke:
Why move to upstream?

ahnaf-tahmid-chowdhury:
Not mandatory, as we don't need to work on Dockerfile. But, couldn't it  be easy for me to fork? Or I may add your remote branch to my repo.

PR #804, opened 2022-04-12
Ensure PYTHONPATH and pymoab install directory are present for Linux CI builds

pshriwise:
## Description
This adds the installation directory of PyMOAB to the `PYTHONPATH` env var in the linux builds used in CI. It also ensures that the installation directory is present by explicitly creating that directory before building MOAB. 

I recognize that this is a band-aid, but the root problem is almost certainly upstream in MOAB and that will take time to remedy.

## Motivation and Context
CI is broken.

## Changelog file
~I don't think this change really warrants an entry in the changelog, but I will add one if requested.~

EDIT: Nevermind I guess we _require_ a changelog check.

### DIFF
--- a/.github/workflows/linux_build_test.yml
+++ b/.github/workflows/linux_build_test.yml
@@ -10,7 +10,7 @@ on:
       - '.github/workflows/docker_publish.yml'
       - '.github/workflows/housekeeping.yml'
       - 'CI/**'
-      - 'news/**' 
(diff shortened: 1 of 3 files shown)


gonuke:
Yikes... When I first looked at this, I couldn't figure out how it made a difference... and then I tried my own PR...

Thanks for this band-aid @pshriwise

PR #807, opened 2022-05-18
V322

gonuke:
Update version numbers and pyne hash for microrelease


ngranda:
Thanks @gonuke! One step closer to the micro release

PR #808, opened 2022-05-19
Perform microrelease for v3.2.2

gonuke:
Update version numbers and pyne hash, as well as a tweak for the changelog testing based on GH security changes


ngranda:
Looks like all test passed! Thanks @gonuke

PR #816, opened 2022-07-11
Adding safe folder to allow CI to compile DAGMC

shimwell:
## Description
Attempt to fix the CI that builds and publishes the docker images

## Motivation and Context
Why is this change required? What problem does it solve?
The latest action to publish the docker image failed with this error message 
```
CMake Warning at CMakeLists.txt:27 (message):
-- Submodule update
  Could not determine the commit SHA for DAGMC.
fatal: unsafe repository ('/__w/DAGMC/DAGMC' is owned by someone else)
To add an exception for this directory, call:
	git config --global --add safe.directory /__w/DAGMC/DAGMC
CMake Error at CMakeLists.txt:38 (message):
  git submodule update --init --recursive failed with 128, please checkout
  submodules
See also "/root/build_dir/DAGMC-static/bld/CMakeFiles/CMakeOutput.log".
```
https://github.com/svalinn/DAGMC/actions/runs/2637740871

This PR adds the folder to the safe directory folders

## Changes
bug fix

## Behavior
What is the current behavior? What is the new behavior?
broken CI - hopefully a fixed CI (I can't be sure as file owership issues in the CI can't be fully reproduced)

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -218,7 +218,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
-          set-safe-directory: "/__w/DAGMC/DAGMC"
+          set-safe-directory: ${GITHUB_WORKSPACE}
 
 
       - name: Setup
@@ -266,7 +266,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
-          set-safe-directory: "/__w/DAGMC/DAGMC"
+          set-safe-directory: ${GITHUB_WORKSPACE}
 
       - name: Setup
         run: |
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -218,6 +218,8 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
+          set-safe-directory: 'DAGMC/DAGMC'
+
 
       - name: Setup
         run: |
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -163,7 +163,7 @@ jobs:
           1.10.4,
         ]
         moab_versions : [
-          5.3.1,
+          5.4.0,
           develop,
           master,
         ]
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -218,11 +218,10 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
-          set-safe-directory: ${GITHUB_WORKSPACE}
-
 
       - name: Setup
         run: |
+          git config --global --add safe.directory ${GITHUB_WORKSPACE}
           mkdir /root/build_dir/
           echo "REPO_SLUG=${GITHUB_REPOSITORY}" >> $GITHUB_ENV
           echo "PULL_REQUEST=$(echo  $GITHUB_REF | cut -d"/" -f3)" >> $GITHUB_ENV
@@ -266,10 +265,10 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
-          set-safe-directory: ${GITHUB_WORKSPACE}
 
       - name: Setup
         run: |
+          git config --global --add safe.directory ${GITHUB_WORKSPACE}
           echo "MOAB_VERSION=${{ matrix.moab_versions }}" >> $GITHUB_ENV
           echo "COMPILER=${{ matrix.compiler }}" >> $GITHUB_ENV
           echo "HDF5_VERSION=${{ matrix.hdf5_versions }}" >> $GITHUB_ENV
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -218,7 +218,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
-          set-safe-directory: '/__w/DAGMC/DAGMC'
+          set-safe-directory: "/__w/DAGMC/DAGMC"
 
 
       - name: Setup
@@ -266,6 +266,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
+          set-safe-directory: "/__w/DAGMC/DAGMC"
 
       - name: Setup
         run: |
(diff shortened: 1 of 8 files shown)


.github/workflows/docker_publish.yml
```diff
@@ -261,9 +263,10 @@ jobs:
     name: Installing and Testing DAGMC
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           submodules: recursive
+          set-safe-directory: "/__w/DAGMC/DAGMC"
```
pshriwise: For my education, what does the syntax here mean? Is this the default location of repo checkouts using the GHA tool?

I think we want to be explicit about this safe directory as you've done here @shimwell, but I'll paste this option here too from the relevant change in the GHA checkout action
 PR in case it's useful in the future. 

https://github.com/actions/checkout/pull/770#issuecomment-1104448853

shimwell: this was based on the git error message, that identified the folder directory
```git config --global --add safe.directory /__w/DAGMC/DAGMC```

**Review by pshriwise — APPROVED** [review]

File: .github/workflows/docker_publish.yml [hunk 2]
```diff
@@ -215,9 +215,11 @@ jobs:
     name: Running Housekeeping scrips
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           submodules: recursive
+          set-safe-directory: "/__w/DAGMC/DAGMC"
```
gonuke: Is this supposed to work?  It's essentially what I did in #807 
```suggestion
          set-safe-directory: ${GITHUB_WORKSPACE}
```

.github/workflows/docker_publish.yml
gonuke: See above comment for version using variable

gonuke (commented):
Suggested using the `${GITHUB_WORKSPACE}` variable instead

.github/workflows/docker_publish.yml
shimwell:
```suggestion
          set-safe-directory: ${GITHUB_WORKSPACE}
```

gonuke:
Judging from the action logs in your account, I don't think this works.  I think it needs to be added with a `git` command as in `changelog_test.yml`

shimwell:
> Judging from the action logs in your account, I don't think this works. I think it needs to be added with a `git` command as in `changelog_test.yml`

Rerunning the docker build CI on my fork https://github.com/shimwell/DAGMC/actions/runs/2672087279

shimwell:
Looks like there is [another error](https://github.com/shimwell/DAGMC/runs/7348179098?check_suite_focus=true) while compiling moab in the docker file.

```
error: bad install directory or PYTHONPATH
```

bam241:
I think the root problem occurred before:
```
[ 55%] Building CXX object test/CMakeFiles/file_options_test.dir/__/src/FileOptions.cpp.o
[1479](https://github.com/shimwell/DAGMC/runs/7348179098?check_suite_focus=true#step:5:1480)
The directory '/github/home/.cache/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
[1480](https://github.com/shimwell/DAGMC/runs/7348179098?check_suite_focus=true#step:5:1481)
The directory '/github/home/.cache/pip' or its parent directory is not owned by the current user and caching wheels has been disabled. check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
[1481](https://github.com/shimwell/DAGMC/runs/7348179098?check_suite_focus=true#step:5:1482)
Obtaining file:///root/build_dir/moab/bld/pymoab

```

I think we need to declare as safe both the build dir and the install dir.

shimwell:
Looks like the latest error is due to PyMoab being installed into a folder that is not in the PYTHONPATH and also the PYTHONPATH is empty :eyes: 

https://github.com/shimwell/DAGMC/runs/7362738426?check_suite_focus=true

gonuke:
I guess that didn't work - Maybe try the 5.3.2 release candidate now?

.github/workflows/docker_publish.yml
```diff
@@ -163,7 +163,7 @@ jobs:
           1.10.4,
         ]
         moab_versions : [
-          5.3.0,
+          5.3.1,
```
shimwell:
```suggestion
          5.4.0,
```

shimwell:
Looks like adding the PYTHONPATH resulted in the dockerfiles building
https://github.com/shimwell/DAGMC/actions/runs/2780302592

Nice work @bam241

CI/Dockerfile
```diff
@@ -80,6 +80,8 @@ RUN /root/etc/CI/docker/build_hdf5.sh
 
 FROM hdf5 AS moab
 
+ENV PYTHONPATH=/root/build_dir/moab/bld/pymoab/lib/python3.6/site-packages
```
gonuke, line 83: This makes me wonder if we are installing MOAB in the wrong place?  Or not installing it at all?  Or we are installing it in the right place and we should reference *that* place here instead?

gonuke (changes requested):
While the addition of the `PYTHONPATH` allows it to work, I'm dubious of whether it's the 100% correct solution.  Maybe it's temporary until we untangle the other parts of the build process?

CI/Dockerfile
shimwell: we could add a python import command to the dockerfile to see if it is found by python if that helps check it it was installed

```bash
python -c "import pymoab"
```

bam241: The problem does not come from us.
It comes from Moab.
when they build and install `pymoab` they install it locally with `pip -e` for testing purposes
The problem is this is prevented because `/root/build_dir/moab/bld/pymoab/lib/python3.6/site-packages` is not in our `PYTHONPATH`

see:
[here in pymoab CMakelist.txt](https://bitbucket.org/fathomteam/moab/src/e6a23a793c5f140f12928c64c5ab57fac5b85bb6/pymoab/CMakeLists.txt?at=master)

gonuke: Hmmm… maybe the problem is that line 91 in the pymoab CMake is commented out?!?

gonuke: Looks like @pshriwise commented this out in 2018 🤣

shimwell: Wow that survived a long time

gonuke:
Thanks @shimwell - we'll go with this pending a future change in MOAB

PR #855, opened 2023-02-13
Graveyard updates

pshriwise:
## Description
These changes update some behavior of the `DagMC::create/remove_graveyard` function. 

### `create_graveyard`
  - `create_graveyard` can now create a graveyard without acceleration data structures using the vertices of the model to create a bounding box as a basis for the graveyard volume.
  - `create_graveyard` will create an implicit complement if none exists so the graveyard volume can be inserted into the model topology correctly.
  - `create_graveyard` will only (re)construct acceleration data structures for the new graveyard volume and the implicit complement if  these data structures were present when the method was initially called.

### `remove_graveyard`

There's a bug when calling `create_graveyard` with `overwrite == true` when a graveyard is not present. Previously, this would fail when trying to remove a graveyard that isn't present. This adds a check at the beginning of `remove_graveyard` to simply return `MB_SUCCESS` if no graveyard is present.

### DIFF
--- a/src/dagmc/DagMC.cpp
+++ b/src/dagmc/DagMC.cpp
@@ -555,8 +555,10 @@ ErrorCode DagMC::create_graveyard(bool overwrite) {
   MB_CHK_SET_ERR(rval, "Failed to update the geometry sets");
 
   // delete the implicit complement tree (but not the surface trees)
-  rval = remove_bvh(implicit_complement, true);
-  MB_CHK_SET_ERR(rval, "Failed to delete the implicit complement tree");
+  if (has_acceleration_datastructures()) {
+    rval = remove_bvh(implicit_complement, true);
+    MB_CHK_SET_ERR(rval, "Failed to delete the implicit complement tree");
+  }
 
   // create BVH for both the new implicit complement and the new graveyard
   // volume
--- a/src/dagmc/DagMC.cpp
+++ b/src/dagmc/DagMC.cpp
@@ -261,7 +261,6 @@ ErrorCode DagMC::get_graveyard_group(EntityHandle& graveyard_group) {
 }
 
 ErrorCode DagMC::remove_graveyard() {
-
   if (!has_graveyard()) return MB_SUCCESS;
 
   ErrorCode rval;
@@ -412,7 +411,8 @@ ErrorCode DagMC::create_graveyard(bool overwrite) {
       // get the bounding box of the volume
       moab::EntityHandle surf = this->entity_by_index(2, i + 1);
       moab::Range vertices;
-      rval = this->moab_instance()->get_entities_by_type(surf, moab::MBVERTEX, vertices);
+      rval = this->moab_instance()->get_entities_by_type(surf, moab::MBVERTEX,
+                                                         vertices);
       MB_CHK_SET_ERR(rval, "Failed to get surface vertices");
       double coords[3];
       for (auto vertex : vertices) {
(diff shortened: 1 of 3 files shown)


src/dagmc/DagMC.cpp
```diff
@@ -391,33 +390,51 @@ ErrorCode DagMC::create_graveyard(bool overwrite) {
...
   ErrorCode rval;
 
-  // remove existing graveyard
+  // remove existing graveyard if overwrite is true
   if (overwrite) {
     remove_graveyard();
   }
 
-  // if a graveyard already exists at this point,
+  // if a graveyard already exists and we aren't overwriting it,
```
gonuke, line 397: If we are overwriting, then there shouldn't be a graveyard anymore.  Isn't this redundant?

gonuke: Also, doesn't the previous form of this catch a failure to remove the graveyard?

src/dagmc/DagMC.cpp
```diff
@@ -527,20 +555,25 @@ ErrorCode DagMC::create_graveyard(bool overwrite) {
...
-  rval = build_bvh(volume_set);
-  MB_CHK_SET_ERR(
-      rval,
-      "Failed to build accel. data structure for the new graveyard volume");
-
-  rval = build_bvh(implicit_complement);
-  MB_CHK_SET_ERR(
-      rval,
-      "Failed to build accel. data structure for the new implicit complement");
+  if (has_acceleration_datastructures()) {
```
gonuke: Combine this with previous conditional?

gonuke (changes requested):
Thanks @pshriwise - just a couple of maintainability questions

src/dagmc/DagMC.cpp
pshriwise: Mmm yeah you're right. Thanks!

src/dagmc/DagMC.cpp
pshriwise: Yep! Late night change.

makeclean:
Should we ever replace an existing graveyard? Should we just not add another to the set?

pshriwise:
> Should we ever replace an existing graveyard? Should we just not add another to the set?

I can see someone wanting to replace a spherical shell graveyard in case they're learning DAGMC and didn't realize how many triangles this might add to their model without having to re-tessellate the entire thing. Granted, some kind of utility program in the repo would be better than writing it yourself own, which is what they'd have to now.

gonuke (approved):

gonuke:
I'm ready to merge but want to give @makeclean a chance to either withdraw or amplify his concern?

src/dagmc/DagMC.cpp
```diff
@@ -471,9 +488,21 @@ ErrorCode DagMC::create_graveyard(bool overwrite) {
...
 
   // tear down the implicit complement tree
-  EntityHandle implicit_complement;
+  EntityHandle implicit_complement = 0;
   rval = geom_tool()->get_implicit_complement(implicit_complement);
-  MB_CHK_SET_ERR(rval, "Failed to get the implicit complement");
+  if (rval != MB_ENTITY_NOT_FOUND && rval != MB_SUCCESS) {
+    MB_CHK_SET_ERR(rval, "Could not get the implicit complement");
+  }
+  // create the implicit complement if it coesn't exist at this point
```
```suggestion
  // create the implicit complement if it doesn't exist at this point
```

gonuke (changes requested):
Only one typo in a comment....

gonuke:
Thanks @pshriwise

gonuke (approved):

PR #859, opened 2023-02-17
removed unused circle ci script

shimwell:
## Description
This PR is an attempt to remove a single config.yml file that was part of the Circle CI system which is not longer used. Also I think if we do switch back to circle CI this script would need updating so it is not so useful to keep it around now we have moved to github actions (and do a lot more CI than previously).

## Motivation and Context
unused file

## Changes
reduced repo size

## Behavior
no change


gonuke:
Thanks for the cleanup @shimwell

PR #822, opened 2022-07-25
Putting install scripts into dockerfile

shimwell:
work in progress, not ready for review

## Description
As mentioned in project https://github.com/svalinn/DAGMC/projects/6 it has been proposed to move the install scripts into the dockerfile.

## Motivation and Context
This simplifies the passing of environments to cmake and other commands as they are all visible in one file and no longer creating in one script to be used in another. 

## Changes
What kind of change does this PR introduce? (Bug fix, feature, documentation update...)
refactored dockerfile to include more ARGs ENVs and install commands.

## Behavior
What is the current behavior? What is the new behavior?
old = dockerfile copies in sh scripts and runs them, new = dockerfile runs install commands directly

Issue #858, opened 2023-02-17
DBUILD_MW_REG_TESTS requires h5m models that can't be found as URL missing


shimwell:
**Describe the Bug**<br/>
The CI has a variable called ```DBUILD_MW_REG_TESTS``` which is if set will trigger the downloading of a  tar.gz file containing some h5m geometries from a URL. Unfortunately it looks like the URL used by the wget command has been lost in in the CI transition from travis-ci to circle-ci. I understand that these models are large and the tests are lengthy hence it was an optional test. It looks like we have accidentally gone on for a while without noticing that these tests were not running any more and this was recently noticed during  #822 

**To Reproduce**<br/>
compile dagmc with the ```DBUILD_MW_REG_TESTS``` variable
```
mkdir build
cd build
cmake .. DBUILD_MW_REG_TESTS=ON
make -j
make install
make test
```
**Expected Behavior**<br/>
tests should pass

**Please complete the following information regarding your system:**<br/>
all OS and MOAB versions

**Additional Context**<br/>
[Link to PR ](https://github.com/svalinn/DAGMC/pull/822#issuecomment-1425724550)where this issue was found and is also discussed

gonuke:
I tagged @pshriwise only because he may actually know where the necessary H5M files are for testing....

shimwell:
Filenames of the missing files

```
class ITERMakeWatertightRegressionTest : public MakeWatertightTest {
 protected:
  virtual void setFilename() { filename = "iter_imprinted.h5m"; };
};

class FNSFMakeWatertightRegressionTest : public MakeWatertightTest {
 protected:
  virtual void setFilename() { filename = "model1_360_1.h5m"; };
};

class BLiteMakeWatertightRegressionTest : public MakeWatertightTest {
 protected:
  virtual void setFilename() { filename = "bllite30matls.h5m"; };
};
```

ahnaf-tahmid-chowdhury:
Is the issue already fixed?

shimwell:
I think this one was never solved. But perhaps not a high priority

gonuke:
I think we lost the files, right @pshriwise ?

ahnaf-tahmid-chowdhury:
Is it possible to recreate the files if they are lost? I am willing to work on that. I need instructions on how to create h5m geometries for this.

gonuke:
These were a set of specific H5M files that were useful for these tests.  It is not as simple as just creating some new H5M files.

[...]

PR #803, opened 2022-04-06
Patch to compile with Geant4 10.6

helen-brooks:
## Description
Patch to permit compilation with Geant4 10.6, solves #696 

## Motivation and Context
From the Geant4 10.6 Release notes:
- **The preprocessor macros G4UI_USE and G4VIS_USE are deprecated in 10.6** - these are used in /src/geant4/app/exampleN01.cc to test whether to include some headers. 
- **added namespace scope to MeshScoreMap** - the class MeshScoreMap has been moved inside the class G4VScoringMesh. The file ExN01UserScoreWriter.cc needs the statement
using MeshScoreMap = G4VScoringMesh::MeshScoreMap;

## Changes
- Modify cmake files to check geant4 version, and define a preprocessor macro GEANT4_GT_10_6
- Use preprocessor directive to include headers and define G4VScoringMesh alias in backwards-compatible way.

## Behavior
DAG-Geant4 now compiles with Geant4 10.6 

## Other Information
I haven't touched the CI. I don't know what you want to do about this - presumably it's overkill to add yet another variation to the CI matrix....?

PR #863, opened 2023-03-02
single dockerfile and single step in the docker building CI

shimwell:
## Description
This PR is an alternative to #822 
In addition to moving the scripts into a single dockerfile this PR also changes the CI for publishing dockerfiles.
Currently we had a branching and caching approach to the CI dockerfile building that is efficient but complex.
This approach reduces the complexity of the CI and allows all the perturbations to run in parallel. 

This might actually end up being quicker to build and quicker to fail when something is broken as the CI doesn't have breaks waiting for all the docker stages to reach the same stage to proceed to the next stage. Also if caching can be figured out then this workflow could make use of pre-built images to be very fast

## Motivation and Context
simpler is nice for maintainers :smile: 

## Changes
refactored CI

## Behavior
single dockerfile instead of scripts and simpler CI


shimwell:
Action is running on my branch with 24 jobs in parallel 
https://github.com/shimwell/DAGMC/actions/runs/4315972934
Screenshot from 2023-03-02 16-13-40

shimwell:
All jobs on my fork CI passed, it took 3 hours 28mins

.github/workflows/docker_publish.yml
```diff
@@ -96,16 +26,25 @@ jobs:
...
+         ON,
         ]
 
-    name: Ubuntu ${{ matrix.ubuntu_versions }} + ${{ matrix.compiler }} hdf5 ${{ matrix.hdf5_versions }}
-    steps:
+    name: Installing DAGMC
+    steps:        
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
+        uses: docker/setup-qemu-action@v2
```
gonuke: We don't need QEMU

.github/workflows/docker_publish.yml
```diff
@@ -114,266 +53,62 @@ jobs:
...
         uses: actions/checkout@v3
 
-      - name: Build and push Dockerfile_1_external_deps
-        uses: docker/build-push-action@v2
+      - name: Install DAGMC in Docker image
+        uses: docker/build-push-action@v3
         with:
           file: CI/Dockerfile
-          target: external_deps
+          target: dagmc
```
gonuke: Recall that this should only build the images to be used in CI for building/testing DAGMC.

We should only build to `moab`, and then test using this image for building DAGMC as we will later be doing in CI.

.github/workflows/docker_publish.yml
```diff
@@ -114,266 +53,62 @@ jobs:
...
-        with:
-          registry: ghcr.io
-          username: ${{ github.repository_owner }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Checkout repository
-        uses: actions/checkout@v3
+            MOAB=${{ matrix.moab_versions }}
+            build_mw_reg_tests=ON
+          tags: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}-dagmc:ci_testing
```
gonuke: If this is `push: false` then tags don't matter, but I think we **should** push.

.github/workflows/docker_publish.yml
```diff
@@ -114,266 +53,62 @@ jobs:
...
-
-    name: Ubuntu${{ matrix.ubuntu_versions }} + ${{ matrix.compiler }} hdf5 ${{ matrix.hdf5_versions }} + Moab ${{ matrix.moab_versions }}':' latest -> stable
+            build_mw_reg_tests=ON
+          tags: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}-dagmc-test:ci_testing
 
-    steps:
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
+        if: ${{ github.repository_owner == 'svalinn' }}
+        uses: docker/setup-qemu-action@v2
```
gonuke: Also don't need QEMU here

.github/workflows/docker_publish.yml
```diff
@@ -5,86 +5,16 @@ on:
...
-          target: base
-          push: true
-          context: .
-          build-args: |
-            UBUNTU_VERSION=${{ matrix.ubuntu_versions }}
-          tags: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}:ci_testing
-
-  build-housekeeping-img:
-    needs: build-base-img
+  build-test-dagmc-img:
```
gonuke: rename based on comments below

.github/workflows/docker_publish.yml
```diff
@@ -114,266 +53,62 @@ jobs:
...
-
-      - name: Checkout repository
-        uses: actions/checkout@v3
+            MOAB=${{ matrix.moab_versions }}
+            build_mw_reg_tests=ON
+          tags: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}-dagmc:ci_testing
 
-      - name: Build and push Dockerfile_3_moab
-        uses: docker/build-push-action@v2
+      - name: Test DAGMC in Docker image
```
gonuke, line 93: Testing these images should not use our Dockerfile, but should use GH actions directly using one of the docker images we have prepared.

(Also - this starts the build from scratch, hence the long times for this action)

gonuke (changes requested):
Thanks for getting this started @shimwell.

I think it went a little to far/too fast and has lost the narrative of what we want to do:
1. build docker images to be used by CI later on that only have appropriate dependencies - up to moab (this part can proceed in parallel)
2. push those images to the registry to be used later using a temporary tag
3. build and test DAGMC using those images as the container (this part can proceed in parallel)
4. rename the docker images in the registry to a more stable tag

BONUS: figure out how to populate a cache for the first docker build using the existing docker images available in the registry. (separate PR?)

shimwell:
Had a go at all those comments and pushed, CI is running on my fork https://github.com/shimwell/DAGMC/actions/runs/4325712183

shimwell:
CI on fork passed https://github.com/shimwell/DAGMC/actions/runs/4325712183

2nd stage only took a few seconds so the local stage cache appears to have worked. 

This used the local stage cache instead of the container repo. But the layer has the same tag locally and I'm the container repo.

Perhaps this is not what we want

.github/workflows/docker_publish.yml
```diff
@@ -96,16 +26,22 @@ jobs:
...
+         1.10.4,
+        ]
+        moab_versions : [
+         5.3.0,
+         develop,
+         master,
+        ]
+        static_exe : [
+         OFF,
+         ON,
```
gonuke: We don't appear to be passing this into the docker build or the tag naming and therefore have duplicate images being built and tagged.  We should discuss whether we really need the statically linked version.  Not sure I remember who requires/desires that.

.github/workflows/docker_publish.yml
```diff
@@ -184,196 +109,42 @@ jobs:
...
-         master,
-        ]
-
-    name: Ubuntu${{ matrix.ubuntu_versions }} + ${{ matrix.compiler }} hdf5 ${{ matrix.hdf5_versions }} + Moab ${{ matrix.moab_versions }}':' latest -> stable
-
-    steps:
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
+            build_mw_reg_tests=ON
+            dependency_image_location=ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}:ci_testing
```
gonuke: happy to see this working, but wonder if we are better off with a different strategy that uses the cache instead? something to discuss.

gonuke (changes requested):
This definitely looks simpler and probably easier to maintain over the long term.  I think there is still a discussion point about how to use the docker image cache.

.github/workflows/docker_publish.yml
```diff
@@ -141,14 +63,17 @@ jobs:
             UBUNTU_VERSION=${{ matrix.ubuntu_versions }}
             COMPILER=${{ matrix.compiler }}
             HDF5=${{ matrix.hdf5_versions }}
-          tags: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}:ci_testing
+            MOAB=${{ matrix.moab_versions }}
+            build_mw_reg_tests=ON
```
gonuke: don't we want this off?

.github/workflows/docker_publish.yml
```diff
@@ -184,196 +109,42 @@ jobs:
...
-         develop,
-         master,
-        ]
-
-    name: Ubuntu${{ matrix.ubuntu_versions }} + ${{ matrix.compiler }} hdf5 ${{ matrix.hdf5_versions }} + Moab ${{ matrix.moab_versions }}':' latest -> stable
-
-    steps:
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
+            build_mw_reg_tests=ON
```
gonuke: Don't we want this off?

gonuke (changes requested):
a couple of additional comments

.github/workflows/docker_publish.yml
```diff
@@ -184,196 +109,42 @@ jobs:
...
 
-      - name: Build and push Dockerfile_3_moab
-        uses: docker/build-push-action@v2
+      - name: Test DAGMC in Docker image
+        uses: docker/build-push-action@v3
         with:
           file: CI/Dockerfile
-          target: moab
+          target: dagmc_test
           context: .
```
pshriwise: I'm thinking we may not need this line in this case. The [action documentation](https://github.com/docker/build-push-action#git-context) indicates that the right version of the repository will be checked out in the Docker image by default. This isn't true in the case of our dependencies though, so I can see why we'd need it there.

bquan0:
I just opened a [PR](https://github.com/shimwell/DAGMC/pull/5) to the branch this PR is on to replace the docker build actions with the [multistage build action](https://github.com/marketplace/actions/multi-stage-docker-build).

gonuke:
@shimwell - this appears to need a rebase...

gonuke:
This also failed here (https://github.com/shimwell/DAGMC/actions/runs/4752350398/jobs/8442595961) - apparently it failed to push to GHCR?  and it was successful for @bquan0 here (https://github.com/bquan0/DAGMC/actions/runs/4749240986/jobs/8436860572) but was based on cached version??

bquan0:
I ran into that problem a few times on my workflow runs too and I mentioned it in [the PR](https://github.com/shimwell/DAGMC/pull/5). I solved it by going to the settings of the package it was trying to push to, then checking the DAGMC repo under the "Manage Actions access" section.

CI/Dockerfile
```diff
@@ -1,9 +1,43 @@
...
 ARG UBUNTU_VERSION=18.04
+ARG MOAB_BRANCH=5.3.0
+ARG double_down=OFF
+ARG static_exe=OFF
+ARG ci_jobs=20
+
+# dependency_image_location defaults to using a local image for all the dependencies
+# This also allows one to specify a remote image from a container registry
+# The DAGMC container registry can be found here https://github.com/orgs/svalinn/packages
+ARG dependency_image_location=moab
```
gonuke: I don't think we need this anymore thanks to the multistage-build

CI/Dockerfile
```diff
@@ -1,9 +1,43 @@
+# Svalinn DAGMC Dockerfile
+
+# build from repository root directory with command
+# docker build -t dagmc -f CI/Dockerfile .
+
+# Global ARGS set before the first build stage are accessable by all build stages
+ARG build_mw_reg_tests=OFF
```
gonuke: I think we ignore this right now since MW Regression testing is broken

.github/workflows/docker_publish.yml
```diff
@@ -253,121 +82,52 @@ jobs:
...
-         5.3.0,
-        ]
-
-    name: Ubuntu${{ matrix.ubuntu_versions }} + ${{ matrix.compiler }} hdf5 ${{ matrix.hdf5_versions }} + Moab ${{ matrix.moab_versions }}':' latest -> stable
-
-    steps:
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
+          repository: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}
+          stages: base, external_deps, hdf5, moab, dagmc
```
gonuke: We don't need to explicitly build all these stages at this point.  Most were just built anyway.  This action should use the container repository to blow past these anyway, but for maintainability I recommend we remove all except `dagmc`

.github/workflows/docker_publish.yml
```diff
@@ -182,56 +48,19 @@ jobs:
...
-          echo "REPO_SLUG=${GITHUB_REPOSITORY}" >> $GITHUB_ENV
-          echo "PULL_REQUEST=$(echo  $GITHUB_REF | cut -d"/" -f3)" >> $GITHUB_ENV
-          ln -s $GITHUB_WORKSPACE /root/build_dir/DAGMC
+          repository: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}
+          stages: base, external_deps, hdf5
+          server-stage: moab
+          quiet: false
+          tag-latest-on-default: false
+          dockerfile: CI/Dockerfile
+          build-args: OWNER=${{ github.repository_owner }}, TAG=ci_testing, UBUNTU_VERSION=${{ matrix.ubuntu_versions }}, COMPILER=${{ matrix.compiler }}, HDF5=${{ matrix.hdf5_versions }}, MOAB=${{ matrix.moab_versions }}, build_mw_reg_tests=ON
```
gonuke: I don't think we use the `TAG` build-arg anywhere

gonuke: or the `OWNER` arg?

CI/Dockerfile
```diff
@@ -33,63 +61,212 @@ RUN apt-get -y update; \
...
+    # cmake ../moab -DENABLE_HDF5=ON -DHDF5_ROOT=${hdf5_install_dir} \
+    #               -DCMAKE_INSTALL_PREFIX=${moab_install_dir} \
+    #               -DCMAKE_CXX_COMPILER=${CXX} \
+    #               -DCMAKE_C_COMPILER=${CC} \
+    #               -DENABLE_BLASLAPACK=OFF \
+    #               -DBUILD_SHARED_LIBS=OFF \
+    #               -DENABLE_FORTRAN=OFF && \
+    # make -j${ci_jobs} && \
+    # make install && \
+    # rm -rf * && \
```
shimwell:
```suggestion
```
we can remove this

CI/Dockerfile
```diff
@@ -33,63 +61,212 @@ RUN apt-get -y update; \
...
+                  -DCMAKE_INSTALL_PREFIX=${moab_install_dir} \
+                  -DCMAKE_CXX_COMPILER=${CXX} \
+                  -DCMAKE_C_COMPILER=${CC} \
+                  -DENABLE_BLASLAPACK=OFF \
+                  -DBUILD_SHARED_LIBS=ON \
+                  -DENABLE_FORTRAN=OFF \
+                  -DENABLE_PYMOAB=ON && \
+    mkdir -p ${moab_build_dir}/build/pymoab/lib/python3.6/site-packages && \
+    PYTHONPATH=${moab_build_dir}/build/pymoab/lib/python3.6/site-packages:${PYTHONPATH} make -j${ci_jobs} && \
+    PYTHONPATH=${moab_build_dir}/build/pymoab/lib/python3.6/site-packages:${PYTHONPATH} make install && \
```
gonuke: remove any reference to this hardcoded python3.6 site-packages path
```suggestion
    make -j${ci_jobs} && \
    make install && \
```

.github/workflows/docker_publish.yml
```diff
@@ -253,121 +82,52 @@ jobs:
...
-        ubuntu_versions : [18.04,]
-
-    name: Housekeeping':' latest -> stable
-    steps:
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
+        static_exe : [
+         OFF,
+         ON,
+        ]
```
gonuke: unused build-arg

CI/Dockerfile
shimwell: resolved by removed this arg from the dockerfile

CI/Dockerfile
shimwell: Yes and this is tracked in an issue
https://github.com/svalinn/DAGMC/issues/858

Resolved as I have removed this build ARG

CI/Dockerfile
shimwell: removed ✔️

.github/workflows/docker_publish.yml
shimwell: resolved by removing these lines from the matrix

.github/workflows/docker_publish.yml
shimwell: resolved by removing both of these build args from the command

.github/workflows/docker_publish.yml
shimwell: resolved by removing all but dagmc from the stages

.github/workflows/docker_publish.yml
shimwell: this line has been removed from the yml file so I think this one is resolved

.github/workflows/docker_publish.yml
shimwell: this build arg has been removed from the yml and dockerfile so I think this is now resolved

.github/workflows/docker_publish.yml
shimwell: this build arg has been removed from the yml and dockerfile so I think this is now resolved

.github/workflows/docker_publish.yml
shimwell: dependency_image_location has now been removed and the PR has moved to a multistage build so I think this is resolved

.github/workflows/docker_publish.yml
shimwell: static_exe has been removed from the matrix so I think this is resolved

.github/workflows/docker_publish.yml
shimwell: The multistage approach now speeds up the run time of the action and uses prebuild docker image if available so I think this is resolved

.github/workflows/docker_publish.yml
shimwell: this has now been renamed to ```build-dependency-img``` so I think this is resolved

.github/workflows/docker_publish.yml
shimwell: QEMU has been removed from the yml file so I think this is resolved

.github/workflows/docker_publish.yml
shimwell: QEMU has been removed from the yml file so I think this is resolved

.github/workflows/docker_publish.yml
shimwell: pushing now taken care of by the multistage docker build action so I think this is resolved

.github/workflows/docker_publish.yml
shimwell: server-stage of multi stage build is set to moab so I think this is now resolved

CI/Dockerfile
shimwell: these lines have now been removed so I think this is resolved

shimwell:
I have gone through each comment and done the changes need to resolve them.

I have not clicked the resolve button just yet as I want to see if the [CI works over on my branch](https://github.com/shimwell/DAGMC/actions)

I shall report back once the CI finishes

CI/Dockerfile
```diff
@@ -179,9 +179,6 @@ RUN mkdir -p ${moab_build_dir}/build && \
                   -DBUILD_SHARED_LIBS=ON \
                   -DENABLE_FORTRAN=OFF \
                   -DENABLE_PYMOAB=ON && \
-    mkdir -p ${moab_build_dir}/build/pymoab/lib/python3.6/site-packages && \
-    PYTHONPATH=${moab_build_dir}/build/pymoab/lib/python3.6/site-packages:${PYTHONPATH} make -j${ci_jobs} && \
-    PYTHONPATH=${moab_build_dir}/build/pymoab/lib/python3.6/site-packages:${PYTHONPATH} make install && \
```
gonuke: This took out too many lines!  Need to keep the `make` lines:
```
make -j${ci_jobs} && \
make install && \
```

shimwell: Ah darn missed that. Will fix straight away

shimwell: just pushed a fix, CI is [rerunning](https://github.com/shimwell/DAGMC/actions/runs/4823792354)

shimwell: CI is passing again :tada:

shimwell: also it ran in just 23 mins. Nice and quick

CI/Dockerfile
```diff
@@ -1,9 +1,36 @@
...
+# build from repository root directory with command
+# docker build -t dagmc -f CI/Dockerfile .
+
+# Global ARGS set before the first build stage are accessable by all build stages
+ARG EMBREE_BRANCH='v3.6.1'
+ARG geant4_version=10.5.1
 ARG UBUNTU_VERSION=18.04
+ARG MOAB_BRANCH=5.3.0
+ARG double_down=OFF
+ARG static_exe=OFF
```
gonuke: Remove all reference to this `ARG`

CI/Dockerfile
```diff
@@ -33,63 +60,199 @@ RUN apt-get -y update; \
...
-ENV install_dir=/root/opt
+ENV build_dir=${build_dir}
+ENV install_dir=${install_dir}
 
 
 FROM base as external_deps
 
-#setting the COMPILER variable
-ARG COMPILER=gcc
-ENV COMPILER=${COMPILER}
```
gonuke, line 51: I think we got rid of this build arg, and have no way to indicate different compiler choices.  Our workflow still claims to do both `gcc` and `clang` but I do t think we do.  I think we end up doing `gcc` twice.  I was able to confirm this looking at the newest build.

.github/workflows/docker_publish.yml
```diff
@@ -182,56 +44,19 @@ jobs:
...
-            HDF5=${{ matrix.hdf5_versions }}
-            MOAB=${{ matrix.moab_versions }}
-          tags: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}:ci_testing
+          repository: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}
+          stages: base, external_deps, hdf5
+          server-stage: moab
+          quiet: false
+          tag-latest-on-default: false
+          dockerfile: CI/Dockerfile
+          build-args: OWNER=${{ github.repository_owner }}, TAG=ci_testing, UBUNTU_VERSION=${{ matrix.ubuntu_versions }}, COMPILER=${{ matrix.compiler }}, HDF5=${{ matrix.hdf5_versions }}, MOAB=${{ matrix.moab_versions }}
```
gonuke: We aren't using `TAG` or `COMPILER`. Either remove here or figure out if/how to use them. (See comment about `COMPILER` below)

.github/workflows/docker_publish.yml
```diff
@@ -182,56 +44,19 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
 
-      - name: Build and push Dockerfile_3_moab
-        uses: docker/build-push-action@v2
+      - name: Installing Dependencies in Docker image
+        uses: firehed/multistage-docker-build-action@v1
```
gonuke, line 47: @bquan0 - should this use `v1.7` for access to newest capabilities?

.github/workflows/docker_publish.yml
```diff
@@ -253,121 +78,48 @@ jobs:
...
-        with:
-          src: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-housekeeping:ci_testing
-          dst: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-housekeeping:stable
+      - name: Checkout repository
+        uses: actions/checkout@v3
 
-      - name: Push Image as latest img
-        uses: akhilerm/tag-push-action@v1.0.0
+      - name: Test DAGMC in Docker image
+        uses: firehed/multistage-docker-build-action@v1
```
gonuke, line 94: See above comment about version? ( @bquan0 )

.github/workflows/docker_publish.yml
```diff
@@ -253,121 +78,48 @@ jobs:
...
+          stages: dagmc
+          server-stage: dagmc_test
+          quiet: false
+          tag-latest-on-default: false
+          dockerfile: CI/Dockerfile
+          build-args: UBUNTU_VERSION=${{ matrix.ubuntu_versions }}, COMPILER=${{ matrix.compiler }}, HDF5=${{ matrix.hdf5_versions }}, MOAB=${{ matrix.moab_versions }}
 
       - name: Log in to the Container registry
-        uses: docker/login-action@v1
+        if: ${{ github.repository_owner == 'svalinn' }}
```
gonuke, line 131: Either make this a new job (preferred? As it was previously) or we don't need to login to GHCR again

.github/workflows/docker_publish.yml
```diff
@@ -96,84 +26,16 @@ jobs:
...
-          18.04,
-          20.04,
-          ]
-        compiler : [
-          gcc,
-          clang,
-          ]
-        hdf5_versions : [
-          1.10.4,
+         1.10.4,
```
gonuke, line 28: How does this interact with the `env` settings above? Do we need both?

CI/Dockerfile
```diff
@@ -1,9 +1,36 @@
+# Svalinn DAGMC Dockerfile
+
+# build from repository root directory with command
+# docker build -t dagmc -f CI/Dockerfile .
+
+# Global ARGS set before the first build stage are accessable by all build stages
+ARG EMBREE_BRANCH='v3.6.1'
+ARG geant4_version=10.5.1
```
gonuke: I think we merged an update for up to v10.7.4 (#803)

We'll probably need a new SHA below... which begs the question about whether we should make this a build-arg or not? (Or maybe we don't use the SHA anyway?)

CI/Dockerfile
```diff
@@ -33,63 +60,199 @@ RUN apt-get -y update; \
...
+    cd ${dagmc_build_dir} && \
+    cd build && \
+    cmake ../${dagmc} -DMOAB_DIR=${moab_install_dir} \
+                      -DBUILD_GEANT4=ON \
+                      -DGEANT4_DIR=${geant4_install_dir} \
+                      -DBUILD_CI_TESTS=ON \
+                      # reg tests disabled see issue https://github.com/svalinn/DAGMC/issues/858
+                      -DBUILD_MW_REG_TESTS=OFF \
+                      -DBUILD_STATIC_EXE=${static_exe} \
+                      -DBUILD_STATIC_LIBS=${static_exe} \
```
gonuke: Remove or hardcode to OFF?

gonuke (changes requested):
A bunch of final(?) cleanup....

CI/Dockerfile
shimwell: I would prefer explicit hard code to off, I changed it to be OFF but we can do something else if perfered

.github/workflows/docker_publish.yml
shimwell: I thought v1 pulls in any version starting with 1 and would therefore get v1.7 but not v2.0

CI/Dockerfile
shimwell: static_exe has been removed with the latest commit , I think this is now resolved

.github/workflows/docker_publish.yml
shimwell: I've removed all ```TAG``` ```OWNER``` and ```COMPILER``` build args, i think this is now resolved

.github/workflows/docker_publish.yml
shimwell: let us remove the ENV setting, I think that is ignored but the code, I think this is resolved

CI/Dockerfile
shimwell: I think these are now controlled by the ```ARG CXX``` and ```ARG CC``` which we have defaults for

the trouble I see he is that one needs to change both together.

if clang then change cc to clang and cxx to clang++
if gcc then change cc to gcc and cxx to g++

so if we add this to the build matrix it would not be every combination but rather two linked options. I think COMPILER use to switch both cc and cxx for us.

.github/workflows/docker_publish.yml
shimwell: I thought v1 pulls in any version starting with 1 and would therefore get v1.*.* but not v2.0

shimwell: latest is also an option

CI/Dockerfile
shimwell: change to 10.7.4 which appears to be the newest supported version

CI/Dockerfile
shimwell: perhaps want to passing ```compiler``` as a build arg and then have some logic to find ```cc``` and ```cxx``` from the ```compiler``` arg?

shimwell:
ok I have done the easy ones :smile: 

the ```compiler``` might require additional logic to get what the docker file needs (cc and cxx)

.github/workflows/docker_publish.yml
gonuke: I thought so too, but I've seen different behaviors with this action stuff...

gonuke: I've confirmed that the owner of this action has updated tags appropriately, so no change is needed

gonuke: I did see some behavior that shows we are not getting all the benefits of the new action.  Maybe we need to turn on the `parallel` flag... @bquan0 ?

.github/workflows/docker_publish.yml
gonuke: Marked resolved per above discussion

CI/Dockerfile
gonuke: Here's what the current system does in a script:
```

if [ "$COMPILER" == "gcc" ]; then
  export  CC=`which gcc`
  export CXX=`which g++`
  export ci_jobs="4"
elif [ "$COMPILER" == "clang" ]; then
  export  CC=`which clang`
  export CXX=`which clang++`
fi
```

gonuke:
We're nearly there!! Excited to see this finished. Hopefully we can look at the final things in my group's software meeting today

gonuke:
There is a PR to this PR branch that should resolve everything

shimwell:
I like the sound of that type of PR

shimwell:
I really like the solution you found for setting compiler and getting two envs set

gonuke:
I think [this is the evidence](https://github.com/shimwell/DAGMC/actions/runs/4840675258) that this is working!

I'm worried that I'm too complicit in the work to merge...  @pshriwise or maybe @bam241 would enjoy the elegance of it all!

pshriwise:
> I really like the solution you found for setting compiler and getting two envs set

Agreed, that's a really nice solution. Better than others I've seen in our google searches.

gonuke (approved):

gonuke (changes requested):
I think I found some new issues - don't merge yet

.github/workflows/docker_publish.yml
```diff
@@ -253,88 +77,42 @@ jobs:
...
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Push Image as stable img
-        uses: akhilerm/tag-push-action@v1.0.0
-        with:
-          src: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-housekeeping:ci_testing
-          dst: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-housekeeping:stable
+      - name: Checkout repository
+        uses: actions/checkout@v3
```
gonuke, line 91: Where does this checkout go?  We are running in the image we have built that uses `/root/build_dir` as the place where it builds everything. The dockerfile creates `/root/build_dir/dagmc/build` and then invokes `cmake` with directory `../${dagmc}` but I can't tell where `dagmc` is set??  So I can't tell where it is invoking `cmake`?

CI/Dockerfile
```diff
@@ -33,63 +54,207 @@ RUN apt-get -y update; \
...
+                      -DBUILD_CI_TESTS=ON \
+                      # reg tests disabled see issue https://github.com/svalinn/DAGMC/issues/858
+                      -DBUILD_MW_REG_TESTS=OFF \
+                      -DBUILD_STATIC_EXE=OFF \
+                      -DBUILD_STATIC_LIBS=OFF \
+                      -DCMAKE_C_COMPILER=${CC} \
+                      -DCMAKE_CXX_COMPILER=${CXX} \
+                      -DCMAKE_Fortran_COMPILER=gfortran \
+                      -DCMAKE_INSTALL_PREFIX=${dagmc_install_dir} \
+                      -DDOUBLE_DOWN=${double_down} \
```
gonuke: not sure where/if `double_down` is set?  I think it should always be `ON` given our current testing plan

gonuke (changes requested):
A couple of issues - one is very subtle...

.github/workflows/docker_publish.yml
gonuke: I'm thoroughly confused about how the dockerfile is even successful at building dagmc?!?! We never clone the repo from within the docker... what am I missing???

gonuke: I was missing something - it was clearly written in the comment!!!  The whole repo gets `COPY`ed into the docker image to the location in question, so it is now clear why/how this works. Sorry for the panic.

CI/Dockerfile
gonuke: This is set to `OFF` by default, but we always build double down anyway...  probably OK like this.

gonuke:
Another [PR](https://github.com/shimwell/DAGMC/pull/7) for some final cleanup

gonuke:
> Another [PR](https://github.com/shimwell/DAGMC/pull/7) for some final cleanup

Github actions for this PR are successful

shimwell:
Thanks Paul, I've merged that in.

gonuke:
Confirmation of success [here](https://github.com/shimwell/DAGMC/actions/runs/4846473875)

gonuke (approved):
I think this is in good shape and is a useful model for many projects going forward.  I'd love another review since I had a hand in authoring this in the end

.github/workflows/docker_publish.yml
```diff
@@ -253,88 +77,42 @@ jobs:
...
 
-      - name: Push Image as latest img
-        uses: akhilerm/tag-push-action@v1.0.0
+      - name: Test DAGMC in Docker image
+        uses: firehed/multistage-docker-build-action@v1
         with:
-          src: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-housekeeping:ci_testing
-          dst: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-housekeeping:latest
+          repository: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}
```
pshriwise:
```suggestion
          repository: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler }}-ext-hdf5_${{ matrix.hdf5_versions }}-moab_${{ matrix.moab_versions }}
```

.github/workflows/docker_publish.yml
```diff
@@ -347,27 +125,26 @@ jobs:
...
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Push Image as stable img
-        uses: akhilerm/tag-push-action@v1.0.0
+        if: ${{ github.repository_owner == 'svalinn' }}
+        uses: akhilerm/tag-push-action@v2.1.0
         with:
-          src: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}:ci_testing
+          src: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}/dagmc:refs_heads_${{ github.ref_name }}-bk0
```
pshriwise:
```suggestion
          src: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler }}-ext-hdf5_${{ matrix.hdf5_versions }}-moab_${{ matrix.moab_versions }}/dagmc:refs_heads_${{ github.ref_name }}-bk0
```

pshriwise: The line below could use some similar cleanup

.github/workflows/docker_publish.yml
```diff
@@ -347,27 +125,26 @@ jobs:
...
 
       - name: Push Image as latest img
-        uses: akhilerm/tag-push-action@v1.0.0
+        if: ${{ github.repository_owner == 'svalinn' }}
+        uses: akhilerm/tag-push-action@v2.1.0
         with:
-          src: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}:ci_testing
-          dst: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}:latest
\ No newline at end of file
+          src: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}/dagmc:refs_heads_${{ github.ref_name }}-bk0
```
pshriwise: Is the `bk0` suffix specific to something or arbitrary?

CI/Dockerfile
```diff
@@ -33,63 +54,207 @@ RUN apt-get -y update; \
...
-        /root/etc/CI/docker/build_moab.sh; \
-    fi;
-    
\ No newline at end of file
+
+RUN mkdir -p ${moab_build_dir}/build && \
+    cd ${moab_build_dir} && \
+    git clone -b ${MOAB_BRANCH} --depth 1 https://bitbucket.org/fathomteam/moab && \
+    cd build && \
+    cmake ../moab -DCMAKE_INSTALL_RPATH=${hdf5_install_dir}/lib:${moab_install_dir}/lib \
```
pshriwise, line 176: IIRC MOAB doesn't set the `CMAKE_BUILD_TYPE` by default. Adding `CMAKE_BUILD_TYPE=Release` might be worthwhile to ensure standard compiler optimization flags are included in the build.

CI/Dockerfile
```diff
@@ -33,63 +54,207 @@ RUN apt-get -y update; \
...
+                      # reg tests disabled see issue https://github.com/svalinn/DAGMC/issues/858
+                      -DBUILD_MW_REG_TESTS=OFF \
+                      -DBUILD_STATIC_EXE=OFF \
+                      -DBUILD_STATIC_LIBS=OFF \
+                      -DCMAKE_C_COMPILER=${CC} \
+                      -DCMAKE_CXX_COMPILER=${CXX} \
+                      -DCMAKE_Fortran_COMPILER=gfortran \
+                      -DCMAKE_INSTALL_PREFIX=${dagmc_install_dir} \
+                      -DDOUBLE_DOWN=${double_down} \
+                      -Ddd_ROOT=${double_down_install_dir} && \
```
pshriwise:
```suggestion
    cmake ../ -DMOAB_DIR=${moab_install_dir} \
              -DBUILD_GEANT4=ON \
              -DGEANT4_DIR=${geant4_install_dir} \
              -DBUILD_CI_TESTS=ON \
              # reg tests disabled see issue https://github.com/svalinn/DAGMC/issues/858
              -DBUILD_MW_REG_TESTS=OFF \
              -DBUILD_STATIC_EXE=OFF \
              -DBUILD_STATIC_LIBS=OFF \
              -DCMAKE_C_COMPILER=${CC} \
              -DCMAKE_CXX_COMPILER=${CXX} \
              -DCMAKE_Fortran_COMPILER=gfortran \
              -DCMAKE_INSTALL_PREFIX=${dagmc_install_dir} \
              -DDOUBLE_DOWN=${double_down} \
              -Ddd_ROOT=${double_down_install_dir} && \
```

pshriwise (changes requested):
Some small line comments from me, but this looks really nice! I was expecting a much larger changeset to accomplish this.

Q: Which of the stages gets uploaded in the end?

.github/workflows/docker_publish.yml
```diff
@@ -347,27 +125,26 @@ jobs:
...
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Push Image as stable img
-        uses: akhilerm/tag-push-action@v1.0.0
+        if: ${{ github.repository_owner == 'svalinn' }}
+        uses: akhilerm/tag-push-action@v2.1.0
         with:
-          src: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}:ci_testing
+          src: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}/dagmc:refs_heads_${{ github.ref_name }}-bk0
           dst: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}:stable
```
shimwell:
```suggestion
          dst: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-ext-hdf5_${{ matrix.hdf5_versions }}-moab_${{ matrix.moab_versions }}:stable
```
adding space as noticed by patrick

.github/workflows/docker_publish.yml
shimwell:
```suggestion
          src: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler }}-ext-hdf5_${{ matrix.hdf5_versions }}-moab_${{ matrix.moab_versions }}/dagmc:refs_heads_${{ github.ref_name }}-bk0
```

shimwell: not sure, but I see ```bk0``` appears in two places. perhaps @gonuke or @bquan0 know where that suffix came from

CI/Dockerfile
shimwell: Can I leave this one to you @gonuke

.github/workflows/docker_publish.yml
shimwell: I've now cleaned up two more cases where were were missing a space next to the brackets }}

.github/workflows/docker_publish.yml
gonuke: its generated by the multistage-build-action - not sure why the author uses that string.

CI/Dockerfile
gonuke: I think this will do it @shimwell 
```suggestion
    cmake ../moab -DCMAKE_INSTALL_RPATH=${hdf5_install_dir}/lib:${moab_install_dir}/lib \
                  -DCMAKE_BUILD_TYPE=Release \
```

gonuke:
> Q: Which of the stages gets uploaded in the end?

Every stage that is explicitly referenced in a `multistage-docker-build-action` is pushed to the repo:
- initially: `base`, `external_deps`, `hdf5`, `moab`
- during testing: `dagmc`, `dagmc_test`

They each get pushed with the custom tag and then we convert push the `dagmc` stage with tags `stable` and `latest` as well, but only when running from the `svalinn` repo.

CI/Dockerfile
shimwell: Thanks for this, that makes sense to me now. I've committed this and [now the CI is running](https://github.com/shimwell/DAGMC/actions/runs/4893025226)

shimwell: ok CI has passed so I think this is now resolved

pshriwise:
I'd love to approve and merge this, but there a runner isn't picking up the Mac testing job unfortunately.

gonuke:
Saw that...☹️

gonuke:
Github has turned off the MacOS 10.15 runners, so I made [this PR](https://github.com/shimwell/DAGMC/pull/8) to move us forward.

gonuke:
This is still passing here: https://github.com/shimwell/DAGMC/actions/runs/4904019862

pshriwise (approved):
Looking forward to utilizing these improvements in future PRs!! Thanks @gonuke and @shimwell  for all the detective work 🤩 📈

gonuke:
Thanks for the teamwork @shimwell @bquan0 & @pshriwise !

shimwell:
Delighted this one got in, a real team effort. Nice work all. What shall we do next 😁

pshriwise:
😢  https://github.com/svalinn/DAGMC/actions/runs/4907989513/jobs/8764666815

shimwell:
posting error message here so we don't lose it when the CI gets old

```
[ 55%] Building CXX object test/CMakeFiles/mbcn_test.dir/mbcn_test.cpp.o
Installing collected packages: pymoab
  Found existing installation: pymoab 5.4.1
    Can't uninstall 'pymoab'. No files were found to uninstall.
  Running setup.py develop for pymoab
    Complete output from command /usr/bin/python3 -c "import setuptools, tokenize;__file__='/root/build_dir/moab/bld/pymoab/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" develop --no-deps --prefix=/root/build_dir/moab/bld/pymoab:
    running develop
    error: can't create or remove files in install directory
    
    The following error occurred while trying to add or remove files in the
    installation directory:
    
        [Errno 2] No such file or directory: '/root/build_dir/moab/bld/pymoab/lib/python3.6/site-packages/test-easy-install-5760.write-test'
    
    The installation directory you specified (via --install-dir, --prefix, or
    the distutils default setting) was:
    
        /root/build_dir/moab/bld/pymoab/lib/python3.6/site-packages
    
    This directory does not currently exist.  Please create it and try again, or
    choose a different installation directory (using the -d or --install-dir
    option).
    
    
    ----------------------------------------
  Can't roll back pymoab; was not uninstalled
Command "/usr/bin/python3 -c "import setuptools, tokenize;__file__='/root/build_dir/moab/bld/pymoab/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" develop --no-deps --prefix=/root/build_dir/moab/bld/pymoab" failed with error code 1 in /root/build_dir/moab/bld/pymoab/
pymoab/CMakeFiles/pymoab-local-install.dir/build.make:58: recipe for target 'pymoab/CMakeFiles/pymoab-local-install' failed
make[2]: *** [pymoab/CMakeFiles/pymoab-local-install] Error 1
CMakeFiles/Makefile2:1238: recipe for target 'pymoab/CMakeFiles/pymoab-local-install.dir/all' failed
make[1]: *** [pymoab/CMakeFiles/pymoab-local-install.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 55%] Linking CXX executable ../bin/mbcn_test
[ 55%] Built target mbcn_test
make: *** [all] Error 2
Makefile:140: recipe for target 'all' failed
Error: Process completed with exit code 2.
```

shimwell:
```
e08954f49424: Pull complete
Digest: sha256:3c332407c190c017dbe049e8e9c9d54c5e44166b484c56fd414821bc54e5c233
Status: Downloaded newer image for akhilerm/repo-copy:latest
crane [copy ghcr.io/svalinn/dagmc-ci-ubuntu-22.04-gcc-ext-hdf5_1.10.4-moab_5.3.0/dagmc:refs_heads_develop-bk0 ghcr.io/svalinn/dagmc-ci-ubuntu-22.04-gcc-ext-hdf5_1.10.4-moab_5.3.0:stable]
2023/05/07 17:46:34 Copying from ghcr.io/svalinn/dagmc-ci-ubuntu-22.04-gcc-ext-hdf5_1.10.4-moab_5.3.0/dagmc:refs_heads_develop-bk0 to ghcr.io/svalinn/dagmc-ci-ubuntu-22.04-gcc-ext-hdf5_1.10.4-moab_5.3.0:stable
Error: fetching "ghcr.io/svalinn/dagmc-ci-ubuntu-22.04-gcc-ext-hdf5_1.10.4-moab_5.3.0/dagmc:refs_heads_develop-bk0": GET https://ghcr.io/v2/svalinn/dagmc-ci-ubuntu-22.04-gcc-ext-hdf5_1.10.4-moab_5.3.0/dagmc/manifests/refs_heads_develop-bk0: MANIFEST_UNKNOWN: manifest unknown
panic: exit status 1

goroutine 1 [running]:
main.main()
	/go/src/github.com/akhilerm/repo-copy/main.go:20 +0x15a
Error: The process '/usr/bin/docker' failed with exit code 2
```

gonuke:
Yep - working on that, too ☹️

Issue #856, opened 2023-02-16
Control GeomQueryTool screen output


aprilnovak:
In Cardinal, we are re-generating an OpenMC DAGMC model on-the-fly based on contours in temperature and density. Each time we re-init DAGMC, the following is printed to the screen

```
Set overlap thickness = 0
Set numerical precision = 0.001
Initializing the GeomQueryTool...
Using faceting tolerance: 0.0001
Building acceleration data structures...
Implicit Complement assumed to be Vacuum
```

In a given simulation, we might be re-initing DAGMC 15 times, so this repeated screen output can de-sensitize the user to our other screen output that we do want them to pay attention to. Ideally, this screen output could be optionally hidden.

gonuke:
We can introduce a convention of this kind of output and a mechanism for users to influence it.  We can pass that through to MOAB for the first two lines.

pshriwise:
Staring at this a little bit more after our call this morning. @gonuke would it be sufficient to add a `verbose` flag to the `GeomQueryTool` and `GeomTopoTool` classes in MOAB? We could then include a method in DAGMC to set the verbosity the classes it depends upon (as well as the verbosity of the `DagMC` class itself).

makeclean:
I thought we had one at some point? maybe I'm mis-remembering?

pshriwise:
Not seeing anything of that nature in the source code 😞 Shame on us...

gonuke:
> Staring at this a little bit more after our call this morning. @gonuke would it be sufficient to add a `verbose` flag to the `GeomQueryTool` and `GeomTopoTool` classes in MOAB? We could then include a method in DAGMC to set the verbosity the classes it depends upon (as well as the verbosity of the `DagMC` class itself).

I think that would do it, but is still fairly intrusive.  We also would have to act on that flag in DAGMC.

pshriwise:
Agreed, we'd have to act on that flag in DAGMC. And while some changes will be required in MOAB, there's a separate question of how we want to handle verbosity in the DagMC class. Here are the aforementioned lines annotated with which project they come from:

| Output Line | Class|
|------|----|
| "Set overlap thickness = 0" | MOAB::GeomQueryTool |
| "Set numerical precision = 0.001" | MOAB::GeomQueryTool |
| "Initializing the GeomQueryTool..." | DagMC |
| "Using faceting tolerance: 0.0001" | DagMC|
| "Building acceleration data structures..." |DagMC|
[...]

PR #872, opened 2023-03-21
Adding console output methods to some classes

pshriwise:
This is a step toward improved console logging in some of our classes. As described in #856, DAGMC and other classes in MOAB always write some output, which can conflict with output in downstream applications and cause confusion.

`message`, `warning`, and `err_msg` methods have been added to the `dagmcMetaData` and `DagMC` classes. Warnings and errors will always be written to screen. The `message` method allows a verbosity level to be associated with the message. The message will either be written to screen or not depending on the verbosity level set on the class instance. I've specified two levels of verbosity for now, but I'm open to a more fine grain set of verbosity levels down the line.

There's some duplicated code here, which could be improved. I'll look into abstracting these capabilieis into an object that both of the affected classes can inherit from perhaps. 

Another thing I don't love about this is that a `std::stringstream` object needs to be created for messages where conversion of non-char/string types occurs via the streaming operator. There are other libraries like `fmt` that handle this more gracefully, but I'll leave that for an improvement in the future.

### DIFF
--- a/.github/workflows/linux_build_test.yml
+++ b/.github/workflows/linux_build_test.yml
@@ -48,6 +48,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
+        with:
+          set-safe-directory: '*'
 
       - name: Setup
         run: |
--- a/.github/workflows/linux_build_test.yml
+++ b/.github/workflows/linux_build_test.yml
@@ -48,8 +48,6 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
-        with:
-          submodules: recursive
 
       - name: Setup
         run: |
--- a/src/dagmc/DagMC.hpp
+++ b/src/dagmc/DagMC.hpp
@@ -75,11 +75,10 @@ class DagMC {
   DagMC(std::shared_ptr<Interface> mb_impl = nullptr,
         double overlap_tolerance = 0., double numerical_precision = .001);
   // Deprecated Constructor
-  [
-      [deprecated("Replaced by DagMC(std::shared_ptr<Interface> mb_impl, ... "
-                  ")")]] DagMC(Interface* mb_impl,
-                               double overlap_tolerance = 0.,
-                               double numerical_precision = .001);
+  [[deprecated(
+      "Replaced by DagMC(std::shared_ptr<Interface> mb_impl, ... "
+      ")")]] DagMC(Interface* mb_impl, double overlap_tolerance = 0.,
+                   double numerical_precision = .001);
   // Destructor
   ~DagMC();
 
--- a/src/dagmc/DagMC.hpp
+++ b/src/dagmc/DagMC.hpp
@@ -466,6 +466,56 @@ class DagMC {
     return MBI_shared_ptr;
   }
 
+  // Verbosity levels
+  // -1 : "override" the current verbosity setting and write message
+  //  0 : No output
+  //  1 : All output
+  int set_verbosity(int val) {
+    int verbosity_max = 0;
+    int verbosity_min = 1;
+    if (val < verbosity_min || val > verbosity_max)
+      warning("Invalid verbosity value " + std::to_string(val) +
+              " will be set to nearest valid value.");
+    val = std::min(std::max(verbosity_min, val), verbosity_max);
+    return verbosity = val;
+  }
+
+  int get_verbosity(int val) { return verbosity; }
+
+  /** console output mechanism
+   *  @param msg Message to be written to the console
+   *  @param lvl Message will not be written if the verbosity level
+   *             is higher than the class instance's current verbosity setting.
+   *             Use of -1 ensures the message will always be written.
+   *  @param newline Whether or not to apend a newline to the message.
+   */
+  void message(const std::string& msg, int lvl = 1, bool newline = true) const {
+    // do not write message if level is higher than
+    // verbosity setting
+    if (lvl > verbosity) return;
+    // otherwise, display message
+    if (newline)
+      std::cout << msg << "\n";
+    else
+      std::cout << msg;
+  }
+
+  /** write a warning message to the console
+   *  @param msg Message to be written to the console
+   *  @param newline Whether or not to apend a newline to the message.
+   */
+  void warning(const std::string& msg, bool newline = true) const {
+    message("WARNING: " + msg, -1, newline);
+  }
+
+  /** write an error message to the console
+   *  @param msg Message to be written to the console
+   *  @param newline Whether or not to apend a newline to the message.
+   */
+  void err_msg(const std::string& msg, bool newline = true) const {
+    message("ERROR: " + msg, -1, newline);
+  }
+
  private:
   /* PRIVATE MEMBER DATA */
 
@@ -485,6 +535,9 @@ class DagMC {
 #endif
 
   std::unique_ptr<RayTracer> ray_tracer;
+
+  /** verbosity setting for DAGMC operations */
+  int verbosity{1};
 
  public:
   Tag nameTag, facetingTolTag;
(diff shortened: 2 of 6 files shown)


src/dagmc/DagMC.cpp
```diff
@@ -290,7 +288,7 @@ ErrorCode DagMC::remove_graveyard() {
   // get the implicit complement, it's children will need updating
   EntityHandle implicit_complement = 0;
   rval = geom_tool()->get_implicit_complement(implicit_complement);
-  if (rval != MB_ENTITY_NOT_FOUND || rval != MB_SUCCESS) {
```
gonuke: how did this ever work?

gonuke:
Did you mean this to be extended from #855 (which we should get back to)?

gonuke:
?

pshriwise:
> Did you mean this to be extended from #855 (which we should get back to)?

Ugh no. Accidentally combined some stashed changes I think. Fixing...

gonuke:
I've merged #855, so maybe it will be easy to just rebase on that?

gonuke:
If you rebase we can probably more forward on this @pshriwise

src/dagmc/DagMC.hpp
```diff
@@ -466,6 +466,56 @@ class DagMC {
...
+    int verbosity_max = 0;
+    int verbosity_min = 1;
+    if (val < verbosity_min || val > verbosity_max)
+      warning("Invalid verbosity value " + std::to_string(val) +
+              " will be set to nearest valid value.");
+    val = std::min(std::max(verbosity_min, val), verbosity_max);
+    return verbosity = val;
+  }
+
+  int get_verbosity(int val) { return verbosity; }
```
gonuke: Why does this take an argument- seems confusing

src/dagmc/DagMC.hpp
```diff
@@ -466,6 +466,56 @@ class DagMC {
...
+  //  0 : No output
+  //  1 : All output
+  int set_verbosity(int val) {
+    int verbosity_max = 0;
+    int verbosity_min = 1;
+    if (val < verbosity_min || val > verbosity_max)
+      warning("Invalid verbosity value " + std::to_string(val) +
+              " will be set to nearest valid value.");
+    val = std::min(std::max(verbosity_min, val), verbosity_max);
+    return verbosity = val;
```
gonuke: Unless my brain is missing something, this will always result in 0?

src/dagmc/DagMC.hpp
```diff
@@ -466,6 +466,56 @@ class DagMC {
...
+   *             Use of -1 ensures the message will always be written.
+   *  @param newline Whether or not to apend a newline to the message.
+   */
+  void message(const std::string& msg, int lvl = 1, bool newline = true) const {
+    // do not write message if level is higher than
+    // verbosity setting
+    if (lvl > verbosity) return;
+    // otherwise, display message
+    if (newline)
+      std::cout << msg << "\n";
```
gonuke, line 497:
```suggestion
      std::cout << msg << std::endl;
```

src/dagmc/dagmcmetadata.cpp
```diff
@@ -283,12 +284,14 @@ void dagmcMetaData::parse_importance_data() {
...
-                     "is allowed"
-                  << std::endl;
+        std::stringstream ss;
+        ss << "Volume with ID " << volid << " has more than one importance "
+           << std::endl;
+        ss << "Assigned for particle type " << pair.first << std::endl;
+        ss << "Only one importance value per volume per particle type "
+              "is allowed"
+           << std::endl;
+        message(ss.str(), -1, false);
```
gonuke:
```suggestion
              "is allowed";
        message(ss.str(), -1);
```

src/dagmc/dagmcmetadata.cpp
```diff
@@ -360,15 +363,16 @@ void dagmcMetaData::parse_boundary_data() {
...
+      ss << surfid << " has the following density assignments" << std::endl;
       for (int j = 0; j < boundary_assignment.size(); j++) {
-        std::cout << boundary_assignment[j] << std::endl;
+        ss << boundary_assignment[j] << std::endl;
       }
-      std::cout << "Please check your boundary condition assignments " << surfid
-                << std::endl;
+      ss << "Please check your boundary condition assignments " << surfid
+         << std::endl;
+      err_msg(ss.str());
```
gonuke:
```suggestion
      ss << "Please check your boundary condition assignments " << surfid;
      err_msg(ss.str());
```

gonuke (changes requested):
A few attempts to tidy... can't help but feel that there is some other external solution we can import

src/dagmc/DagMC.hpp
pshriwise: Mm yeah mentally bit-flipped the initializer values at the beginning of this function.

src/dagmc/DagMC.hpp
pshriwise: I think it's preferred to use a newline literal these days. Using `std::endl` will force the output buffer to flush and is more likely to cause garbled messages in concurrent applications, so I thought I'd start taking us in that direction.

pshriwise: Could do it all in one go in a separate changeset though.

pshriwise:
Thanks for the review @gonuke! Not sure why the Housekeeping/Linux tests are failing at this point. Something about git permissions in the CI image. Made an attempt at addressing it, but no luck yet. I'll try to revisit it later if I have time.

gonuke:
The failing the linux builds is something annoying that I was sure we had resolved many months ago.  Not sure why it's coming up again.

src/dagmc/DagMC.hpp
gonuke: I thought it was preferred because it is platform independent (e.g. Windows carriage returns...).  That's the only reason I suggested it.

gonuke:
I made a quick scan for logging libraries and didn't find anything compelling - all a little heavier weight than I'd like.

Do you think we could write a header-only small class that would do this and include it in the relevant places?

pshriwise:
We certainly could! I don't have the cycles to do so at the moment though

gonuke:
> We certainly could! I don't have the cycles to do so at the moment though

Let me see how I might adapt what you have done in that way

gonuke:
I think this was superseded by #876

.github/workflows/linux_build_test.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
         with:
-          submodules: recursive
+          set-safe-directory: '*'
```
ahnaf-tahmid-chowdhury, line 53: I hope this may fix the linux build test failure.

```suggestion

- name: Add exception for dubious ownership
  run: git config --global --add safe.directory /__w/DAGMC/DAGMC
  
```

ahnaf-tahmid-chowdhury (changes requested):

.github/workflows/linux_build_test.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       - name: Checkout repository
```
ahnaf-tahmid-chowdhury, line 49: I think this may fix the linux build test failure

```suggestion

      - name: Add exception for dubious ownership
        run: git config --global --add safe.directory /__w/DAGMC/DAGMC
        
      - name: Checkout repository
```

ahnaf-tahmid-chowdhury (changes requested):

.github/workflows/linux_build_test.yml
gonuke: Thanks, but I think this PR has been replaced by another ( #876 ), so will be close by @pshriwise soon.

pshriwise: Correct! Closing now.

pshriwise:
Closing in favor of #876. @ahnaf-tahmid-chowdhury, if the changes you've suggested above apply in that PR as well would you mind adding them there also?

PR #875, opened 2023-03-31
Use multistage docker build action in docker_publish.yml

bquan0:
## Description
I replaced most of the docker build actions with a [multistage build action](https://github.com/marketplace/actions/multi-stage-docker-build) that will build all the stages in one job. This is still a work in progress, and I haven't changed the housekeeping job or updated the image names for the jobs that run tests. 

## Motivation and Context
Using the multistage action was briefly mentioned [here](https://github.com/svalinn/DAGMC/pull/822). 
This PR is a proof of concept to see if the multistage build action can be used in the `docker_publish.yml` workflow and how it should be implemented. Let me know if I should branch off of the PR in the previous link since that one modifies the `Dockerfile` to include the install scripts. 

## Changes
This PR makes `docker_publish.yml` build the images with the multistage docker build action, which is simpler and uses caching of images to avoid building stages that have stayed the same since the last build. 

## Behavior
The current behavior is that there is a job for each stage of the `Dockerfile`, and each job is dependent on the previous job in the workflow. The new behavior is that there is one job that will build all the stages in the `Dockerfile` using the multistage docker build action.

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -18,14 +18,21 @@ jobs:
           18.04,
           20.04,
           ]
-    name: Ubuntu ${{ matrix.ubuntu_versions }} basics deps
+        compiler : [
+          gcc,
+          clang,
+          ]
+        hdf5_versions : [
+          1.10.4,
+        ]
+        moab_versions : [
+          5.3.0,
+        ]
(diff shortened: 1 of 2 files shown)


gonuke:
Since this got folded in to #863, perhaps you can withdraw/close this PR?

bquan0:
Closing this PR as requested.

PR #878, opened 2023-04-17
More stable package find instructions

ebknudsen:

## Description
Stabililize the search routines for MOAB and double-down

## Motivation and Context
- double-down installs its .cmake-files explicitly to a subdirectory dd. This PR adds that subdir to the search path explicitly. In cases cmake does not find them otherwise
- The MOAB search directory variable should be quoted. In some cases it can cause FindMOAB to fail. If it is installed to non-standard directory.

## Changes
Bug-fix

## Behavior
Old behaviour: can cause installation to fail.
New behaviour: more stable installation, less dependent on local cmake parameters.


gonuke:
Thanks @ebknudsen - this seems reasonable to me.  I suppose we should be testing with DD at some point in our process - perhaps at merge time, even if not during PR review.  Thoughts @pshriwise ?

pshriwise:
Yes, I agree. I thought we were at some point. Did that go away?

PR #882, opened 2023-06-16
explicitly include the std integer declarations

ebknudsen:

## Description
Explicitly include the definition of int32_t

## Motivation and Context
Fixes a compile blocking error  on some systems

## Changes
bugfix

## Behavior
On some systems depending on the pedanticness of the compiler, not including this header can cause the compilation to fail. Using gcc 13 on my system.

## Other Information
Experienced the issue with gcc-13


gonuke:
Thanks @ebknudsen

PR #893, opened 2023-08-06
add limitation to cython version for MOAB compatibility & more selectivity about when some tests are run

gonuke:
## Description
Force cython version to be less than v3.0.0. At the same, be slightly more selective with triggers for some tests.

## Motivation and Context
The versions of MOAB compatible with DAGMC are not compatible with cython 3+.  pip installing cython recently changed to be v3, so we need to restrict the version.

Also, we are currently running tests on all platforms even when only changing the workflow on another platform.

## Changes
bug fix for building dependencies

Limit when tests are run based on file paths.

### DIFF
--- a/.github/workflows/linux_build_test.yml
+++ b/.github/workflows/linux_build_test.yml
@@ -8,6 +8,8 @@ on:
       - develop
     paths-ignore:
       - '.github/workflows/docker_publish.yml'
+      - '.github/workflows/mac_build_test.yml'
+      - '.github/workflows/windows_build_test.yml'
       - '.github/workflows/housekeeping.yml'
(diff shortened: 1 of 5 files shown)


gonuke:
Test of docker image building are [here](https://github.com/gonuke/DAGMC/actions/runs/5778865788).

gonuke:
This is separated out from #880 to separate docker building steps from testing steps

gonuke:
This built successfully on my fork, and should be ready for review & merge.

The last merge to `develop` here failed to update the docker images correctly so I'll be on the lookout for that this time.

bquan0 (approved):
Looks good to me!

PR #894, opened 2023-08-13
Cleanup docker deploy

gonuke:
## Description
Remove dependence on unstable image tag for retagging.

## Motivation and Context
Recent merges of our new docker image deployment have failed at the last step because we rely on a tag generated by the multi-stage-build-action that is automatically generated.  The documentation for that project recommends that we do NOT rely on that tag.  We could probably figure out something that would work, but it may be fragile based on the guidance from the authors of that action.

## Changes
This uses logic tests to determine whether or not the images should be tagged as `latest` and only does so when it is a push to `develop` on the upstream `svalinn` repository.  Pushes to other branches will still run the tests, building the full set of docker images, but will not tag them as `latest`.

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -148,14 +148,14 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Push Image as latest img
-        if: ${{ github.repository_owner == 'svalinn' }}
+        if: ${{ github.repository_owner == 'svalinn' && github.ref == 'refs/heads/develop' }}
         uses: akhilerm/tag-push-action@v2.1.0
         with:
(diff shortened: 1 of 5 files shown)


gonuke:
Build testing including successful pushes [here](https://github.com/svalinn/DAGMC/actions/runs/5848377461).

gonuke:
✅ [Docker builds](https://github.com/svalinn/DAGMC/actions/runs/5848377461) are successful and other build/test is irrelevant (?) because it's still using old docker images.

Hopefully, merging this will allow us to move forward with proper build/test on linux using the correct docker images... but we'll need to review the final step carefully since it only happens on push to `svalinn/develop`

shimwell (approved):

shimwell:
Just the sort of thing we needed.

PR #895, opened 2023-08-15
Remove CI support scripts

gonuke:
## Description
Move all CI script content into Github actions

## Motivation and Context
Although it once seemed a good idea, to define all of our CI actions in bash scripts that were then launched from github actions, it became a challenge to maintain with too much cognitive redirection to different places.  

## Changes
This moves all the activity into the actions themselves, simply as bocks of bash commands.


shimwell:
wow this is a great tidy up. I shall review this week

gonuke:
Demonstration of the MOAB `master` test to be run only when a PR is merged is [here](https://github.com/gonuke/DAGMC/actions/runs/5879432542).

shimwell (approved):

PR #896, opened 2023-08-16
Fix docker args

gonuke:
## Description
Pass correctly named docker arguments to manage builds.

Also streamlines the docker build & test further taking more advantage of the multi-stage-build action.

## Motivation and Context
We currently build multiple docker images with a matrix of options.  The variable names to control the HDF5 and MOAB versions were not named correctly.  This did not create a flaw in our building/testing, because we only attempted a single version of each and those versions were the defaults anyway.  This PR corrects that.

## Changes
Rename the build args passed to Docker & combine two steps in the workflow.

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -29,7 +29,7 @@ jobs:
          5.3.0,
         ]
 
-    name: Installing Dependencies
+    name: Installing Dependencies, Building DAGMC and running tests
     steps:
       - name: default environment
         run: |
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -60,7 +60,7 @@ jobs:
           parallel: true
           tag-latest-on-default: ${{ env.tag-latest-on-default }}
           dockerfile: CI/Dockerfile
-          build-args: COMPILER=${{ matrix.compiler }}, UBUNTU_VERSION=${{ matrix.ubuntu_versions }}, HDF5=${{ matrix.hdf5_versions }}, MOAB=${{ matrix.moab_versions }}
+          build-args: COMPILER=${{ matrix.compiler }}, UBUNTU_VERSION=${{ matrix.ubuntu_versions }}, HDF5_VERSION=${{ matrix.hdf5_versions }}, MOAB_BRANCH=${{ matrix.moab_versions }}
 
   build-dagmc_test-img:
     needs: [build-dependency-img]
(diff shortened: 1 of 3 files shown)


gonuke:
Success of this build is demonstrated [here](https://github.com/gonuke/DAGMC/actions/runs/5879626238).

pshriwise (dismissed):
Thanks for these fixes @gonuke.

.github/workflows/docker_publish.yml
```diff
@@ -54,70 +54,16 @@ jobs:
...
-          repository: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler }}-ext-hdf5_${{ matrix.hdf5_versions }}-moab_${{ matrix.moab_versions }}
-          stages: moab, dagmc
+          stages: base, external_deps, hdf5, moab, dagmc
           server-stage: dagmc_test
           quiet: false
           parallel: true
           tag-latest-on-default: ${{ env.tag-latest-on-default }}
           dockerfile: CI/Dockerfile
-          build-args: COMPILER=${{ matrix.compiler }}, UBUNTU_VERSION=${{ matrix.ubuntu_versions }}, HDF5=${{ matrix.hdf5_versions }}, MOAB=${{ matrix.moab_versions }}
+          build-args: COMPILER=${{ matrix.compiler }}, UBUNTU_VERSION=${{ matrix.ubuntu_versions }}, HDF5_VERSION=${{ matrix.hdf5_versions }}, MOAB_BRANCH=${{ matrix.moab_versions }}
```
pshriwise, line 63: One quick thought after looking more closely at the Dockerfile -- do we need to set the `HDF5_VERSION_major` arg here as well? There's a note in the Dockerfile that it needs to be set manually.

gonuke: Wow! Great catch - I'll have to dig into that.  From a quick skim, I think we can automate that - will look tonight/tomorrow

gonuke:
New test of the docker building underway [here](https://github.com/gonuke/DAGMC/actions/runs/5891520446).

gonuke:
> New test of the docker building underway [here](https://github.com/gonuke/DAGMC/actions/runs/5891520446).

It appears to have been successful!

.github/workflows/docker_publish.yml
gonuke: Was able to make this work in latests commit

shimwell (approved):
This looks good to me, Paul has put in a nice python command to fix the hdf5 versions / major version error.

@pshriwise is this one ok to merge

pshriwise (approved):
Looks good! Thanks @gonuke!

PR #898, opened 2023-08-17
Add dd on merge

### LINKED ISSUE #897: test dagmc compile with double down in Docker file CI
Following up on the monthly meeting and it was mentioned that testing that DAGMC builds with double down enabled is not currently done with the CI.

We could consider adding double_down to the CI matrix or adding it as an include like this

```
        double_down : [
          OFF
        ]
        include:
          - ubuntu_versions: 22.04
            compiler: gcc
            npm: 1.10.4
            moab_versions: 5.3.0
            double_down: ON
```
And make use of the matrix value in the cmake command
```
-DDOUBLE_DOWN=${matrix.double_down} \
```

gonuke:
## Description
Add a test with Double Down on merge.

## Motivation and Context
We discovered after a recent PR was merged that it broke compatibility with Double Down.  For the time being, we have decided to only test that compatibility at merge and not with each PR.

## Changes
Adds a matrix option to build DAGMC using Double Down.  Also allows jobs to continue on error so that all options are tested even if one fails, since this is advisory post merge.

Also took the opportunity to collapse the docker builds in this workflow, and to exclude changes to this file from triggering other build/test actions.

Fixes #897

### DIFF
--- a/.github/workflows/linux_build_test.yml
+++ b/.github/workflows/linux_build_test.yml
@@ -10,6 +10,7 @@ on:
       - '.github/workflows/docker_publish.yml'
       - '.github/workflows/mac_build_test.yml'
       - '.github/workflows/windows_build_test.yml'
+      - '.github/workflows/linux_build_test_merge.yml'
       - '.github/workflows/housekeeping.yml'
       - '.github/workflows/changelog_test.yml'
       - 'CI/**'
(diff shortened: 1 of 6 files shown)


gonuke:
A successful demonstration of this test is [here](https://github.com/gonuke/DAGMC/actions/runs/5894530369), including the currently failing Double Down test.

pshriwise (approved):
This looks good to me @gonuke. Thanks for including DD in the test suite! I'm assuming that the removed block for tagging and uploading Docker images is taken care of elsewhere as a result of previous PRs.

gonuke:
> This looks good to me @gonuke. Thanks for including DD in the test suite! I'm assuming that the removed block for tagging and uploading Docker images is taken care of elsewhere as a result of previous PRs.

It's not clear whether we want to tag/push these docker images since their purpose was just to test against upstream dependencies.

pshriwise:
Makes sense. Thanks for clarifying!

PR #860, opened 2023-02-19
Updated DAGMC code to be compatible with Geant4 v11.1.1

ahnaf-tahmid-chowdhury:
### Description
In this pull request, I have updated the DAGMC code to be compatible with Geant4 v11.1.1. This involved making changes to `exampleN01.cc` , `ExN01Analysis.hh`, `ExN01DetectorConstruction.cc`, `ExN01UserScoreWriter.cc` and `generate_geant4` which previously were only compatible with Geant4 v10.5.

### Motivation and Context
This change is required because many users are currently using Geant4 v11.1.1 and need DAGMC to be compatible with it. The previous version of DAGMC was causing issues for users attempting to use it with this version of Geant4, as described in issue #857.

### Changes
This PR introduces a bug fix, updating the DAGMC code to be compatible with Geant4 v11.1.1.

### Behavior
Previously, DAGMC was only compatible with Geant4 v10.5. This pull request updates the code to be compatible with Geant4 v11.1.1, allowing users to use DAGMC with this newer version of Geant4.

### Other Information
I have tested these changes and verified that they work correctly. 

    ./dagsolid_unit_tests
    [==========] Running 16 tests from 1 test case.
    [----------] Global test environment set-up.
    [----------] 16 tests from DagSolidTest
    [ RUN      ] DagSolidTest.SetUp
    Set overlap thickness = 0
    Set numerical precision = 0.001
    Loading file test_geom.h5m
    Initializing the GeomQueryTool...
    Using faceting tolerance: 0.0001
    Building acceleration data structures...
    [       OK ] DagSolidTest.SetUp (92 ms)
    [ RUN      ] DagSolidTest.point_in_volume
    Set overlap thickness = 0
    Set numerical precision = 0.001
    Loading file test_geom.h5m
    Initializing the GeomQueryTool...
    Using faceting tolerance: 0.0001
    Building acceleration data structures...
    [       OK ] DagSolidTest.point_in_volume (34 ms)
    [ RUN      ] DagSolidTest.point_on_surface
    Set overlap thickness = 0
    Set numerical precision = 0.001
    Loading file test_geom.h5m
    Initializing the GeomQueryTool...
    Using faceting tolerance: 0.0001
    Building acceleration data structures...
    [       OK ] DagSolidTest.point_on_surface (35 ms)
    [ RUN      ] DagSolidTest.point_out_outside_tolerance
    Set overlap thickness = 0
    Set numerical precision = 0.001
    Loading file test_geom.h5m
[...]

PR #899, opened 2023-08-21
Add tests on merge for various upstream packages

gonuke:
## Description
Create a set of separate tests that are run when a PR is merged to identify problems with upstream packages.

## Motivation and Context
We rely on a number of upstream dependencies and currently pin their version to ensure success on our builds.  It is useful, however, to monitor whether or not we break compatibility with an upstream or optional package due to changes in our code base or their code base.

## Changes
Adds three new tests that only occur on merge.  Badges for these tests are at the bottom of the readme for a quick assessment.

Because we are adding the ability to test multiple versions of Geant, this changes to embed the Geant version in the image name.


gonuke:
Evidence of docker images being built correctly [here](https://github.com/gonuke/DAGMC/actions/runs/5921030727).

gonuke:
When merged, the following should be true:
* MOAB `master` on Ubuntu 22.04 should show success
* Double Down is being build in the docker images with v1.0.0 and will show failure
* Geant4 11.1.1 should show failure (pending updates from #860)

gonuke:
Once merged, we'll need to update the linux build tests for the new image names, in a separate PR.

gonuke:
Any chance that @shimwell or @pshriwise can give this a quick look over before tomorrow's call?

shimwell:
Just reading up on composite actions which are new to me but I have managed to looked over this in time for the call

pshriwise (approved):
Thanks again for the explanation during the community meeting today @gonuke! I like these composite actions. I'll likely refer to this PR when trying them in the future.

PR #904, opened 2023-09-06
Update for newer images with geant name

gonuke:
Following the update in #899 this changes the routine testing on linux to use the new images.


pshriwise (approved):
Thanks @gonuke!

PR #915, opened 2023-10-21
Change source of images for testing

### LINKED ISSUE #914: CI requires contributors to build/publish packages
**Describe the Bug**<br/>
In recent changes to the CI system, we rely on pre-built packages with dependencies.  The current scheme requires each contributor to have the full set of packages on their own fork, which may limit contributions.

**To Reproduce**<br/>
Create a pull request from a fork that has not already build the required packages.

**Expected Behavior**<br/>
Contributors should not have to already build the dependency packages on their own forks.

**Additional Context**<br/>
The current scheme is convenient for contributors who want to test changes that require or depend on changes to the dependencies because it does seek out packages on their own fork.   If a contributor would like to test a change that requires modified dependencies, they will need to develop their own workflow for that.

This issue was noticed nearly simultaneously in cyclus by @bennibbelink and in DAGMC by @ahnaf-tahmid-chowdhury

gonuke:
## Description
Changes the source of docker images for testing.

## Motivation and Context
Fixes #914.

## Changes
Improves collaborator workflow.

### DIFF
--- a/.github/workflows/linux_build_test.yml
+++ b/.github/workflows/linux_build_test.yml
@@ -58,7 +58,7 @@ jobs:
         ]
 
     container:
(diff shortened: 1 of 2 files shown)


bennibbelink (approved):
Looks good to me.

Thanks for catching this in DAGMC @gonuke

PR #922, opened 2023-11-30
Relax brew gcc

gonuke:
## Description
Remove GCC version pin for MacOS/homebrew

## Motivation and Context
The Github CI service no longer supports gcc@6 showing it a deprecated.

## Changes
Remove the pin on gcc and try the newest version.


gonuke:
Apparently gcc comes built-in and we don't need to explicitly install it, so I've taken it out of the `brew` command

gonuke:
I've added this small change to #924

Issue #916, opened 2023-10-29
Add CI soft test for newer/newest HDF5 version


gonuke:
Recently added GitHub actions allow testing against newer versions of dependencies than we typically support.  We don't yet have an implementation of this testing for newer versions of HDF5. Related tools are moving forward with the versions they support, so we should add testing ourselves.

There may be interactions between dependencies that need to be considered.

gonuke:
This should be relatively straightforward but we need to confirm which version(s) we want to try & the MOAB version relationship

gonuke:
ping @bquan0 to look at this

gonuke:
we might want version 1.14.x

PR #925, opened 2023-12-05
Add action to test newer hdf5 version

bquan0:
## Description
I made a copy of one of the `linux_upstream_test.yml` files and modified it slightly to use the composite action with an argument of `hdf5_version: 1.14` as specified in #916. I'm guessing this is what the issue is asking me to do? I'm not entirely sure because it wasn't specific. The issue also mentions that there are interactions between dependencies, which can be easily added with a matrix in the `linux_upstream_test_hdf5.yml` file. However, I don't know what interactions need to be tested, so I didn't add the matrix just yet. 

## Behavior
New behavior is that there will be one more action run when PRs are merged to `develop` that will test whether DAGMC works with version 1.14 of hdf5.

### DIFF
--- a/.github/workflows/linux_upstream_test_hdf5.yml
+++ b/.github/workflows/linux_upstream_test_hdf5.yml
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - develop
+      - newer-hdf5
     paths-ignore:
       - '.github/workflows/docker_publish.yml'
       - '.github/workflows/linux_build_test.yml'
--- a/.github/workflows/linux_upstream_test_hdf5.yml
+++ b/.github/workflows/linux_upstream_test_hdf5.yml
@@ -28,4 +28,4 @@ jobs:
         uses: ./.github/actions/upstream-test
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          hdf5_version: 1.14.0
+          hdf5_version: 1.14.3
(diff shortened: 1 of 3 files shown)


gonuke:
Are you able to manually launch this action on your fork?

Also, please look at the bottom of the README to add a badge for this, as well.

.github/workflows/linux_upstream_test_hdf5.yml
```diff
@@ -0,0 +1,31 @@
...
+    
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: use upstream test composite action
+        uses: ./.github/actions/upstream-test
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          hdf5_version: 1.14.0
```
gonuke: Should we go straight to 1.14.3 instead?

bquan0:
Looks like [the workflow](https://github.com/bquan0/DAGMC/actions/runs/7130999484/job/19418631134) for `hdf5_version: 1.14.0` worked. I've just edited it to be `1.14.3`, so we'll see if that works too. 
I also mapped the hdf5 workflow badge to my repo and branch just to see if it would work. That's something I'll have to change before this PR gets merged.

gonuke:
Thanks @bquan0 - I think this looks OK - go ahead and switch back the README and I'll merge.

gonuke (approved):
Thanks @bquan0

PR #928, opened 2023-12-08
Double Down ci fix

ahnaf-tahmid-chowdhury:
Double Down v1.0.0 in not compatible with new DAGMC. Thus setting develop branch for testing. Suggesting to merge the PR after new Double Down version release.

PR #913, opened 2023-10-16
Updates to allow a release

gonuke:
These changes primarily update version strings for a release and update the pyne hash.

PR #929, opened 2023-12-09
Allow Double Down v1.1.0 Installation in Dockerfile

ahnaf-tahmid-chowdhury:
# Pull Request:
 
This is the upstream version of #928 

## Changes Made:

- Enabled Double Down v1.1.0 to be installed in the Dockerfile.
- Removed `double_down` from the action and added `double_down_version`. In the past, the build process used `double_down=ON/OFF` to build DAGMC, but now DAGMC will default to using Double Down. The version can be specified using `double_down_version` in the actions.

## Details:

- The Dockerfile has been updated to support the installation of Double Down v1.1.0.
- Removed the `double_down` variable from the action and introduced `double_down_version` for specifying the version during the actions.
- Previously, DAGMC was built using `double_down=ON/OFF`, but now it defaults to using Double Down.
- Various variables have been adjusted for consistency and improved readability.

This PR streamlines the installation process, makes the Dockerfile compatible with Double Down v1.1.0, and provides flexibility in choosing the version in the actions.


ahnaf-tahmid-chowdhury:
Not sure why Linux Build/Test is failing. It should use the latest docker image.

.github/workflows/linux_upstream_test_double_down.yml
```diff
@@ -1,4 +1,4 @@
-name: Test against Double Down on PR merge
+name: Test against newer Double Down on PR merge
```
pshriwise:
```suggestion
name: Test against Double Down development branch on PR merge
```

.github/workflows/linux_upstream_test_moab.yml
```diff
@@ -1,4 +1,4 @@
-name: Test against MOAB master on PR merge
+name: Test against newer MOAB on PR merge
```
pshriwise: I think the original version is more specific, but maybe to combine the two:

```suggestion
name: Test against MOAB development branch (master) on PR merge
```

pshriwise (changes requested):
A couple of really tiny nit picks here @ahnaf-tahmid-chowdhury. Thanks for updating the CI variables/Docker build args to use a consistent syntax. I think that's a nice improvement. 

I'll let @gonuke chime in on using double-down by default. My take is that we'd want to leave the default set to MOAB since double-down is an optional dependency of DAGMC.

CI/Dockerfile
```diff
@@ -26,31 +26,29 @@ SHELL ["/bin/bash", "-c"]
...
-                       python3-setuptools \
-                       python3-dev \
-                       libpython3-dev \
-                       wget \
-                       software-properties-common; \
-    add-apt-repository ppa:git-core/ppa; \
-    apt-get update; \
-    apt-get install -y git; \
-    update-alternatives --install /usr/bin/python python /usr/bin/python3 10; \
-    update-alternatives --install /usr/bin/pip pip /usr/bin/pip3 10; \
```
gonuke, line 52: I prefer this approach to the approach of always specifying `python3`.  I think it is more robust.

CI/Dockerfile
```diff
@@ -26,31 +26,29 @@ SHELL ["/bin/bash", "-c"]
...
+    libhdf5-dev \
+    libtool \
+    libeigen3-dev\
+    python3-numpy \
+    python3 \
+    python3-pip \
+    python3-setuptools \
+    python3-dev \
+    libpython3-dev \
+    software-properties-common && \
```
gonuke: I prefer these to be indented a little more to make it clear that they are arguments to `apt-get` and not additional commands to be RUN

gonuke (changes requested):
A couple of additional changes requested.

gonuke:
I'm happy to revisit the plan for testing with Double Down.  As @pshriwise says above, it is an optional dependency.  Therefore, we should definitely be testing without it.  What is not clear is whether we want to require that every change builds successfully with DD.  We could, but perhaps not the full matrix - just add it as one or two special cases.

ahnaf-tahmid-chowdhury:
Hi @gonuke and @pshriwise,

It seems that both Double Down and Geant4 are now optional for DAGMC. I have made some changes in this version. If the `geant4_version` or `double_down_version` is not set, workflow will automatically disable their build. Additionally, Docker images will be created with their names appended to the version string.

Since I've made several changes, please review from the beginning.

pshriwise:
Seems we'll need this to fix this build error: https://github.com/svalinn/DAGMC/actions/runs/7250886790/job/19752059151

@ahnaf-tahmid-chowdhury we should also require a minimum double-down version of 1.1.0 in the CMake `find_package` command. Would you mind adding that requirement?

ahnaf-tahmid-chowdhury:
> Seems we'll need this to fix this build error: https://github.com/svalinn/DAGMC/actions/runs/7250886790/job/19752059151

The workflow seems to be old. Where DD is set as ON and using v1.0.0. In my version I am using a specific version 1.1.0. Or should i make a method to use the latest tag version when we set the value ON?

pshriwise:
> The workflow seems to be old. Where DD is set as ON and using v1.0.0. In my version I am using a specific version 1.1.0. Or should i make a method to use the latest tag version when we set the value ON?

that workflow is older than the changes in this PR, yes. It's from merging #913

cmake/DAGMC_macros.cmake
```diff
@@ -89,7 +89,7 @@ macro (dagmc_setup_options)
   endif()
 
 if (DOUBLE_DOWN)
-  find_package(DOUBLE_DOWN REQUIRED)
+  find_package(DOUBLE_DOWN 1.1.0 REQUIRED)
```
pshriwise:
```suggestion
  find_package(DOUBLE_DOWN REQUIRED)
   if (${DOUBLE_DOWN_VERSION} VERSION_LESS 1.1.0)
    message(FATAL_ERROR "Discovered Double Down Version: ${DOUBLE_DOWN_VERSION}. \
    Please update Double Down to version 1.1.0 or greater.")
  endif()
```

gonuke:
I think this PR is mixing a number of things so that it's hard to review.  You have changed a lot of variable names that clutter and bury the substantive changes you've made in the workflow logic.  Could you submit a separate PR that cleans up the variable names and then we can review the changes in the DD options by themselves?

ahnaf-tahmid-chowdhury:
In this PR I haven't only added dd version, but also add a method to make a docker image for each version of dd as geant4 do. Also, I have created a method that allows not to create docker image when we set dd version to OFF. Furthermore, I have made the same logic for geant4 as it seems it is also a dependency for DAGMC.

Docker images will be created as

```
docker-ci-[Ubuntu with version]-[g++/clang]-[hdf5 with version]-[moab with version] # Geant4=OFF DD=OFF
docker-ci-[Ubuntu with version]-[g++/clang]-[hdf5 with version]-[moab with version]-[geant4 with version] # DD=OFF
docker-ci-[Ubuntu with version]-[g++/clang]-[hdf5 with version]-[moab with version]-[dd with version] # Geant4=OFF
docker-ci-[Ubuntu with version]-[g++/clang]-[hdf5 with version]-[moab with version]-[geant4 with version]-[dd with version]
```

I have also made a method to set build automatically ON for Geant4 and DD if the version is declaired.

.github/actions/upstream-test/action.yml
```diff
@@ -40,14 +40,34 @@ runs:
...
-          repository: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ inputs.ubuntu_version }}-${{ inputs.compiler}}-geant4_${{ inputs.geant_version }}-hdf5_${{ inputs.hdf5_version }}-moab_${{ inputs.moab_version }}
+          repository: ${{ env.image_base_tag }}
           stages: base, external_deps, hdf5, moab, dagmc
           server-stage: dagmc_test
           quiet: false
           parallel: true
           tag-latest-on-default: false
           dockerfile: CI/Dockerfile
-          build-args: COMPILER=${{ inputs.compiler }}, geant4_version=${{ inputs.geant_version }}, UBUNTU_VERSION=${{ inputs.ubuntu_version }}, HDF5_VERSION=${{ inputs.hdf5_version }}, MOAB_BRANCH=${{ inputs.moab_version }}, double_down=${{ inputs.double_down }}
+          build-args: UBUNTU_VERSION=${{ inputs.ubuntu_version }}, COMPILER=${{ inputs.compiler }}, HDF5_VERSION=${{ inputs.hdf5_version }}, MOAB_VERSION=${{ inputs.moab_version }}, GEANT4=${{ env.GEANT4 }}, GEANT4_VERSION=${{ inputs.geant4_version }}, DOUBLE_DOWN=${{ env.DOUBLE_DOWN }}, DOUBLE_DOWN_VERSION=${{ inputs.double_down_version }}
```
gonuke: If it is allowed (I think it is), maybe it's best to list these arguments one per line so that future changes can be more easily identified

.github/workflows/docker_publish.yml
```diff
@@ -57,17 +57,36 @@ jobs:
...
+            GEANT4=ON
+          else
+            GEANT4=OFF
+          fi
+          if [ ${{ matrix.double_down_version }} != "OFF" ] ; then
+            image_base_tag=${image_base_tag}-double_down_${{ matrix.double_down_version }}
+            DOUBLE_DOWN=ON
+          else
+            DOUBLE_DOWN=OFF
+          fi
```
gonuke: See above comments about always tagging with `OFF` or the version number

.github/workflows/docker_publish.yml
```diff
@@ -57,17 +57,36 @@ jobs:
...
-          repository: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-geant4_${{ matrix.geant_version }}-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}
+          repository: ${{ env.image_base_tag }}
           stages: base, external_deps, hdf5, moab, dagmc
           server-stage: dagmc_test
           quiet: false
           parallel: true
           tag-latest-on-default: ${{ env.tag-latest-on-default }}
           dockerfile: CI/Dockerfile
-          build-args: double_down=${{ matrix.double_down}}, geant4_version=${{ matrix.geant_version }}, COMPILER=${{ matrix.compiler }}, UBUNTU_VERSION=${{ matrix.ubuntu_versions }}, HDF5_VERSION=${{ matrix.hdf5_versions }}, MOAB_BRANCH=${{ matrix.moab_versions }}
+          build-args: UBUNTU_VERSION=${{ matrix.ubuntu_version }}, COMPILER=${{ matrix.compiler }}, HDF5_VERSION=${{ matrix.hdf5_version }}, MOAB_VERSION=${{ matrix.moab_version }}, GEANT4=${{ env.GEANT4 }}, GEANT4_VERSION=${{ matrix.geant4_version }}, DOUBLE_DOWN=${{ env.DOUBLE_DOWN }}, DOUBLE_DOWN_VERSION=${{ matrix.double_down_version}}
```
gonuke: one arg per line if possible

.github/workflows/docker_publish.yml
```diff
@@ -108,10 +127,17 @@ jobs:
...
         if: ${{ github.repository_owner == 'svalinn' }}
         run: |
-          echo "image_base_tag=ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler }}-geant4_${{ matrix.geant_version }}-hdf5_${{ matrix.hdf5_versions }}-moab_${{ matrix.moab_versions }}" >> "$GITHUB_ENV"
+          image_base_tag=ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_version }}-${{ matrix.compiler}}-hdf5_${{ matrix.hdf5_version}}-moab_${{ matrix.moab_version }}
+          if [ ${{ matrix.geant4_version }} != "OFF" ] ; then
+            image_base_tag=${image_base_tag}-geant4_${{ matrix.geant4_version }}
+          fi
+          if [ ${{ matrix.double_down_version }} != "OFF" ] ; then
+            image_base_tag=${image_base_tag}-double_down_${{ matrix.double_down_version }}
+          fi
```
gonuke: update to conform with comments above

.github/actions/upstream-test/action.yml
```diff
@@ -40,14 +40,34 @@ runs:
...
+          fi
+          if [ ${{ inputs.double_down_version }} != "OFF" ] ; then
+            image_base_tag=${image_base_tag}-double_down_${{ inputs.double_down_version }}
+            DOUBLE_DOWN=ON
+          else
+            DOUBLE_DOWN=OFF
+          fi
+          echo "image_base_tag=${image_base_tag}" >> $GITHUB_ENV
+          echo "GEANT4=${GEANT4}" >> $GITHUB_ENV
+          echo "DOUBLE_DOWN=${DOUBLE_DOWN}" >> $GITHUB_ENV
```
gonuke: I think we should always tag the image with `OFF` if it's off or the version number if it's on.  This will remove the need for all of this.

gonuke: We can also then use the logic of the Dockerfile to handle the ON vs OFF case

.github/workflows/linux_build_test.yml
```diff
@@ -36,30 +36,30 @@ jobs:
...
           10.7.4,
           11.1.2
         ]
+        double_down_version : [
+          OFF
+        ]
 
     container:
-      image: ghcr.io/svalinn/dagmc-ci-ubuntu-${{ matrix.ubuntu_versions }}-${{ matrix.compiler}}-geant4_${{ matrix.geant_version }}-hdf5_${{ matrix.hdf5_versions}}-moab_${{ matrix.moab_versions }}/moab:latest
+      image: ghcr.io/svalinn/dagmc-ci-ubuntu-${{ matrix.ubuntu_version }}-${{ matrix.compiler}}-hdf5_${{ matrix.hdf5_version}}-moab_${{ matrix.moab_version }}${{ matrix.geant4_version != 'OFF' && '-geant4_' || '' }}${{ matrix.geant4_version != 'OFF' && matrix.geant4_version || '' }}${{ matrix.double_down_version != 'OFF' && '-double_down_' || '' }}${{ matrix.double_down_version != 'OFF' && matrix.double_down_version || '' }}/moab:latest
```
gonuke: update per above

CI/Dockerfile
```diff
@@ -71,55 +73,39 @@ ENV CXX=clang++
...
-    cmake ../embree -DCMAKE_INSTALL_PREFIX=${embree_install_dir} \
-                    -DEMBREE_TASKING_SYSTEM=INTERNAL \
-                    -DEMBREE_ISPC_SUPPORT=OFF \
-                    -DEMBREE_TUTORIALS=OFF \
-                    -DEMBREE_TBB_ROOT=/usr && \
-    make -j${ci_jobs} && \
-    make -j${ci_jobs} install && \
-    cd && \
-    rm -rf ${embree_build_dir}
+RUN if [ "${GEANT4}" = "ON" ]; then \
```
gonuke: This can be changed to 
```suggestion
RUN if [ "${GEANT4_VERSION}" != "OFF" ]; then \
```
with above recommendations

CI/Dockerfile
```diff
@@ -137,15 +123,15 @@ ENV hdf5_install_dir=${install_dir}/hdf5
 
 RUN mkdir -p ${hdf5_build_dir}/build && \
     cd ${hdf5_build_dir} && \
-    export HDF5_VERSION_major=`python -c "print('.'.join('${HDF5_VERSION}'.split('.')[:-1]))"` && \
+    export HDF5_VERSION_major=`python3 -c "print('.'.join('${HDF5_VERSION}'.split('.')[:-1]))"` && \
```
gonuke: If we set the python version alternative correctly, we can just refer to `python`

CI/Dockerfile
```diff
@@ -169,47 +157,76 @@ ENV moab_install_dir=${install_dir}/moab
...
     make install && \
     cd && \
     rm -rf ${moab_build_dir}
 
+ENV embree_install_dir=${install_dir}/embree
+ENV embree_build_dir=${build_dir}/embree
+
+# Clone and install Embree
+# Required for Double-Down
+RUN if [ "${DOUBLE_DOWN}" = "ON" ]; then \
```
gonuke: With previous comments this becomes:
```suggestion
RUN if [ "${DOUBLE_DOWN_VERSION}" != "OFF" ]; then \
```

CI/Dockerfile
```diff
@@ -169,47 +157,76 @@ ENV moab_install_dir=${install_dir}/moab
...
-    cd ${double_down_build_dir} && \
-    git clone --shallow-submodules --single-branch --branch v1.0.0 --depth 1 https://github.com/pshriwise/double-down.git && \
-    cd build && \
-    cmake ../double-down -DMOAB_DIR=${moab_install_dir} \
-                         -DCMAKE_INSTALL_PREFIX=${double_down_install_dir} \
-                         -DEMBREE_DIR=${embree_install_dir} && \
-    make -j${ci_jobs} && \
-    make -j${ci_jobs} install
+# Performed after moab and embree install as double-down requires moab and embree
+RUN if [ "${DOUBLE_DOWN}" = "ON" ]; then \
```
gonuke: As above:
```suggestion
RUN if [ "${DOUBLE_DOWN_VERSION}" != "OFF" ]; then \
```

CI/Dockerfile
```diff
@@ -222,21 +239,22 @@ RUN cd  ${dagmc_build_dir} && git submodule update --init
...
+        -DBUILD_CI_TESTS=ON \
+        # reg tests disabled see issue https://github.com/svalinn/DAGMC/issues/858
+        -DBUILD_MW_REG_TESTS=OFF \
+        -DBUILD_STATIC_EXE=OFF \
+        -DBUILD_STATIC_LIBS=OFF \
+        -DCMAKE_C_COMPILER=${CC} \
+        -DCMAKE_CXX_COMPILER=${CXX} \
+        -DCMAKE_Fortran_COMPILER=gfortran \
+        -DCMAKE_INSTALL_PREFIX=${dagmc_install_dir} \
+        -DDOUBLE_DOWN=${DOUBLE_DOWN} \
```
gonuke: This will need to be managed to get the correct `ON` vs `OFF` here.

cmake/DAGMC_macros.cmake
```diff
@@ -90,6 +90,10 @@ macro (dagmc_setup_options)
 
 if (DOUBLE_DOWN)
   find_package(DOUBLE_DOWN REQUIRED)
```
gonuke, line 92: Can you put the required version in the `find_package` command and let CMake take care of the error?

gonuke (changes requested):
In addition to the changes requested here, this appears to need a rebase on recent changes.  Last time you did that it created a bit of a mess in your PR, so let me know if you want to talk it through.

.github/actions/upstream-test/action.yml
ahnaf-tahmid-chowdhury: I will try

.github/workflows/docker_publish.yml
ahnaf-tahmid-chowdhury: This block checks if the `matrix.geant4_version` is not OFF. If true, it appends the Geant4 version to the `image_base_tag` and sets the GEANT4 variable to ON; otherwise, it sets GEANT4 to OFF. This variable will let install Geant4 in the docker image and install DAGMC with Geant4 support later. Same for Double Down.

.github/workflows/docker_publish.yml
ahnaf-tahmid-chowdhury: You mean to add some documentations there how that works?

.github/actions/upstream-test/action.yml
ahnaf-tahmid-chowdhury: You are saying instead of setting ON/OFF here, we should make this method in the Docker file? This ON value is required for building DAGMC with Geant4 and DD. That means we need these conditions sooner or later.

cmake/DAGMC_macros.cmake
ahnaf-tahmid-chowdhury: I have done this in my previous commit. @pshriwise has come up with this idea. This will terminate the process if the minimum conditions aren't met with a custom message.

.github/workflows/docker_publish.yml
gonuke: I understand what it's doing, but I don't think it's necessary.  We can overload the purpose of the `geant4_version` variable and assign it a value of `OFF` or a value of the version in question.  We can then check that in the Dockerfile to either skip building it or building the requested version.  It creates fewer opportunities for these two variables to break inconsistent.

.github/workflows/docker_publish.yml
gonuke: I mean to use a single variable to either turn double down off or set the version.

cmake/DAGMC_macros.cmake
gonuke: @pshriwise is there a reason to not just let CMake take care of the error?

.github/actions/upstream-test/action.yml
```diff
@@ -40,14 +40,32 @@ runs:
...
+        shell: bash -l {0}
+        run: |
+          image_base_tag=ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ inputs.ubuntu_version }}-${{ inputs.compiler}}-hdf5_${{ inputs.hdf5_version}}-moab_${{ inputs.moab_version }}
+          if [ ${{ inputs.geant4_version }} != "OFF" ] ; then
+            image_base_tag=${image_base_tag}-geant4_${{ inputs.geant4_version }}
+          fi
+          if [ ${{ inputs.double_down_version }} != "OFF" ] ; then
+            image_base_tag=${image_base_tag}-double_down_${{ inputs.double_down_version }}
+          fi
+          echo "image_base_tag=${image_base_tag}" >> $GITHUB_ENV
```
gonuke: Please always include the geant and DD state in the image tag.  We then don't need this logic since we can just define it in the tag below as we have done in the past.

.github/workflows/docker_publish.yml
```diff
@@ -57,17 +57,34 @@ jobs:
...
+        run: |
+          image_base_tag=ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ matrix.ubuntu_version }}-${{ matrix.compiler}}-hdf5_${{ matrix.hdf5_version}}-moab_${{ matrix.moab_version }}
+          if [ ${{ matrix.geant4_version }} != "OFF" ] ; then
+            image_base_tag=${image_base_tag}-geant4_${{ matrix.geant4_version }}
+          fi
+          if [ ${{ matrix.double_down_version }} != "OFF" ] ; then
+            image_base_tag=${image_base_tag}-double_down_${{ matrix.double_down_version }}
+          fi
+          echo "image_base_tag=${image_base_tag}" >> "$GITHUB_ENV"
+
```
gonuke: Once again, if the geant and DD state is always in the tag name, we don't need this logic

.github/workflows/docker_publish.yml
gonuke: Once again, we don't need the logic here if we always include geant and DD in the tag name

.github/workflows/linux_build_test.yml
```diff
@@ -36,30 +36,40 @@ jobs:
...
+      image: > 
+        ghcr.io/svalinn/dagmc-ci
+        -ubuntu-${{ matrix.ubuntu_version }}
+        -${{ matrix.compiler}}
+        -hdf5_${{ matrix.hdf5_version}}
+        -moab_${{ matrix.moab_version }}
+        ${{ matrix.geant4_version != 'OFF' && '-geant4_' || '' }}
+        ${{ matrix.geant4_version != 'OFF' && matrix.geant4_version || '' }}
+        ${{ matrix.double_down_version != 'OFF' && '-double_down_' || '' }}
+        ${{ matrix.double_down_version != 'OFF' && matrix.double_down_version || '' }}
```
gonuke: Don't need this logic if we have geant and DD state always in the image tag

.github/workflows/linux_build_test.yml
gonuke, line 1: We'll probably need to update this file in a separate PR to make sure the docker files are available.

cmake/DAGMC_macros.cmake
gonuke: I see... I wonder why @pshriwise doesn't want to just rely on CMake's error message?

gonuke (changes requested):
Still some unnecessary logic assuming we always encode the Geant and DD state in the tag name, even if `OFF`

.github/actions/upstream-test/action.yml
ahnaf-tahmid-chowdhury: But there is an issue. We can't use  capital letters in the image tag string. It is invalid for docker url. So, I need to change OFF variable. Any alternative suggestion?

ahnaf-tahmid-chowdhury: I am setting this variable as `no`. `OFF=>no`

gonuke: I see - I forgot about that.  Could we use `off` instead of `no`?  Will CMake allow `off` to be treated like `OFF`?

ahnaf-tahmid-chowdhury: Actually, I think we can't use `off` in CMake for `OFF`. However, we can set any string here as we wish here. There are some conditions that only check these strings. But we can't set empty string in workflow matrix. So, let's go with `off`.

gonuke:
I've just noticed a couple of concepts that have been forgotten in this update.

Even though GEANT4 is an optional dependency, we always want to test with GEANT4 in CI.  There is no reason not to.  It should never be `off`.  It also leads to needing fewer Docker images in our container registry.

Also, Embree was always installed as part of the external dependencies for a few reasons, and independent of whether or not Double Down is being tested.  Again, since this is for CI, there is reason not to have Embree always installed.  If always installing it, it is best to do it early in the build since it changes least often and then the Docker cache doesn't have to rebuild it if/when MOAB changes.

It is easy to forget these things or not realize why they are true, so I suppose I should leave more comments for some of these subtle decisions.

ahnaf-tahmid-chowdhury:
If it is necessary to build Geant4 every time, we can keep this always on. Moreover, with this update we can use both mode with matrix. Same rule for Double Down. 

If Embree is always required to be installed (optional part of DAGMC) we can make remove the current condisions, also, we can make similar method for Embree as we have done for Geant4 and Double Down.

gonuke:
I think it makes sense to always build GEANT4, Embree & Double Down, but we don't always need to test with them.  For CI purposes, there is no down side to having the maximum number of dependencies installed, and then use them in different ways during testing.

ahnaf-tahmid-chowdhury:
As I have already mentioned, the current update can let us do both. Yes, this approach is using some extra conditions, but it may cost an unnoticeable time to process, indeed it may speed up the build process in some cases. However, if you are satisfied with extra builds, I can remove all the conditions and may be hardcoded the OFF value to the DAGMC build.

doc/CHANGELOG.rst
```diff
@@ -7,9 +7,13 @@ DAGMC Changelog
 Next version
 ====================
 
+v3.2.3
```
gonuke: Something looks wrong here relative to the released version v3.2.3 - maybe needs a rebase?

gonuke (changes requested):
I think this still me need a rebase at least for the CHANGELOG?

gonuke:
> As I have already mentioned, the current update can let us do both. Yes, this approach is using some extra conditions, but it may cost an unnoticeable time to process, indeed it may speed up the build process in some cases. However, if you are satisfied with extra builds, I can remove all the conditions and may be hardcoded the OFF value to the DAGMC build.

Maybe it is OK in its current form.  The only issue I'm not sure about is that we only build Embree when DD is on, and we do so after MOAB.  If we did it always (whether or not DD is on) as part of the external dependencies (e.g. near GEANT or HDF5) it would make the test for DD faster because it would not always have to also build Embree.

ahnaf-tahmid-chowdhury:
Actually, Embree will be built only once for the docker tag that contains `double_down_[version]` (different tag). It will not be built every time if we do not modify anything in the Docker file. Even if we change the version of Double Down, it will not be built, as it comes before Double Down. When considering the order of dependencies we build first, we may need to take into account which of them generally gets updated often. So, let me know how do I reorder them.

ahnaf-tahmid-chowdhury:
Is there anything more to change on this? Or, I think @gonuke  is awaiting @pshriwise's response on using `find_package` through CMake to specify the minimum version of DD.

cmake/DAGMC_macros.cmake
gonuke: Pinging @pshriwise again on this question - it may be the last thing holding up this PR

ahnaf-tahmid-chowdhury:
@gonuke, do you like to enable testing for DD v1.1.0 as you have mentioned earlier?

CI/Dockerfile
```diff
@@ -169,47 +155,75 @@ ENV moab_install_dir=${install_dir}/moab
...
+            -DEMBREE_TASKING_SYSTEM=INTERNAL \
+            -DEMBREE_ISPC_SUPPORT=OFF \
+            -DEMBREE_TUTORIALS=OFF \
+            -DEMBREE_TBB_ROOT=/usr && \
+        make -j${CI_JOBS} && \
+        make -j${CI_JOBS} install && \
+        cd && \
+        rm -rf ${embree_build_dir}; \
+    fi
+
```
gonuke: Embree should be built/installed as part of `external_deps` since it is truly external and we expect it to change less often.

cmake/DAGMC_macros.cmake
gonuke: ping @pshriwise

gonuke (changes requested):
One more change and still waiting on @pshriwise to chime in

CI/Dockerfile
gonuke: I know we discussed this before and my instinct is once again for it to come earlier.  We do change MOAB more frequently than we change Embree, so it makes sense for Embree to come first - and it is "external" as that stage suggests.

ahnaf-tahmid-chowdhury:
@gonuke, Please note DD tests are turned off. Let me know, if you like to turn this on to v1.1.0

cmake/DAGMC_macros.cmake
pshriwise: I think @gonuke is right. We should use the version argument of `find_package`.

pshriwise: Ah, but now I'm remembering why we opted for this approach originally -- version ranges in `find_packages` are only supported since CMake 3.19 and our [CMakeLists.txt supports CMake back to 3.1](https://github.com/svalinn/DAGMC/blob/d9adb178ac97d19b423daff07c2bd731fbcc92cf/CMakeLists.txt#L2).

pshriwise: So, if the desire is to specify a specific version of Double-Down then I'd vote to move to using `find_package` to do the version check, but if we really do want to support versions of Double-Down >= 1.1.0 then I think we need to keep the version check separate from the `find_package` call or update our minimum req'd CMake version.

ahnaf-tahmid-chowdhury: We are currently using `find_package` to set min version of Geant4 through PR #803 [here](https://github.com/svalinn/DAGMC/blob/50e5c8caa5990b27d66427dc69272302de3326f6/src/geant4/CMakeLists.txt#L3).

ahnaf-tahmid-chowdhury: Given that, CMake 3.1 is now quite old, I'm wondering if it's still capable of building the current build since we typically use Ubuntu 20.04 and 22.04 to build DAGMC in our workflow, which comes with a more recent version of CMake. Additionally, I've noticed a deprecation warning when using CMake 3.1, which suggests we upgrade to version 3.5.

pshriwise: > We are currently using find_package to set min version of Geant4 through PR https://github.com/svalinn/DAGMC/pull/803 [here](https://github.com/svalinn/DAGMC/blob/50e5c8caa5990b27d66427dc69272302de3326f6/src/geant4/CMakeLists.txt#L3).

I believe that this use of the signature indicates that the Geant4 should _match_ the specified version, not a minimum required version. 

---

From the CMake documentation on `find_package`: 

The [version] argument requests a version with which the package found should be compatible. There are two possible forms in which it may be specified:

  - A single version with the format major[.minor[.patch[.tweak]]], where each component is a numeric value.

  - A version range with the format versionMin...[<]versionMax where versionMin and versionMax have the same format and constraints on components being integers as the single version. By default, both end points are included. By specifying <, the upper end point will be excluded. Version ranges are only supported with CMake 3.19 or later.

ahnaf-tahmid-chowdhury: @gonuke, what do you like to do? Should I consider updating CMake version to v3.19 from v3.1 (Ubuntu 20.04 uses [CMake v3.16.3](https://launchpad.net/ubuntu/+source/cmake/3.16.3-1ubuntu1))? Or I may change the current method for setting Geant4 min version?

gonuke: The Geant4 handling is interesting since we currently list a required version, but then handle versions newer than the required version and it works.  I guess I'm a little confused about what **_should_** work and what **_does_** work.

ahnaf-tahmid-chowdhury: Ubuntu 20.04 is failing as it uses CMake 3.16. Here is the [workflow](https://github.com/svalinn/DAGMC/actions/runs/7987368761/job/21809714439#step:7:1). I have set the DD version to 1.0.0 and it seems CMake is ignoring to check the minimum version of DD.

.github/workflows/docker_publish.yml
```diff
@@ -14,28 +14,29 @@ jobs:
...
-        double_down : [
-          OFF,
-        ]
-        geant_version : [
+        geant4_version : [
           10.7.4,
           11.1.2
         ]
+        double_down_version : [
+          v1.0.0,
```
gonuke: We know this version fails, so we should remove it.

We may also want to add `off` to this list so we can test without DD.

.github/workflows/docker_publish.yml
```diff
@@ -77,28 +95,28 @@ jobs:
...
-        double_down : [
-          OFF,
-        ]
-        geant_version : [
+        geant4_version : [
           10.7.4,
           11.1.2
         ]
+        double_down_version : [
+          off
```
gonuke: This matrix should probably match the matrix above to make sure all the right images exist.

gonuke (changes requested):
We've been churning away at this for a while, so it probably makes sense to recap what the goals of this PR are:
* Building CI testing images
    * install a valid version of Double Down in all CI testing images - currently v1.1.0 but NOT v1.0.0
    * test the CI testing images with DAGMC both with DD and without
* testing for each PR
    * test with DD and without for each PR
* testing for merge
    *  test DD `develop` branch upon merge as advisory

In addition, there is a lot of cleanup of `build-arg` variables and other variables.

Given all this, in addition to the code comments, I think we should turn on DD testing in `linux_build_test.yml`

.github/workflows/docker_publish.yml
ahnaf-tahmid-chowdhury: I have added this to check if CMake 3.16 shows any error messages since it is lower than 1.1.0, but it just passed. That means, CMake will also get failed for Geant4 as well.

gonuke:
Needs a rebase following the recent housekeeping update

gonuke (approved):

gonuke:
I think we've addressed your comment @pshriwise and are ready to merge

pshriwise (approved):
Thanks for your effort here @ahnaf-tahmid-chowdhury! And thanks to @gonuke for the thorough discussion.

.github/actions/upstream-test/action.yml
```diff
@@ -43,11 +43,24 @@ runs:
       - name: Installing Dependencies in Docker image
         uses: firehed/multistage-docker-build-action@v1
         with:
-          repository: ghcr.io/${{ github.repository_owner }}/dagmc-ci-ubuntu-${{ inputs.ubuntu_version }}-${{ inputs.compiler}}-geant4_${{ inputs.geant_version }}-hdf5_${{ inputs.hdf5_version }}-moab_${{ inputs.moab_version }}
+          repository: >
```
gonuke, line 46: @ahnaf-tahmid-chowdhury : This change introduced a failure - you can’t split this across multiple lines - that we didn’t review carefully enough in the PR.  Please submit an update to fix this.

ahnaf-tahmid-chowdhury: Let's wait for all the new Docker images to finish building, then we can rerun the Linux tests to see what happens.

gonuke: I looked at the failed logs and saw that the failure is because this multi-line container name is resolving with spaces in the name, so it definitely needs to be fixed.

PR #932, opened 2023-12-20
Update Double Down to 1.1.0

ahnaf-tahmid-chowdhury:
Update Double Down to 1.1.0

Please note: This PR sets `DOUBLE_DOWN=ON` by default


gonuke:
I don't think this has all the logic quite right for when we use `double_down_version` and when we use `double_down`.

Perhaps contrary to some previous advice, I think we need to pass both as `build-args`.  One let's us update and test newer versions of Double Down and one lets us decide whether to build with it or not.  I think we may have also decided to always test/build with DD now.

ahnaf-tahmid-chowdhury:
Should I follow this PR or close this one as I think I am working on PR #929? As you have mentioned to do the both test, I say, according to the current stage of the PR #929, we can do the build both just adding a single word to the Docker build workflow:

```
matrix:
    double_down_version : [
        off, 
        v1.1.0
    ]
```

gonuke:
I forgot about #929  and agree that maybe we should close/withdraw this PR.

Issue #934, opened 2024-01-10
Implicit Complement Position in DAGMC Indices


pshriwise:
### Short Version

A constant defining the structure of the `EntityHandle` space in MOAB was changed between MOAB 5.3.0 and MOAB 5.4.0 (commit linked below), revealing an issue with placement of the implicit complement handle in the DAGMC indices structure.

For a file with enough `MeshSet`'s to be read, the implicit complement handle may not appear at the end of the `EntityHandle` space, meaning that one may perform a point containment check on the implicit complement volume before other, explicit volumes if looping over all volumes in the model. While its surprising that this constant changed, this is not the fault of MOAB, `EntityHandles` are required to be unique but not monotonically increasing when generated. This bug was present before the change in MOAB, but the change makes the likelihood of it occurring much higher.

Searching the implicit complement before other volumes can cause problems for source points that are coincident with surfaces. This may only be true for problems with imperfect topology relationships (failed imprinting/merging) and is why it didn't show up in the OpenMC tests, which use only well-formed models.

The fix for us would be to take extra care to ensure that the implicit complement volume is placed at the back of our data structures in `DagMC::build_indices`.

In external code interfaces, it would probably be wise to have a check that the implicit complement is checked last. In OpenMC, for example, this would look like the following in  `openmc::DAGUniverse::find_cell`:

```C++
bool DAGUniverse::find_cell(Particle& p) const
{
  // if the particle isn't in any of the other DagMC
  // cells, place it in the implicit complement
  for (auto cell_idx : cells_) {
	// skip the implicit complement for now
    if (cell_idx == this->implicit_complement_idx()) continue;
    
    const auto& cell = model::cells[cell_idx];
    if(cell->contains(p.r(), p.u(), p.surface())) {
      p.lowest_coord().cell = cell_idx;
      return true;
    }
  }

  // finally, check the implicit complement for containment
  const auto& ipc_cell = model::cells[implicit_complement_idx()];
  if (cell->contains(p.r(), p.u(), p.surface())) {
    p.lowest_coord().cell = implicit_complement_idx();
    return true;
  }

  if (model::universe_map[this->id_] != model::root_universe) {
    p.lowest_coord().cell = implicit_complement_idx();
    return true;
  }
  return false;
}
```

#### What changed in MOAB?

[...]

PR #935, opened 2024-01-11
Ensure implicit complement is placed at the back of the DAGMC index space

pshriwise:
## Description
This PR enforces placement of the implicit complement handle at the back of the surfaces and volumes vectors (contained in the `DagMC::entHandles` data member). It also changes the `EntityHandle` -> DAGMC index mapping structure (`DagMC::entIndices`) s.t. DAGMC indexes are independent of the EntityHandle ordering produced by MOAB. I decided on `std::unordered_map` for a lookup complexity O(1). The downside here is that the container use more memory, but this should be a small percentage of the overall memory use in DAGMC as I understand it. If others have approaches for maintaining the `std::vector<int>` option I'm all ears, but I have a feeling it would be more trouble than it's worth.

I've confirmed that this results in the same eigenvalue as MOAB 5.3.0 for the MSRE model mentioned in #934 with OpenMC.

## Motivation and Context
The motivation for this is described in #934.

## Changes
- Manual placement of the implicit complement handle at the back of the `entHandles` container.
- Structural change in the container type used for DAGMC indices.

## Behavior
In some cases, the implicit complement handle may not be placed at the back of the DAGMC index space and could be checked in point containment queries before other explicit volumes in the model.

### DIFF
--- a/src/dagmc/DagMC.hpp
+++ b/src/dagmc/DagMC.hpp
@@ -8,6 +8,7 @@
 #include <memory>
 #include <stdexcept>
 #include <string>
+#include <unordered_map>
 #include <vector>
 
 #include "DagMCVersion.hpp"
@@ -77,10 +78,12 @@ class DagMC {
         double overlap_tolerance = 0., double numerical_precision = .001,
         int verbosity = 1);
   // Deprecated Constructor
-  [[deprecated(
-      "Replaced by DagMC(std::shared_ptr<Interface> mb_impl, ... "
-      ")")]] DagMC(Interface* mb_impl, double overlap_tolerance = 0.,
-                   double numerical_precision = .001, int verbosity = 1);
+  [
+      [deprecated("Replaced by DagMC(std::shared_ptr<Interface> mb_impl, ... "
+                  ")")]] DagMC(Interface* mb_impl,
+                               double overlap_tolerance = 0.,
+                               double numerical_precision = .001,
+                               int verbosity = 1);
   // Destructor
   ~DagMC();
 
--- a/src/dagmc/DagMC.hpp
+++ b/src/dagmc/DagMC.hpp
@@ -4,7 +4,6 @@
 #include <assert.h>
 
 #include <limits>
-#include <unordered_map>
 #include <map>
 #include <memory>
 #include <stdexcept>
@@ -78,12 +77,10 @@ class DagMC {
         double overlap_tolerance = 0., double numerical_precision = .001,
         int verbosity = 1);
   // Deprecated Constructor
-  [
-      [deprecated("Replaced by DagMC(std::shared_ptr<Interface> mb_impl, ... "
-                  ")")]] DagMC(Interface* mb_impl,
-                               double overlap_tolerance = 0.,
-                               double numerical_precision = .001,
-                               int verbosity = 1);
+  [[deprecated(
+      "Replaced by DagMC(std::shared_ptr<Interface> mb_impl, ... "
+      ")")]] DagMC(Interface* mb_impl, double overlap_tolerance = 0.,
+                   double numerical_precision = .001, int verbosity = 1);
   // Destructor
   ~DagMC();
 
--- a/src/dagmc/DagMC.hpp
+++ b/src/dagmc/DagMC.hpp
@@ -4,6 +4,7 @@
 #include <assert.h>
 
 #include <limits>
+#include <unordered_map>
 #include <map>
 #include <memory>
 #include <stdexcept>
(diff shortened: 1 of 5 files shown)


gonuke:
This housekeeping failure has dogged me before... there is some inconsistency that I haven't bothered to get to the bottom of, and it requires me to manually override the style guide for this line.

src/dagmc/DagMC.hpp
```diff
@@ -571,8 +570,8 @@ inline EntityHandle DagMC::entity_by_index(int dimension, int index) const {
 }
 
 inline int DagMC::index_by_handle(EntityHandle handle) const {
-  assert(handle - setOffset < entIndices.size());
-  return entIndices[handle - setOffset];
+  assert(entIndices.count(handle) > 0);
+  return entIndices.at(handle);
```
gonuke, line 574: Is there a reason to use this instead of the `[]` operator?

gonuke (commented):
Looks good to me - should we add a test for this?  I guess we'd have to contrive a set of volumes where the implicit complement was in the wrong place?

src/dagmc/DagMC.hpp
pshriwise: This method is `const`. The `[]` operator violates this as it would also support assignment.

pshriwise:
> Looks good to me - should we add a test for this? I guess we'd have to contrive a set of volumes where the implicit complement was in the wrong place?

Yeah, we'd ideally test this. I think we could do it by using the `DacMC` constructor that takes a MOAB instance. I think we could artificially occupy a bunch of MeshSet handles before passing it to DAGMC to ensure that loading a DAGMC file triggers this situation.

ahnaf-tahmid-chowdhury:
You mean we will use a sample model (DAGMC file) in `DacMC` constructor and then we will check the output with predefined results?

src/dagmc/DagMC.hpp
gonuke: Thanks for the education!

gonuke:
We would probably not use a file. Probably manually create a minimal DAGMC model in memory for testing.

ahnaf-tahmid-chowdhury:
Thanks for the clarification! Creating a minimal DAGMC model in memory for testing sounds like a good approach to minimize source code size. 

I'm always happy to contribute by making a PR and generating the test file. However, I'm unsure about the process for creating tests for DAGMC. Could you please provide a sample or some guidance to help me get started?

gonuke:
Proposed minimal model:
* 3 volumes created in order 1, 2, 3
* populate set 1 with surfaces/facets that represent a central cube
* populate set 3 with surfaces/facets that represent an outer graveyard prismatic shell
* populate set 2 as the implicit complement built from the surfaces of sets 1 & 2

Load all of that into a MOAB entity as if it was load_file and then run the code to test this PR's addition.

gonuke:
Alternative:
* load a file with an existing implicit complement (IC)
* create a new volume set (assumed to be added AFTER the IC)
* move all the topological relationships from some other volume set (the first one?) to this new one
* delete the original volume set

Run these methods to move the IC and test

pshriwise:
> Proposed minimal model:
> 
> * 3 volumes created in order 1, 2, 3
> * populate set 1 with surfaces/facets that represent a central cube
> * populate set 3 with surfaces/facets that represent an outer graveyard prismatic shell
> * populate set 2 as the implicit complement built from the surfaces of sets 1 & 2
> 
> Load all of that into a MOAB entity as if it was load_file and then run the code to test this PR's addition.

I ended up taking a slightly different approach here. To trigger this condition we need to load a file with at least `DEFAULT_MESHSET_SEQUENCE_SIZE`  (currently set to `16 * 1024`) sets. I've loaded a test DAGMC file, added a bunch of sets to it (we technically only need 16,384, but I've used one million to be safe), we then write a temporary file, read it back in, and setup the DAGMC implicit complement and indices. All of the additional sets should be ignored, but the file read should still trigger this condition.

To convince myself that this is true, I created a branch that adds this test on top of `develop` to ensure that it fails with the implicit complement placed at the beginning of the DAGMC index space.

https://github.com/pshriwise/DAGMC/actions/runs/7748743790/job/21131962325#step:5:100

image

But with the latest CI run, that same test now passes:

https://github.com/svalinn/DAGMC/actions/runs/7748757514/job/21132001898?pr=935

image

I rather like this test as it demonstrates that the behavior is caused by the number of meshsets in the file, even if they have no bearing on the definition of a DAGMC model.

gonuke (approved):
Thanks for taking care of this @pshriwise

PR #941, opened 2024-02-15
Remove eigen3

### LINKED ISSUE #799: Remove need for downstream codes to have Eigen3 installed

ahnaf-tahmid-chowdhury:
This PR removes Eigen3 from DAGMC dependency.

### DIFF
--- a/.github/workflows/mac_build_test.yml
+++ b/.github/workflows/mac_build_test.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Initial setup
         shell: bash -l {0}
         run: |
-          brew install eigen hdf5
+          brew install hdf5
 
       - name: Environment Variables
         shell: bash -l {0}
--- a/.github/workflows/windows_build_test.yml
+++ b/.github/workflows/windows_build_test.yml
@@ -108,6 +108,7 @@ jobs:
                 -DBUILD_MAKE_WATERTIGHT=ON \
                 -DBUILD_OVERLAP_CHECK=OFF \
                 -DBUILD_TESTS=ON \
+                -DENABLE_BLASLAPACK=OFF \
                 -DMOAB_DIR=../install_dir \
                 -DHDF5_ROOT="${CONDA_LOC}" \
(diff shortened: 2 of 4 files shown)


ahnaf-tahmid-chowdhury:
What!!! I have turned off BLASLAPACK, but it seems the CMake is still considering it enabled!

ahnaf-tahmid-chowdhury:
I see, we need to enable `blaslapack` and need to find `lapack`, which will enable Fortran support. In that way we can ignore Eigen3 support for MOAB 😕. @gonuke, do you like to proceed with `blaslapack`?

More info: [here](https://bitbucket.org/fathomteam/moab/src/ab73c4a1301bbdaf4cb8bf845389ef8493923317/CMakeLists.txt#lines-264)

gonuke:
We definitely want Eigen3 and not BLAS and LAPACK.

gonuke:
I'm not sure what problem this is trying to solve?

ahnaf-tahmid-chowdhury:
This one #799

gonuke:
> This one #799

I thought that might be it.  This is a different problem.  The problem is not that Eigen3 is a dependency of DAGMC.  The problem is that Eigen3 is included in header files of MOAB and DAGMC that are required by downstream apps, even when it may not be necessary.  The solution to this requires a little more careful analysis of each header file that included Eigen3 headers, moving the Eigen3 include to the source file, and seeing whether some forward declarations are necessary where the Eigen3 has been removed from headers.

ahnaf-tahmid-chowdhury:
Thank you for highlighting the issue with Eigen3 inclusion in header files.

PR #943, opened 2024-02-20
Simplify Housekeeping Process for DAGMC

ahnaf-tahmid-chowdhury:
# Description
This PR simplifies the housekeeping process for DAGMC by removing the dependency on a Docker container and installing `clang-format` directly through the workflow.

# Related Issues
Housekeeping of DAGMC is very complicated now. And it seems in the docker file, we are just installing clang-format through apt. Also, I noticed the docker publish file is changed and there is no way to rebuild a new docker image for housekeeping.

# Changes
- Removed the use of a Docker container for housekeeping.
- Installed `clang-format` directly through the workflow.

### DIFF
--- a/src/geant4/app/include/ExN01UserScoreWriter.hh
+++ b/src/geant4/app/include/ExN01UserScoreWriter.hh
@@ -1,7 +1,6 @@
 #ifndef ExN01UserScoreWriter_h
 #define ExN01UserScoreWriter_h 1
 
-// #include "MBCore.hpp"
 #include <vector>
 
 #include "G4VScoreWriter.hh"
--- a/src/geant4/app/include/ExN01UserScoreWriter.hh
+++ b/src/geant4/app/include/ExN01UserScoreWriter.hh
@@ -1,7 +1,7 @@
 #ifndef ExN01UserScoreWriter_h
 #define ExN01UserScoreWriter_h 1
 
-//#include "MBCore.hpp"
+// #include "MBCore.hpp"
 #include <vector>
 
 #include "G4VScoreWriter.hh"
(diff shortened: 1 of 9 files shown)


gonuke:
perhaps we should update these two files to match what clang-format is generating as part of this PR so that it passes tests

gonuke (changes requested):
This looks good - I thought I had done this already :) but maybe it's lost in my local branches somewhere.

We can perhaps take care of the two formatting inconsistencies in the current code that are causing this to fail right now.  I wonder if they are due to slight changes in `clang-format`'s behavior since the docker images was generated.

src/geant4/app/include/ExN01UserScoreWriter.hh
```diff
@@ -1,7 +1,7 @@
 #ifndef ExN01UserScoreWriter_h
 #define ExN01UserScoreWriter_h 1
 
-//#include "MBCore.hpp"
+// #include "MBCore.hpp"
```
gonuke: I think we can just delete this line

gonuke (changes requested):
One small request

gonuke (dismissed):
Thanks for streaming in this @ahnaf-tahmid-chowdhury

gonuke:
This now has a conflict in the `CHANGELOG.rst`

gonuke:
It looks like there is an update in CMake version in the new Windows runner... I'm not sure yet if/why that is causing a problem.

gonuke:
We may need to provide more information to our CMake config for MOAB. It may be true that MOAB should improve their custom `FindEigen` macro, but providing `EIGEN3_DIR` may be easier

gonuke (approved):
Thanks for cleaning this up @ahnaf-tahmid-chowdhury

PR #944, opened 2024-02-22
Allow Double Down v1.1.0 Installation in Dockerfile

### LINKED ISSUE #947: Failiure to build DAGMC with Double Down
Looks to be a recently introduced bug in the DAGMC_macros.cmake file, where looking to compare the version number of double down. The issue is however in the `ddConfigVersion.cmake` produced from the DD build, the version number is set as a string "1.1.0" for example, but the `DAGMC_macros.cmake` file compares against against I dont really know what - the number is input as 1.1.0 - but its not a string nor an int. 

I think either double down should be setting DD_MAJOR_VERSION 1 & DD_MINOR_VERSION 1, and those get checked in the macros file, or something similar. I'm surprised CI passes right now as is. Happy to fix, but need to know how DD should be fixed before upstream changes are made.

ahnaf-tahmid-chowdhury:
Trying to fix the broken Docker build

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -34,7 +34,6 @@ jobs:
           11.1.2
         ]
         double_down_version : [
-          off,
           v1.1.0,
         ]
 
@@ -115,7 +114,6 @@ jobs:
           11.1.2
         ]
         double_down_version : [
-          off,
           v1.1.0,
         ]
 
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -30,10 +30,12 @@ jobs:
           5.5.1,
         ]
         geant4_version : [
+          off,
           10.7.4,
           11.1.2
         ]
         double_down_version : [
+          off,
           v1.1.0,
         ]
 
@@ -110,10 +112,12 @@ jobs:
           5.5.1,
         ]
         geant4_version : [
+          off,
           10.7.4,
           11.1.2
         ]
         double_down_version : [
+          off,
           v1.1.0,
         ]
 
--- a/.github/actions/upstream-test/action.yml
+++ b/.github/actions/upstream-test/action.yml
@@ -24,7 +24,7 @@ inputs:
   moab_version:
     description: Version of MOAB
     required: false
-    default: 5.3.0
+    default: 5.5.1
   double_down_version:
     description: Version of Double Down
     required: false
--- a/CI/Dockerfile
+++ b/CI/Dockerfile
@@ -39,7 +39,6 @@ RUN apt-get update --yes && \
         g++ \
         gcc \
         gfortran \
-        libhdf5-dev \
         libtool \
         libeigen3-dev\
         python3 \
--- a/cmake/DAGMC_macros.cmake
+++ b/cmake/DAGMC_macros.cmake
@@ -90,7 +90,7 @@ macro (dagmc_setup_options)
 
 if (DOUBLE_DOWN)
   find_package(DOUBLE_DOWN REQUIRED)
-  if (DOUBLE_DOWN_VERSION VERSION_LESS 1.1.0)
+  if (dd_VERSION VERSION_LESS 1.1.0)
     message(FATAL_ERROR "Discovered Double Down Version: ${DOUBLE_DOWN_VERSION}. \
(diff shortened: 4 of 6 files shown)


ahnaf-tahmid-chowdhury:
Is it necessary to test every time our build, once with DD off and once with DD on? This may create extra runs every time.

.github/workflows/docker_publish.yml
```diff
@@ -30,6 +30,7 @@ jobs:
           5.5.1,
         ]
         geant4_version : [
+          off,
```
gonuke, line 33: Not sure it's worth have an extra image without GEANT?

gonuke (changes requested):
Is there a reason to add a geant4 off version

.github/workflows/docker_publish.yml
ahnaf-tahmid-chowdhury: While updating [action.yml](https://github.com/svalinn/DAGMC/blob/develop/.github/actions/upstream-test/action.yml#L11), I have noticed we are not using Geant4 by default. Do you like to change the default?

ahnaf-tahmid-chowdhury:
We need to delete the following cache images:

https://github.com/orgs/svalinn/packages/container/package/dagmc-ci-ubuntu-20.04-clang-hdf5_1.14.3-moab_5.4.1-geant4_10.7.4-double_down_v1.1.0%2Fmoab

https://github.com/orgs/svalinn/packages/container/package/dagmc-ci-ubuntu-20.04-clang-hdf5_1.14.3-moab_5.4.1-geant4_11.1.2-double_down_v1.1.0%2Fmoab

https://github.com/orgs/svalinn/packages/container/package/dagmc-ci-ubuntu-20.04-clang-hdf5_1.14.3-moab_5.5.1-geant4_10.7.4-double_down_v1.1.0%2Fmoab

https://github.com/orgs/svalinn/packages/container/package/dagmc-ci-ubuntu-20.04-clang-hdf5_1.14.3-moab_5.5.1-geant4_11.1.2-double_down_v1.1.0%2Fmoab

https://github.com/orgs/svalinn/packages/container/package/dagmc-ci-ubuntu-22.04-clang-hdf5_1.14.3-moab_5.4.1-geant4_10.7.4-double_down_v1.1.0%2Fmoab

https://github.com/orgs/svalinn/packages/container/package/dagmc-ci-ubuntu-22.04-clang-hdf5_1.14.3-moab_5.4.1-geant4_11.1.2-double_down_v1.1.0%2Fmoab

https://github.com/orgs/svalinn/packages/container/package/dagmc-ci-ubuntu-22.04-clang-hdf5_1.14.3-moab_5.5.1-geant4_10.7.4-double_down_v1.1.0%2Fmoab

https://github.com/orgs/svalinn/packages/container/package/dagmc-ci-ubuntu-22.04-clang-hdf5_1.14.3-moab_5.5.1-geant4_11.1.2-double_down_v1.1.0%2Fmoab

.github/workflows/docker_publish.yml
gonuke: The treatment of GEANT and Double Down are a little different from each other.

1) Testing with or without GEANT only demonstrates whether DAGMC will integrate successfully with GEANT as a transport code.

2) Testing with or without Double Down modifies the underlying fundamental behavior of DAGMC with all transport codes.

A few conclusions of this:

A) It makes sense to have fewer images, i.e. none that are build without GEANT and without DD because we can test our build without them even if they are installed.

B) It makes sense to do the advisory testing with the minimal number of options turned on, ie. without GEANT even if it is installed in the image.

ahnaf-tahmid-chowdhury: I've made changes to streamline the build process and reduce the number of arguments as much as possible. But it seems, I've ended up creating extra Docker images. 🙂

I think we have two options:

1. Add the extra arguments again and remove conditions from the Dockerfile.
2. Consider setting GEANT4 and DD always on in `action.yml`.

ahnaf-tahmid-chowdhury: Let me know, which step to proceed. Or, any other suggestions?

gonuke (dismissed):

gonuke:
Looks like there are merge conflicts now

gonuke (approved):

PR #945, opened 2024-02-26
Add inline documentation for the DagMC and dagmcmetadata classes.

pshriwise:

## Description

Adding some inline documentation so we can start leveraging it for rendered documentation.

### DIFF
--- a/src/dagmc/DagMC.hpp
+++ b/src/dagmc/DagMC.hpp
@@ -218,20 +218,24 @@ class DagMC {
   *
   * @param volume The volume within which the ray is fired.
   * @param ray_start A 3-element array representing the starting point of the ray.
-  * @param ray_dir A 3-element array representing the direction of the ray.
+  * @param ray_dir A 3-element array representing the unit direction of the ray.
   * @param[out] next_surf The handle of the next surface hit by the ray.
   * @param[out] next_surf_dist The distance from the ray start to the next surface.
   * @param history Optional. A pointer to a RayHistory object for storing the ray's path.
   * @param dist_limit Optional. The maximum distance the ray should travel. Default is 0, which means no limit.
-  * @param ray_orientation Optional. The orientation of the ray. Default is 1.
-  * @param stats Optional. A pointer to a TrvStats object for storing traversal statistics.
+  * @param ray_orientation Optional. The orientation triangle normals considered valid for a hit.Default is 1 (forward).
+  * @param stats Optional. A pointer to a TrvStats object for storing traversal statistics of the ray. Default is NULL.
   *
   * @return Returns an ErrorCode indicating the success or failure of the operation.
   */
-  ErrorCode ray_fire(const EntityHandle volume, const double ray_start[3],
-                     const double ray_dir[3], EntityHandle& next_surf,
-                     double& next_surf_dist, RayHistory* history = NULL,
-                     double dist_limit = 0, int ray_orientation = 1,
+  ErrorCode ray_fire(const EntityHandle volume,
+                     const double ray_start[3],
+                     const double ray_dir[3],
+                     EntityHandle& next_surf,
+                     double& next_surf_dist,
+                     RayHistory* history = NULL,
+                     double dist_limit = 0,
+                     int ray_orientation = 1,
                      OrientedBoxTreeTool::TrvStats* stats = NULL);
 
 
@@ -240,25 +244,32 @@ class DagMC {
   *
   * @param volume The volume within which the point is being tested.
   * @param xyz A 3-element array representing the coordinates of the point.
-  * @param[out] result The result of the operation. It will be set to 1 if the point is inside the volume, 0 if it is outside, and -1 if the point is on the boundary.
-  * @param uvw Optional. A 3-element array representing the direction from the point to the volume. Default is NULL.
-  * @param history Optional. A pointer to a RayHistory object for storing the ray's path. Default is NULL.
+  * @param[out] result The result of the operation. It will be set to 1 if the
+  * point is inside the volume and 0 if it is outside.
+  * @param uvw Optional. A 3-element array representing the unit direction from the
+  * point to the volume. Default is NULL. A randomly generated direction will be
+  * used if not provided.
+  * @param history Optional. A pointer to a RayHistory object for masking out previously hit triangles. Default is NULL.
   *
-  * @return Returns an ErrorCode indicating the success or failure of the operation.
+  * @return Returns an ErrorCode indicating the success or failure of the
+  * operation.
   */
-  ErrorCode point_in_volume(const EntityHandle volume, const double xyz[3],
-                            int& result, const double* uvw = NULL,
+  ErrorCode point_in_volume(const EntityHandle volume,
+                            const double xyz[3],
+                            int& result,
+                            const double* uvw = NULL,
                             const RayHistory* history = NULL);
   /**
   * @brief Determines whether a given point is inside a specified volume. This function is slower than point_in_volume.
   *
   * @param volume The volume within which the point is being tested.
   * @param xyz A 3-element array representing the coordinates of the point.
-  * @param[out] result The result of the operation. It will be set to 1 if the point is inside the volume, 0 if it is outside, and -1 if the point is on the boundary.
+  * @param[out] result The result of the operation. It will be set to 1 if the point is inside the volume and 0 if it is outside.
   *
   * @return Returns an ErrorCode indicating the success or failure of the operation.
   */
-  ErrorCode point_in_volume_slow(const EntityHandle volume, const double xyz[3],
+  ErrorCode point_in_volume_slow(const EntityHandle volume,
+                                 const double xyz[3],
                                  int& result);
 
 #if MOAB_VERSION_MAJOR == 5 && MOAB_VERSION_MINOR > 2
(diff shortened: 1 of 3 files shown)


src/dagmc/dagmcmetadata.hpp
```diff
@@ -8,118 +8,317 @@
...
+   */
   ~dagmcMetaData() = default;
 
-  // load the dagmc properties into maps
+  /**
+   * @brief Loads the DagMC properties into internal data maps.
+   *
+   * This function reads the properties from the associated DagMC instance and
+   * stores them in internal data structures for efficient access.
+   */
```
gonuke, line 51: This repeated line is causing a build error.

gonuke (changes requested):
Looks like you ended up doubling a line here causing a build error.

src/dagmc/dagmcmetadata.hpp
pshriwise: Indeed. Thanks

src/dagmc/dagmcmetadata.hpp
```diff
@@ -8,118 +8,317 @@
...
+  void load_property_data();
```
pshriwise, line 52:
```suggestion
```

src/dagmc/DagMC.hpp
```diff
@@ -212,44 +212,214 @@ class DagMC {
...
+   * @return Returns an ErrorCode indicating the success or failure of the
+   * operation.
+   */
   ErrorCode point_in_volume(const EntityHandle volume, const double xyz[3],
                             int& result, const double* uvw = NULL,
                             const RayHistory* history = NULL);
-
+  /**
+   * @brief Determines whether a given point is inside a specified volume. This
+   * function is slower than point_in_volume.
```
gonuke, line 266: Is it fair to say that this method is more robust? That is, should we provide some indication about when/why to use a slower method?

src/dagmc/DagMC.hpp
```diff
@@ -212,44 +212,214 @@ class DagMC {
...
                                  int& result);
+
 #if MOAB_VERSION_MAJOR == 5 && MOAB_VERSION_MINOR > 2
+  /**
+   * @brief Finds the volume containing a given point.
+   *
+   * @param xyz A 3-element array representing the coordinates of the point.
+   * @param[out] volume The volume containing the point.
+   * @param uvw Optional. A 3-element array representing the unit direction to
+   * be used when firing a test ray. Default is NULL.
```
gonuke: Will this also use a random direction if nothing is specified?

src/dagmc/dagmcmetadata.hpp
```diff
@@ -8,118 +8,316 @@
...
+  /**
+   * @brief Splits a string on the basis of the first delimiter it finds.
+   *
+   * @param property_string The string to be split.
+   * @param delimiter The delimiter on which to split the string.
+   *
+   * @return Returns a pair of strings. The first element of the pair is the
+   *         substring before the first occurrence of the delimiter. The second
+   *         element is the substring after the first occurrence of the
+   * delimiter.
```
ahnaf-tahmid-chowdhury:
```suggestion
   *         delimiter.
```

src/dagmc/DagMC.hpp
```diff
@@ -212,44 +212,214 @@ class DagMC {
...
+   * crossing. It can be used to check whether a direction change redirects the
+   * ray back into the originating volume.
+   *
+   * @param volume The volume to be tested.
+   * @param surface The surface to be tested.
+   * @param xyz A 3-element array representing the coordinates of the point.
+   * @param uvw A 3-element array representing the unit direction from the point
+   * to the volume.
+   * @param[out] result The result of the operation. If present and
+   * non-empty, the history is used to look up the surface facet at which the
```
gonuke: It looks like there is an entry missing and that this description applies to the `RayHistory` and not this `results` parameter

src/dagmc/dagmcmetadata.hpp
```diff
@@ -8,118 +8,316 @@
...
+   * @brief Finalizes the counters after all data has been parsed.
+   */
   void finalise_counters();
 
-  // Parse property for entities with the specified dimension and delimiters.
-  // Optionally remove duplicate property values if necessary.
+  /**
+   * @brief Parses property for entities with the specified dimension and
+   *        delimiters. Optionally removes duplicate property values if
+   * necessary.
```
ahnaf-tahmid-chowdhury:
```suggestion
   *        necessary.
```

src/dagmc/dagmcmetadata.hpp
```diff
@@ -8,118 +8,316 @@
 
 class dagmcMetaData {
  public:
-  // Constructor
+  /**
+   * @brief Constructs a new DagmcMetadata object.
+   *
+   * @param dagmc A pointer to the DagMC instance associated with this metadata.
+   */  // Constructor
```
ahnaf-tahmid-chowdhury:
```suggestion
   */ 
```

ahnaf-tahmid-chowdhury (commented):
Thank you, @pshriwise, for creating this documentation support. It has been very helpful in understanding the source code. I have a few formatting suggestions, although they are not absolutely necessary.

src/dagmc/DagMC.hpp
```diff
@@ -212,44 +212,214 @@ class DagMC {
...
+   *
+   * @return Returns an ErrorCode indicating the success or failure of the
+   * operation.
+   */
   ErrorCode surface_sense(EntityHandle volume, EntityHandle surface,
                           int& sense_out);
 
+  /**
+   * @brief Gets the angle between a specified surface and a ray from a given
+   * point in a specified direction.
```
gonuke: Not sure why we named this `get_angle` but it appears to return the surface normal of the facet, and not an angle...

src/dagmc/dagmcmetadata.hpp
```diff
@@ -8,118 +8,316 @@
 
 class dagmcMetaData {
  public:
-  // Constructor
+  /**
+   * @brief Constructs a new DagmcMetadata object.
+   *
+   * @param dagmc A pointer to the DagMC instance associated with this metadata.
```
gonuke:
```suggestion
   * @param DAGptr A pointer to the DagMC instance associated with this metadata.
```

src/dagmc/dagmcmetadata.hpp
gonuke: Shouldn't we include the two optional parameters?

src/dagmc/dagmcmetadata.hpp
```diff
@@ -8,118 +8,316 @@
...
+   * @return Returns true if the string can be converted to an integer,
+   *         false otherwise.
+   */
   bool try_to_make_int(std::string value);
 
   // private member functions
  private:
-  // parse the material data
+  /**
+   * @brief Parses the material data from the associated DagMC instance.
```
gonuke: What does this actually do? Changes the state of some data structures in some way, but how? (And similarly for following methods)

src/dagmc/dagmcmetadata.hpp
```diff
@@ -8,118 +8,316 @@
...
+   *         are vectors of property values.
+   */
   std::map<moab::EntityHandle, std::vector<std::string>>
   get_property_assignments(std::string property, int dimension,
                            std::string delimiters,
                            bool remove_duplicates = true);
 
-  // remove duplicate properties from the vector of properties
+  /**
+   * @brief Removes duplicate properties from a vector of properties.
```
gonuke, line 223: Is it possible to give a simple example of how this might manifest itself? Like the neutron example you gave below for the `set_remove_rich()` method?

src/dagmc/dagmcmetadata.hpp
```diff
@@ -8,118 +8,316 @@
 
 class dagmcMetaData {
```
gonuke, line 9: This class would probably also benefit from some top level general description of how it all works, e.g. what's a "property" and how are key:value pairs specified, etc...

gonuke (changes requested):
Thanks for this great effort @pshriwise.  I have a few necessary changes (typos, etc) and a few additional requests, although I feel a little sheepish "punishing" your effort with additional requests.

src/dagmc/DagMC.hpp
pshriwise: I have a sense that it is more robust, but am not terribly familiar (or am no longer terribly familiar) with it myself. Do you recall in what situations we should use it over the `point_in_volume` method? I'd be happy to add them if so.

For now, I'll include where the method originates so a developer can make a decision based on the publication.

src/dagmc/DagMC.hpp
pshriwise: Yep yep. I'll include that.

src/dagmc/DagMC.hpp
pshriwise: Mea culpa. Thanks!!

src/dagmc/DagMC.hpp
pshriwise: I'm not sure either, but probably not going to tackle that here. I did try to add some clarity to the brief description though.

src/dagmc/dagmcmetadata.hpp
pshriwise: Gave it a go. Lemme know what you think.

src/dagmc/dagmcmetadata.hpp
pshriwise: ✅

src/dagmc/dagmcmetadata.hpp
```diff
@@ -7,119 +7,339 @@
...
+  /**
+   * @brief Parses the tally volume data from the associated DagMC instance.
+   */
   void parse_tally_volume_data();
-  // finalise the count data
+
+  /**
+   * @brief Finalizes the counters after all data has been parsed.
+   */
   void finalise_counters();
```
pshriwise: I didn't see a definition for this method and have removed it.

src/dagmc/DagMC.hpp
```diff
@@ -212,44 +212,223 @@ class DagMC {
...
+   * @param[out] sense_out The determined sense.
+   *
+   * @return Returns an ErrorCode indicating the success or failure of the
+   * operation.
+   */
   ErrorCode surface_sense(EntityHandle volume, EntityHandle surface,
                           int& sense_out);
 
+  /**
+   * @brief Gets the angle between the normal of a specified surface and a ray
```
gonuke: I don't think this gets an angle. I think it just gets the facet normal at the location of the surface crossing. (Hence my incomplete comment on the poor choice of name... that I don't expect you to fix.)

gonuke (changes requested):
One small clarification

src/dagmc/DagMC.hpp
```diff
@@ -212,44 +212,223 @@ class DagMC {
...
+   * from the ray origin in a specified direction.
+   *
+   * @param surf The surface with respect to which the angle is determined.
+   * @param xyz A 3-element array representing the coordinates of the point.
+   * @param angle[out] A 3-element array in which to store the determined angle.
```
ahnaf-tahmid-chowdhury:
```suggestion
   * @brief Gets the normal vector of the surface at the specified point of
   * intersection by a ray from the ray origin in a specified direction.
   *
   * @param surf The surface with respect to which the normal vector is determined.
   * @param xyz A 3-element array representing the coordinates of the point of 
   * intersection.
   * @param angle[out] A 3-element array in which to store the determined normal vector.
```

gonuke: I don't expect to change the name of the output parameter in this PR since it's only meant to update comments/documentation.

ahnaf-tahmid-chowdhury: Is this right?

src/dagmc/DagMC.hpp
pshriwise: Ah, yeah you're right. Sorry. In a rush updating this last night. On it.

src/dagmc/DagMC.hpp
```diff
@@ -212,44 +212,231 @@ class DagMC {
...
+  /**
+   * @brief Returns the normal vector of a surface at the specified point. from
+   * the ray origin in a specified direction. It is assumed that the specified
+   * point is on the surface.
+   *
+   * This method first identifies which triangle contains this point and then
+   * calculates the unit outward normal of that triangle.  The triangle of the
+   * provided volume that is nearest the provided point is used for this
+   * calculation. The search for that triangle can be circumvented by providing a
+   * RayHistory, in which case the last triangle of the history will be used.
```
gonuke: It looks like our `clang-format` wants to wrap this line differently 🤷 

```suggestion
   * calculation. The search for that triangle can be circumvented by providing
   * a RayHistory, in which case the last triangle of the history will be used.
```

gonuke (changes requested):
Sorry for this clang-format concern...

src/dagmc/DagMC.hpp
pshriwise: No worries! Ran clang-format locally and pushed before I saw this. Sorry to not incorporate the suggestion!

ahnaf-tahmid-chowdhury (approved):
I think everything is fine now. Let's ask @gonuke to merge the changes.

gonuke (approved):
Thanks @pshriwise

PR #951, opened 2024-03-02
Streamline dependencies of docker CI images

### LINKED ISSUE #608: System HDF5 on CI
We should be testing whether it's possible to build with system HDF5 on CI.

### LINKED ISSUE #750: Streamline dependencies of docker CI images
A review of the the dependencies, esp. apt-get and pip, that are included for CI images in order to minimize the size and communicate the true dependency list.

The current CI docker images install some libraries that may no longer be necessary (e.g. BLAS/LAPACK).

ahnaf-tahmid-chowdhury:
## Description

This PR aims to optimize the dependencies of the docker CI images by:

1. Removing unused dependencies.
2. Updating Geant4 to version 11.2.1.
3. Changing the download method of HDF5 and Geant4 to use `git clone` from the source.
4. Switching solely to `git clone` to download dependencies, removing the dependency on `wget`.
5. Creating a new stage to store only binary files, thereby reducing the docker image size.

## Impact

These changes are expected to streamline the dependencies of the docker CI images, reducing the image size and potentially improving performance.

## Related Issues

Closes #750.


ahnaf-tahmid-chowdhury:
Is it necessary to create an extra stage like HDF5 and MOAB, given that these are also external dependencies? I think we can remove these and instead update our workflow to create a binary/final stage to store only the binaries with minimal size, and we may choose the tag as dagmc.

CI/Dockerfile
```diff
@@ -250,28 +237,44 @@ RUN mkdir -p ${dagmc_build_dir}/build && \
...
-    cmake . -DDAGMC_ROOT=${dagmc_install_dir} && \
+RUN cd ${DAGMC_BUILD_DIR}/cmake/test_config && \
+    cmake . -DDAGMC_ROOT=${DAGMC_INSTALL_DIR} && \
     CTEST_OUTPUT_ON_FAILURE=1 make all test
 
+FROM scratch AS binaries
+
+ARG INSTALL_DIR
+
+COPY --from=dagmc_test ${INSTALL_DIR} ${INSTALL_DIR}
```
gonuke, line 269: What files are you copying here and why?  If they aren't installed by `make install` then they probably aren't necessary.

gonuke (changes requested):
One question about installing things in the Dockerfile...

CI/Dockerfile
ahnaf-tahmid-chowdhury: In this step, I am using an empty image to copy all the installed files from the DAGMC step to the exact location. Therefore, in this image, we only contain the files available in /opt/*. There are no extra files or dependencies in this new image, which keeps the image size minimal, less than 90MB.

ahnaf-tahmid-chowdhury: In this step we have, DAGMC, MOAB, HDF5, Double Down, Embree, Geant4. However, We can make it more minimal by copying only `/opt/dagmc` folder. I have chosen to copy the full folder as I am not sure if they are core dependency of DAGMC or only necessary to build DAGMC since we have all the binaries available that works without them.

ahnaf-tahmid-chowdhury: By applying this step, we can update our docker_publish workflow to add a new tag as `dagmc:latest` upon PR merge, and `dagmc:version/stable` upon new release. This can help streamline other projects as well, such as PyNE and NukeLab. We can also update our documentations to use the docker version of DAGMC as well for the general users. Let me know if you find the idea useful.

gonuke:
> Is it necessary to create an extra stage like HDF5 and MOAB, given that these are also external dependencies? I think we can remove these and instead update our workflow to create a binary/final stage to store only the binaries with minimal size, and we may choose the tag as dagmc.

I think these stages mostly appear for historical reasons to limit the amount of rebuilding of images that must occur.  The philosophy is that things that occur earlier in the Dockerfile change least often and least under our control.  Thus, having them installed in early as a different stage can mean fewer instances of rebuilding those sections of the file.  GEANT4 is the most resource intensive to build so one of the main things to avoid.

ahnaf-tahmid-chowdhury:
> I think these stages mostly appear for historical reasons to limit the amount of rebuilding of images that must occur. The philosophy is that things that occur earlier in the Dockerfile change least often and least under our control. Thus, having them installed in early as a different stage can mean fewer instances of rebuilding those sections of the file. GEANT4 is the most resource intensive to build so one of the main things to avoid.

I understand. We've included these stages to minimize the need for rebuilding HDF5 and MOAB whenever we make changes to the Dockerfile. For instance, if we modify the MOAB part, Docker will trace back to the HDF5 stage and then build MOAB. Currently, GitHub workflow supports a full chase log. This means that the steps where we make changes in the Dockerfile will be traced accordingly.

gonuke (approved):

CI/Dockerfile
ahnaf-tahmid-chowdhury: Ah! You've already merged this PR. I was waiting for your response. The binary stage won't run until we update the docker_publish file. There are currently no instructions for this. I think I should create a new PR for it.

CI/Dockerfile
```diff
@@ -125,44 +115,42 @@ RUN if [ "${GEANT4_VERSION}" != "off" ]; then \
...
-    export HDF5_VERSION_major=`python -c "print('.'.join('${HDF5_VERSION}'.split('.')[:-1]))"` && \
-    wget https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-${HDF5_VERSION_major}/hdf5-${HDF5_VERSION}/src/hdf5-${HDF5_VERSION}.tar.gz && \
-    tar -xzf hdf5-${HDF5_VERSION}.tar.gz && \
+RUN mkdir -p ${HDF5_BUILD_DIR}/build && \
+    cd ${HDF5_BUILD_DIR} && \
+    git clone -b ${HDF5_VERSION} --depth 1 https://github.com/HDFGroup/hdf5.git && \
     cd build && \
-    ../hdf5-${HDF5_VERSION}/configure --enable-shared \
-        --prefix=${hdf5_install_dir} \
+    ../hdf5/configure --enable-shared \
```
gonuke, line 139: @ahnaf-tahmid-chowdhury - are you familiar with `configure` and autotools?  The tarball that we retrieve with wget has the `configure` file already generated.  If we pull from the repo, we need to do those steps ourselves.  Do they support a CMake based build?  That might be more generalizable.

ahnaf-tahmid-chowdhury: Not that much as I know cmake. I think autotools are based on autoconf similar to cmake. Handle things in an easier way than cmake.

ahnaf-tahmid-chowdhury: I have noticed we have installed autoconf and libtool thus I haven't tried with cmake for this step. I think I can give this a try. To make it more simplified.

PR #952, opened 2024-03-06
Streamline dependencies of docker CI images Part 2

ahnaf-tahmid-chowdhury:
## Description

This PR aims to streamline the dependencies of Docker CI images by removing autoconf and extra docker stages `moab` and `hdf5`, and merging them into the `external_stage` stage. HDF5 will now be built using CMake instead of autoconf.

## Changes Made

- Removed autoconf dependency.
- Merged `moab` and `hdf5` stages into the `external_stage` stage.
- Built HDF5 using CMake instead of autoconf.

### DIFF
--- a/.github/actions/upstream-test/action.yml
+++ b/.github/actions/upstream-test/action.yml
@@ -59,7 +59,7 @@ runs:
             }}-double_down_${{
               inputs.double_down_version
             }}
-          stages: base, external_deps, hdf5, moab, dagmc
+          stages: base, external_deps, dagmc
           server-stage: dagmc_test
           quiet: false
           parallel: true
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -15,7 +15,6 @@ jobs:
     strategy:
       matrix:
         ubuntu_version : [
-          20.04,
           22.04,
         ]
         compiler : [
@@ -97,7 +96,6 @@ jobs:
     strategy:
       matrix:
         ubuntu_version : [
-          20.04,
           22.04,
         ]
         compiler : [
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -75,7 +75,7 @@ jobs:
         uses: firehed/multistage-docker-build-action@v1
         with:
           repository: ${{ env.image_base_tag }}
-          stages: base, external_deps, hdf5, moab, dagmc
+          stages: base, external_deps, dagmc, final
           server-stage: dagmc_test
           quiet: false
           parallel: true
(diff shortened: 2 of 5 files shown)


ahnaf-tahmid-chowdhury:
Bad luck. It appears that HDF5 requires CMake 3.18 as the minimum version for support in the `find_package` method. However, Ubuntu 20.04 only has CMake 3.16. Any suggestions?

gonuke:
> Bad luck. It appears that HDF5 requires CMake 3.18 as the minimum version for support in the `find_package` method. However, Ubuntu 20.04 only has CMake 3.16. Any suggestions?

Hmmm.... I guess we could drop 20.04 support?  We're only a few weeks away from 24.04 being released....

.github/actions/upstream-test/action.yml
```diff
@@ -59,7 +59,7 @@ runs:
             }}-double_down_${{
               inputs.double_down_version
             }}
-          stages: base, external_deps, hdf5, moab, dagmc
+          stages: base, external_deps, dagmc
           server-stage: dagmc_test
```
ahnaf-tahmid-chowdhury, line 63: How it works? I have noticed it just caching everything. And tests are already passed on the final stage. Maybe I have misconfigured something?

gonuke: What do you expect to be different?

.github/workflows/docker_publish.yml
```diff
@@ -75,7 +74,7 @@ jobs:
         uses: firehed/multistage-docker-build-action@v1
         with:
           repository: ${{ env.image_base_tag }}
-          stages: base, external_deps, hdf5, moab, dagmc
+          stages: base, external_deps, dagmc, final
           server-stage: dagmc_test
```
ahnaf-tahmid-chowdhury, line 78: Sorry, I'd like to ask my previous question here. I believe the final stage could be set as `server-stage`, and `dagmc_test` may be moved to `stages` part.

ahnaf-tahmid-chowdhury: Should I put `dagmc_test` to stages and `final` to server-stage?

gonuke: I don't think it makes a lot of different for our purposes.

ahnaf-tahmid-chowdhury: Then we are good to go.

gonuke (approved):
Thanks @ahnaf-tahmid-chowdhury - this creates a lot of different packages.  At some point we'll want to clean them up and remove the deprecated ones.

ahnaf-tahmid-chowdhury:
Please explain or make an issue. I will try to solve that accordingly.

PR #960, opened 2024-07-19
add more info to Cmake error msg

gonuke:
## Description
Provide more feedback when CMAKE_BUILD_TYPE fatal error occurs

## Motivation and Context
In some build configurations, CMake fails with a fatal error about the CMAKE_BUILD_TYPE, but doesn't indicate what build type has been specified.  

## Changes
This adds the specified build type to the error message.


ahnaf-tahmid-chowdhury (approved):
Thanks for addressing this!

PR #962, opened 2024-10-08
Adding additional output during CMake configuration when double down …

pshriwise:
## Description
Adds some informational messages when double down is enabled in DAGMC for clarity.

## Motivation and Context
Motivated by desire for clarity on whether or not double down will be enabled in the DAGMC build.

### DIFF
--- a/cmake/DAGMC_macros.cmake
+++ b/cmake/DAGMC_macros.cmake
@@ -89,11 +89,13 @@ macro (dagmc_setup_options)
   endif()
 
 if (DOUBLE_DOWN)
+  message(STATUS "DOUBLE_DOWN has been enabled for ray tracing. Searching for package...")
   find_package(DOUBLE_DOWN REQUIRED)
   if (dd_VERSION VERSION_LESS 1.1.0)
     message(FATAL_ERROR "Discovered Double Down Version: ${DOUBLE_DOWN_VERSION}. \
     Please update Double Down to version 1.1.0 or greater.")
   endif()
+  message(STATUS "Found DOUBLE_DOWN.")
 endif()
 
 
--- a/doc/CHANGELOG.rst
+++ b/doc/CHANGELOG.rst
@@ -9,6 +9,7 @@ Next version
 
 **Changed:**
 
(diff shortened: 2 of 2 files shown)


gonuke:
Does Double Down provide a a config file for CMake to find? or do we use our won `FindDOUBLE_DOWN.cmake` file?  If the latter, it already includes extra output like this.

It's not in the version that @bquan0 is trying to update on conda-forge, but would be part of the next release.

gonuke:
> Does Double Down provide a a config file for CMake to find? or do we use our won `FindDOUBLE_DOWN.cmake` file? If the latter, it already includes extra output like this.
> 
> It's not in the version that @bquan0 is trying to update on conda-forge, but would be part of the next release.

Hmmm... but maybe that's not working either....??

gonuke:
It was working - as seen in [this build from 3 days ago](https://github.com/svalinn/DAGMC/actions/runs/11196912771/job/31126278213#step:4:41).

Your change adds more as seen in [this build from this PR](https://github.com/svalinn/DAGMC/actions/runs/11239571075/job/31246935095?pr=962#step:4:41).

Can't hurt to have additional messages, but there was already this info.

The reason that the conda-forge build isn't seeing it is because this output wasn't in the right version.

pshriwise:
> It was working - as seen in [this build from 3 days ago](https://github.com/svalinn/DAGMC/actions/runs/11196912771/job/31126278213#step:4:41).
> 
> Your change adds more as seen in [this build from this PR](https://github.com/svalinn/DAGMC/actions/runs/11239571075/job/31246935095?pr=962#step:4:41).
> 
> Can't hurt to have additional messages, but there was already this info.
> 
> The reason that the conda-forge build isn't seeing it is because this output wasn't in the right version.

Agreed, there was some trace that the `DOUBLE_DOWN` option was respected. I think it's a little more clear now as someone can search for the same option presented in DAGMC's CMakeLists.txt in the output rather than trying to find `Found dd` (perhaps my fault for an incongruous package name)

gonuke (approved):
Thanks for this extra clarity @pshriwise

PR #964, opened 2024-11-11
Adding moab download if not found via fetch content

shimwell:
## Description
To help with [this](https://github.com/openmc-dev/openmc/pull/3193) PR on the openmc repo we would need DAGMC to fetch MOAB if it is not automatically found. Just making this PR to show the idea, this won't currently work as I think changes in MOAB are needed

## Motivation and Context
This could ultimately allow openmc to be installed with dagmc without the user having to specify dirs for moab and dagmc and install them prior to the openmc compile.

## Changes
changes to the cmakelists.txt

## Behavior
current: if moab is not found cmake command terminals
with this PR: if moab not found moab gets downloaded

## Changelog file

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,12 +59,12 @@ if(BUILD_MCNP5 OR BUILD_MCNP6)
   enable_language(Fortran)
 endif()
 
-include(FetchContent)
 
 # Try to find MOAB
 # find_package(MOAB QUIET)
 
 if (NOT MOAB_FOUND)
+  include(FetchContent)
   message(STATUS "MOAB not found. Fetching MOAB...")
   FetchContent_Declare(
     moab
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ if (NOT MOAB_FOUND)
   FetchContent_Declare(
     moab
     GIT_REPOSITORY https://bitbucket.org/fathomteam/moab.git
-    GIT BRANCH master
+    GIT_BRANCH master
     # GIT_TAG        5.5.1
   )
   FetchContent_MakeAvailable(moab)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,14 +66,19 @@ if (NOT MOAB_FOUND)
   message(STATUS "MOAB not found. Fetching MOAB...")
   FetchContent_Declare(
     moab
-    GIT_REPOSITORY https://bitbucket.org/fathomteam/moab.git
-    GIT_BRANCH master
+    # making use of branch on fork as it has a one line change that help this fetch command work
+    # GIT_REPOSITORY https://bitbucket.org/fathomteam/moab.git
+    # GIT_BRANCH master
     # GIT_TAG        5.5.1
+    GIT_REPOSITORY https://bitbucket.org/moab-fork/moab.git
+    GIT_TAG Jonathan-Shimwell/adding-current-to-dir
+    GIT_SHALLOW TRUE
   )
+  
   FetchContent_MakeAvailable(moab)
-  add_subdirectory(${moab_SOURCE_DIR} ${moab_BINARY_DIR}) # MOAB repository is not properly configured for a FetchContent build
+  add_subdirectory(${moab_SOURCE_DIR} ${moab_BINARY_DIR}/moab-build)
   # Set MOAB_DIR to the fetched content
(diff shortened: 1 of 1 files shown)


gonuke:
We'd probably also want to do the same with Embree and double down?

shimwell:
Yes, keen to add them,

they would fist have to check that double down is enabled and if so then fetch the contents. Also they might need changes to the repos.

So perhaps best to keep the PR minimal and see if we can get something simple workign first. To get this working I think we need some changes in moab

pshriwise:
Q: If MOAB is fetched,

pshriwise:
Ah sorry! Mis-clicked and closed this momentarily.

makeclean:
Double down will be fun due to the double-down -> embree -> OneTbb

ahnaf-tahmid-chowdhury:
I am not a fan of using FetchContent; instead, if possible, I prefer using [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html), as it allows us to build MOAB with HDF5 support.

shimwell:
> I am not a fan of using FetchContent; instead, if possible, I prefer using [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html), as it allows us to build MOAB with HDF5 support.

Interesting, I shall make a draft PR for ExternalProject and we can compare the two approaches.

I though hdf5 was automatically brought into moab after [this PR](https://bitbucket.org/fathomteam/moab/pull-requests/706) so perhaps this is not a problem anymore (after the next moab release)?

ahnaf-tahmid-chowdhury:
I have noticed that current [master branch](https://bitbucket.org/fathomteam/moab/src/910fd0a9ec70a6ed7205a50e71f7be969d8632e5/CMakeLists.txt#lines-116) doesn't have HDF5 enabled by default.

CMakeLists.txt
```diff
@@ -59,7 +59,24 @@ if(BUILD_MCNP5 OR BUILD_MCNP6)
...
+
+if (NOT MOAB_FOUND)
+  include(FetchContent)
+  message(STATUS "MOAB not found. Fetching MOAB...")
+  FetchContent_Declare(
+    moab
+    GIT_REPOSITORY https://bitbucket.org/fathomteam/moab.git
+    GIT_BRANCH master
+    # GIT_TAG        5.5.1
+  )
```
shimwell, line 76: How about we add HDF5 and other cmake args like this and then we can still use fetch
```suggestion
    CMAKE_ARGS
      -DENABLE_HDF5=ON
      -DENABLE_NETCDF=ON
      -DBUILD_SHARED_LIBS=OFF
      -DENABLE_FORTRAN=OFF
      -DENABLE_BLASLAPACK=OFF
  )
```

ahnaf-tahmid-chowdhury: I am not sure if it's a vaild command for FetchContent_Declare, but it is available in ExternalProject_Add. For more: https://discourse.cmake.org/t/fetchcontent-declare-with-cmake-args-does-not-work/11227/2

ahnaf-tahmid-chowdhury:
I think, if you don't mind, I should consider creating a PR with ExternalProject_Add since I have some previous experience with that. Or, You may find how I did this on [PyNE](https://github.com/pyne/pyne/blob/scikit-build-core/CMakeLists.txt#L221).

shimwell:
> I think, if you don't mind, I should consider creating a PR with ExternalProject_Add since I have some previous experience with that. Or, You may find how I did this on [PyNE](https://github.com/pyne/pyne/blob/scikit-build-core/CMakeLists.txt#L221).

Sounds good to me, I think eitherway (fetch or external) we will bump into [this](https://bitbucket.org/fathomteam/moab/issues/185/request-for-moab-compatible-with) issue.

shimwell:
I've made a small change to the MOAB cmakelists.txt on a fork and making use of that fork for now in this fetch command.

This change gets further through the compile 

The new error starts with ...

```
CMake Error at build/_deps/moab-src/src/io/CMakeLists.txt:111 (add_library):
  add_library cannot create target "moabio" because another target with the
  same name already exists.  The existing target is created in source
  directory "/home/jon/DAGMC/build/_deps/moab-src/src/io".  See documentation
  for policy CMP0002 for more details.

...
```

shimwell:
@bam241 made a new PR to replace this one, closing in favor of #969

PR #965, opened 2024-11-11
added options to the cmakelists.txt file

jon-proximafusion:
## Description
Adding options to the cmakelists.txt, I think this helps show the available options.

## Motivation and Context
hard to find out all the options without this

## Behavior
code should compile in the same way

## Changelog file


makeclean:
These options are already available in DAGMC_macros.cmake, I don't understand why you want to repeat them? You can always find available options by examining the results of `ccmake ..` when run from the bld directory for example.

shimwell:
Apologies I did not notice that file in the cmake subfolder. Closing as unnecessary

PR #969, opened 2024-12-04
Automatic download and build of MOAB 

bam241:
## Description
Add capability to automate pull and build MOAB at DAGMC building time 

This is redondant with #964 

## Motivation and Context
This is part of the project from @shimwell on getting a pip install for OpenMC with DAGMC.

## Changes
tweaked a little bit the `findMOAB.cmake` 
- add a make files that contains the proper ExternalProject configuration
- add a CMAKE variable to enable pull and build and change the dips and link for the other DAGMC library

## Behavior
Nothing changes if DDL_DEPS is not set at ON on Cmake configuration.
Otherwise pull build MOAB from MOAB repo builds it, and build the DAGMC lib against it.

## Changelog file
- [x] Done

This is very much inspired by @ahnaf-tahmid-chowdhury [work on PyNE cmake](https://github.com/pyne/pyne/blob/183cd7fdc91c0b1e93c6e1bd906964baed344bf4/CMakeLists.txt#L221).

_This works is sponsored by [Proxima Fusion](https://www.proximafusion.com/)_


bam241:
@ahnaf-tahmid-chowdhury @shimwell @gonuke even if this is still a draft, feel free to provide any feed back!

src/dagmc/CMakeLists.txt
```diff
@@ -13,7 +13,8 @@ set(LINK_LIBS_EXTERN_NAMES MOAB_LIBRARIES HDF5_LIBRARIES)
 include_directories(${CMAKE_BINARY_DIR}/src/dagmc)
 
 dagmc_install_library(dagmc)
-
```
bam241, line 16: The plan is to add conditional there, to do that only if MOAB source have been downloaded and set for compilation

src/pyne/CMakeLists.txt
```diff
@@ -34,3 +34,4 @@ if(NOT (${HDF5_VERSION} VERSION_LESS 1.12.0))
  endif()
 
 dagmc_install_library(pyne_dagmc)
+add_dependencies(pyne_dagmc-shared MOAB)
```
bam241: Same here !
This should be conditional

src/dagmc/CMakeLists.txt
```diff
@@ -13,7 +13,8 @@ set(LINK_LIBS_EXTERN_NAMES MOAB_LIBRARIES HDF5_LIBRARIES)
...
+add_dependencies(dagmc-shared MOAB)
+target_link_libraries(dagmc-shared PUBLIC ${MOAB_INSTALL_PREFIX}/lib64/libMOAB.so)
```
bam241: The plan is to add conditional there, to do that only if MOAB source have been downloaded and set for compilation

cmake/FindMOAB.cmake
```diff
@@ -74,8 +119,8 @@ message(STATUS "MOAB_LIBRARIES_STATIC: ${MOAB_LIBRARIES_STATIC}")
 if (MOAB_INCLUDE_DIRS AND (MOAB_LIBRARIES_SHARED OR NOT BUILD_SHARED_LIBS) AND
     (MOAB_LIBRARIES_STATIC OR NOT BUILD_STATIC_LIBS))
   message(STATUS "Found MOAB")
-else ()
-  message(FATAL_ERROR "Could not find MOAB")
+else()
+  # message(FATAL_ERROR "Could not find MOAB")
```
bam241: This should still return a fatal_error if `ExternalProject` encounters an issue

bam241:
The safeguard I added should be improve... (i.e.: have them only MOAB download is triggered...)

cmake/FindMOAB.cmake
```diff
@@ -13,12 +13,57 @@ find_path(MOAB_CMAKE_CONFIG
 if (MOAB_CMAKE_CONFIG)
   set(MOAB_CMAKE_CONFIG ${MOAB_CMAKE_CONFIG}/MOABConfig.cmake)
   message(STATUS "MOAB_CMAKE_CONFIG: ${MOAB_CMAKE_CONFIG}")
+  include(${MOAB_CMAKE_CONFIG})
+elseif (DDL_INSTALL_DEPS)
+  message(STATUS "MOAB will be downloaded and built")
+  include(ExternalProject)
+  # Configure MOAB
+  if(NOT MOAB_VERSION)
+    set(MOAB_VERSION "5.5.1")
```
shimwell:
```suggestion
  include(ExternalProject)
   # Configure MOAB
  if(NOT MOAB_VERSION)
    set(MOAB_VERSION "5.5.1")
  message(STATUS "MOAB version ${MOAB_VERSION} will be downloaded and built")
```

shimwell: Perhaps if we change the order of this part of code the user message can then include the version of MOAB that is going to be automatically installed.

cmake/FindMOAB.cmake
```diff
@@ -13,12 +13,57 @@ find_path(MOAB_CMAKE_CONFIG
...
   set(MOAB_CMAKE_CONFIG ${MOAB_CMAKE_CONFIG}/MOABConfig.cmake)
   message(STATUS "MOAB_CMAKE_CONFIG: ${MOAB_CMAKE_CONFIG}")
+  include(${MOAB_CMAKE_CONFIG})
+elseif (DDL_INSTALL_DEPS)
+  include(ExternalProject)
+   # Configure MOAB
+  if(NOT MOAB_VERSION)
+    set(MOAB_VERSION "5.5.1")
+  message(STATUS "MOAB version ${MOAB_VERSION} will be downloaded and built")
+  endif()
```
shimwell:
```suggestion
  if(NOT MOAB_VERSION)
    set(MOAB_VERSION "5.5.1")
  endif()
  message(STATUS "MOAB version ${MOAB_VERSION} will be downloaded and built")

```

shimwell: Sorry I think I missed the endif in the earlier suggestion

bam241:
@gonuke this still needs some polishing here and there, but I think this is ready for a first opinion.

cmake/DAGMC_macros.cmake
```diff
@@ -253,6 +253,11 @@ macro (dagmc_install_library lib_name)
             EXPORT DAGMCTargets
             LIBRARY DESTINATION ${INSTALL_LIB_DIR}
             PUBLIC_HEADER DESTINATION ${INSTALL_INCLUDE_DIR})
+    # Required to ensure that MOAB is built before DAGMC and to properly link against MOAB
+    if(DDL_INSTALL_DEPS)
+      add_dependencies(${lib_name}-shared MOAB)
+      target_link_libraries(${lib_name}-shared PUBLIC ${MOAB_LIBRARY_DIRS}/libMOAB.so)
```
ahnaf-tahmid-chowdhury: It seems it will only work on linux. We should consider other platform as well.

cmake/MOAB_PullAndMake.cmake
```diff
@@ -0,0 +1,44 @@
...
+        -DENABLE_PYMOAB:BOOL=OFF 
+      DOWNLOAD_EXTRACT_TIMESTAMP true
+      BUILD_BYPRODUCTS "${MOAB_LIBRARY_DIRS}/*${CMAKE_SHARED_LIBRARY_SUFFIX}*"
+      INSTALL_DIR "${MOAB_INSTALL_PREFIX}"
+    )
+    # Setup a interface library for MOAB based on ExternalProoject MOAB_EP
+    add_library(MOAB INTERFACE)
+
+    target_include_directories(MOAB SYSTEM INTERFACE ${MOAB_INCLUDE_DIRS})
+    target_link_libraries(MOAB INTERFACE ${MOAB_LIBRARY_DIRS}/libMOAB.so)
```
ahnaf-tahmid-chowdhury: This is also platform specific.

cmake/MOAB_PullAndMake.cmake
```diff
@@ -0,0 +1,44 @@
...
+    ExternalProject_Add(MOAB_ep
+      PREFIX ${MOAB_ROOT}   
+      GIT_REPOSITORY https://bitbucket.org/fathomteam/moab.git
+      GIT_TAG ${moab_version}
+      CMAKE_ARGS
+        -DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR>
+        -DBUILD_SHARED_LIBS:BOOL=ON
+        -DENABLE_HDF5:BOOL=ON
+        -DHDF5_ROOT:PATH=${HDF5_ROOT}
+        -DCMAKE_INSTALL_RPATH=${HDF5_ROOT}/lib:${MOAB_INSTALL_PREFIX}/lib
```
ahnaf-tahmid-chowdhury: This method may work here for now, but may create trouble while working with scikit-build-core depended project. However, I am not sure here.

.github/workflows/linux_build_test.yml
```diff
@@ -50,6 +50,9 @@ jobs:
           5.4.1,
           5.5.1,
         ]
+        ddl_deps : [
```
ahnaf-tahmid-chowdhury: What is the use of it?

ahnaf-tahmid-chowdhury (commented):
Thank you very much for the update. I have a few comments for improvement.

.github/workflows/linux_build_test.yml
ahnaf-tahmid-chowdhury: This is like Enable Download MOAB or Double Down?

bam241: this only enables MOAB Download, otherwise everything work as it used to.

bam241: I don't have a strong opinion about the name of the variable, 
but eventually one might want to include `Embree` and `DD` the same way...

if your question is about the addition into the matrix, I needed it to allow the CI to test for this kind of build...

cmake/MOAB_PullAndMake.cmake
bam241: I have no experience about `scikit-build-core`, maybe we shall fix it, if the problem occurs ?

shimwell: I believe MOAB master branch now has scikit-build core while version 5.5.1 does not.

Perhaps changing the moab version to master would provide a quick way of checking if this continues to work
```
      GIT_REPOSITORY https://bitbucket.org/fathomteam/moab.git
      GIT_TAG ${moab_version}
```

bam241: Trying it now.

.github/workflows/linux_build_test.yml
bam241: this seems to [pass](https://github.com/svalinn/DAGMC/actions/runs/12199530439/job/34033629285), I'll revert the change

jon-proximafusion:
Testing locally with this docker file and it compiles for me
Note this points to Bam's fork and this PR branch
Also note, this example is very minimal with no DD, Embree, Pyne

```
FROM ubuntu:24.04

RUN apt-get update -y
RUN apt-get install git cmake libhdf5-serial-dev g++ libeigen3-dev -y 

RUN mkdir DAGMC && \
	cd DAGMC && \
	git clone --single-branch --branch auto_moab --depth 1 https://github.com/bam241/DAGMC.git && \
	mkdir build && \
	cd build && \
	cmake ../DAGMC -DBUILD_TALLY=ON \
                   -DBUILD_STATIC_EXE=OFF \
                   -DBUILD_STATIC_LIBS=OFF \
                   -DCMAKE_INSTALL_PREFIX=/DAGMC/ \
                   -DBUILD_UWUW=OFF \
                   -DPULL_INSTALL_MOAB=ON && \
	make -j install
```

cmake/MOAB_PullAndMake.cmake
```diff
@@ -0,0 +1,45 @@
...
+MACRO (moab_pull_make moab_version)
+    message(STATUS "MOAB will be downloaded and built")
+    include(ExternalProject)
+    message("HDF5_ROOT: ${HDF5_ROOT}")
+    SET(MOAB_INSTALL_PREFIX "${CMAKE_INSTALL_PREFIX}/moab")
+    set(MOAB_ROOT "${CMAKE_BINARY_DIR}/moab")
+    set(MOAB_INCLUDE_DIRS "${MOAB_INSTALL_PREFIX}/include")
+    set(MOAB_LIBRARY_DIRS "${MOAB_INSTALL_PREFIX}/lib")
+    message("MOAB_LIBRARY_DIRS: ${MOAB_LIBRARY_DIRS}")
+    MEsSAGE("CMAKE_SHARED_LIBRARY_SUFFIX: ${CMAKE_SHARED_LIBRARY_SUFFIX}")
```
jon-proximafusion: This the uppercase lower case a typo here
```suggestion
    message("CMAKE_SHARED_LIBRARY_SUFFIX: ${CMAKE_SHARED_LIBRARY_SUFFIX}")
```

cmake/MOAB_PullAndMake.cmake
```diff
@@ -0,0 +1,45 @@
+# this Macro sets up the download and build of MOAB using ExternalProject
+# few tweak are done in src/dagmc/CMakeLists.txt and src/PyNE/CMakelists.txt 
+# to make sure that MOAB is built before DAGMC.
+MACRO (moab_pull_make moab_version)
+    message(STATUS "MOAB will be downloaded and built")
+    include(ExternalProject)
+    message("HDF5_ROOT: ${HDF5_ROOT}")
+    SET(MOAB_INSTALL_PREFIX "${CMAKE_INSTALL_PREFIX}/moab")
```
jon-proximafusion: should this be lower case like the others
```suggestion
    set(MOAB_INSTALL_PREFIX "${CMAKE_INSTALL_PREFIX}/moab")
```

cmake/MOAB_PullAndMake.cmake
jon-proximafusion: Just to mention @bam241 tried this with the master branch of MOAB (which has scikit build core) and reported that it works

cmake/MOAB_PullAndMake.cmake
bam241: You are probably right, I added your suggestions.

I usually prefer to keep `CMake` instructions in all caps, this explains the confusion. and the strange `MEsSAGE`

cmake/MOAB_PullAndMake.cmake
bam241: have a doubt about what I did, I'll try one more time

bam241:
[Test](https://github.com/svalinn/DAGMC/actions/runs/12295008383/job/34310994000?pr=969) on [5fe643b](https://github.com/svalinn/DAGMC/pull/969/commits/5fe643b9329349f7690e5e8ea1d1aa7ce3667721)
checked the proper work of this against MOAB Master branch

jon-proximafusion:
IMO this PR makes building DAGMC easier for users. Currently we ask users to compile MOAB and then compile DAGMC passing some of the info on MOAB to the DAGMC compile. I believe the current situation is overly complicated.

This PR simplifies the compile for users which I think will be appreciated and lowers the bar slightly for installing DAGMC which might help bring in a few more users.

Screenshot From 2025-01-09 17-35-25

Additionally it retains backwards compatibility so if someone does what to compile both MOAB and DAGMC in the current way then they can still do that.

The main change I would suggest is that we make the ```pull_install_moab``` variable accept a version number of MOAB and default to the current latest. This would offer greater user control over the automated download. However that could be a follow up PR if desired.

*update this also now allows users to specific the version of MOAB automatically downloaded

CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 project(DAGMC)
-cmake_minimum_required(VERSION 3.18)
+cmake_minimum_required(VERSION 3.1)
```
gonuke: I think the rebase has caused this regression - we should still with 3.18

cmake/FindMOAB.cmake
gonuke, line 1: Since we're including and running another macro here now, can we add a comment at the top of this that lists all the variables that expect to be set once this FindMOAB is complete?  Then we can make sure the same/similar comment is at the top of that new macro.

cmake/FindMOAB.cmake
```diff
@@ -10,15 +10,28 @@ find_path(MOAB_CMAKE_CONFIG
   PATHS ${MOAB_SEARCH_DIRS}
   NO_DEFAULT_PATH
 )
-if (MOAB_CMAKE_CONFIG)
+
+# First check if we are forcing the download of MOAB
+if (PULL_INSTALL_MOAB)
```
gonuke: Have we thought about whether we put this condition up one level?  That is, if we are going to PULL_INSTALL, then we don't call `FindPackage` at all and just do this macro instead?  Then we don't need to mess with the `FindMOAB` file.  I think that will work???

gonuke (changes requested):
A few comments trying to streamline and localize these changes.

src/mcnp/CMakeLists.txt
```diff
@@ -18,6 +18,11 @@ add_library(mcnp_funcs OBJECT mcnp_funcs.cpp)
 message(STATUS "Building object library: meshtal_funcs")
 add_library(meshtal_funcs OBJECT meshtal_funcs.cpp)
 
+if(PULL_INSTALL_MOAB)
```
pshriwise, line 21: I'm curious as to why these lines are needed now when they weren't before.

cmake/MOAB_PullAndMake.cmake
pshriwise: Any final resolution on this?

pshriwise (dismissed):
I'm nominally on board here and want to make the installation process easier, but I do have concerns. There are secondary dependencies to consider and how we might specify those to the direct dependencies of MOAB. For example, there's a requirement here for an `HDF5_ROOT` that's expected to be set prior to the CMake call (from what I can tell). 

I think what's here really would make the installation process better for all, but I do want to think through these scenarios thoroughly.

pshriwise (commented):
Sorry, meant to comment not approve.. yet 😄

bam241:
💔

cmake/MOAB_PullAndMake.cmake
bam241: I have tested this against MOAB master, this works.

[Test](https://github.com/svalinn/DAGMC/actions/runs/12295008383/job/34310994000?pr=969) on [5fe643b](https://github.com/svalinn/DAGMC/pull/969/commits/5fe643b9329349f7690e5e8ea1d1aa7ce3667721)
checked the proper work of this against MOAB Master branch

makeclean:
IMO you should also add `-DENABLE_FORTRAN=OFF`, `-DENABLE_BLASLAPCK=OFF` you might need to pass an EIGEN3_DIR or something

src/mcnp/CMakeLists.txt
bam241: The goal is to allow if asked for, the possibility to pull and compile MOAB at the DAGMC build time 

This variable has been added to detect if such auto download and compilation of MOAB is requested.

bam241:
@gonuke, I moved this 1 level.

I added a `HDF5_macor.cmake` file that factories HDF5 search and variables definition.

bam241:
@gonuke @pshriwise @makeclean I think I have addressed most(all?) your comments.

let me know if you have any additional :)

src/mcnp/CMakeLists.txt
gonuke: I think @pshriwise has a different question: why does this method explicitly require adding the dependencies?  As I recall from a conversation last month, since the MOAB libraries are not built until build time, CMake can't add them implicitly.  They are not found by CMake and added to the linked library list, so we have to tell it to happend.

cmake/MOAB_PullAndMake.cmake
```diff
@@ -0,0 +1,93 @@
...
+  
+  # Ensure STATIC LIBS Build
+  IF(DAGMC_BUILD_STATIC_LIBS)
+    MESSAGE(FATAL_ERROR "PULL_INSTALL_MOAB is ONLY compatible with shared libraries.")
+  endIF()
+  
+  # Ensure MOAB_VERSION
+  IF(NOT MOAB_VERSION)
+    SET(MOAB_VERSION "5.5.1")
+  endIF()
```
gonuke: It might be useful to do this at the top level so this pinned version string is more obvious if/when it needs to be changed.

gonuke (changes requested):
This is definitely a lot cleaner and more modular @bam241 - thanks!

bam241:
> This is definitely a lot cleaner and more modular @bam241 - thanks!

NP, I do agree this is cleaner !
do you have anything else ?

CMakeLists.txt
```diff
@@ -44,7 +44,18 @@ if(BUILD_MCNP5 OR BUILD_MCNP6)
   enable_language(Fortran)
 endif()
 
-find_package(MOAB REQUIRED)
+# Dealing with MOAB
+if (PULL_INSTALL_MOAB)
```
gonuke, line 48: I think we need to declare this variable somewhere to define it's default

gonuke (changes requested):
One request for initializing this new variable.

cmake/MOAB_PullAndMake.cmake
```diff
@@ -0,0 +1,88 @@
...
+  TARGET_INCLUDE_DIRECTORIES(MOAB SYSTEM INTERFACE ${MOAB_INCLUDE_DIRS})
+  TARGET_LINK_LIBRARIES(MOAB INTERFACE ${MOAB_LIBRARY_DIRS}/libMOAB${CMAKE_SHARED_LIBRARY_SUFFIX})
+  ADD_DEPENDENCIES(MOAB MOAB_ep)
+  INSTALL(TARGETS MOAB LIBRARY DESTINATION ${MOAB_LIBRARY_DIRS}
+                      PUBLIC_HEADER DESTINATION ${INSTALL_INCLUDE_DIR})
+  INCLUDE_DIRECTORIES(${MOAB_INCLUDE_DIRS})
+  LINK_DIRECTORIES(${MOAB_LIBRARY_DIRS})
+  
+  # FIND_PACKAGE(Eigen3 REQUIRED NO_MODULE)
+  # INCLUDE_DIRECTORIES(${EIGEN3_INCLUDE_DIRS})
```
pshriwise: Are these inherited lines? If they're commented perhaps they can go?

CMakeLists.txt
bam241: it is already in the `DAGMC_macros.cmake` file

gonuke: and possibly add some documentation that CMake can report

shimwell:
The main concern I have is that I'm still keen to not have a boolean value for downloading or not. I'm keen to pass in the moab version as a string like "5.5.1"

gonuke:
> The main concern I have is that I'm still keen to not have a boolean value for downloading or not. I'm keen to pass in the moab version as a string like "5.5.1"

What if we had `PULL_INSTALL_MOAB=5.5.1` ?

gonuke (dismissed):

pshriwise:
We should update the documentation to make users aware of this option.

cmake/MOAB_PullAndMake.cmake
```diff
@@ -0,0 +1,85 @@
...
+  SET(MOAB_INCLUDE_DIRS "${MOAB_INSTALL_PREFIX}/include")
+  SET(MOAB_LIBRARY_DIRS "${MOAB_INSTALL_PREFIX}/lib")
+  MESSAGE("MOAB_LIBRARY_DIRS: ${MOAB_LIBRARY_DIRS}")
+  MESSAGE("CMAKE_SHARED_LIBRARY_SUFFIX: ${CMAKE_SHARED_LIBRARY_SUFFIX}")
+  SET(MOAB_LIBRARIES_SHARED "")
+  ExternalProject_Add(MOAB_ep
+    PREFIX ${MOAB_ROOT}   
+    GIT_REPOSITORY https://bitbucket.org/fathomteam/moab.git
+    GIT_TAG ${moab_version}
+    CMAKE_ARGS
```
pshriwise, line 30: This is the argument I mentioned on the call today. It looks like they do now set that value, but for ideal optimization it would be set to release. Often libraries include debug symbols in the default build for testing consistency.

https://bitbucket.org/fathomteam/moab/src/master/CMakeLists.txt


For this PR I think we should match the `BUILD_TYPE` to the setting for the DAGMC build.

shimwell:
> > The main concern I have is that I'm still keen to not have a boolean value for downloading or not. I'm keen to pass in the moab version as a string like "5.5.1"
> 
> What if we had `PULL_INSTALL_MOAB=5.5.1` ?

Yes this would be great.
This suggestion would also mean we don't have to update the code when a new version of moab comes out

cmake/MOAB_PullAndMake.cmake
bam241: will do that

bam241:
@pshriwise I added the missing configuration flag for MOAB installation, and ordered them as it is done in the DockerFile.
I also added documentation about the new flag.

@jon-proximafusion @shimwell I changed the flag from BOOLEAN (ON/OFF) to a BOOLEAN with value:[see [cmake option](https://cmake.org/cmake/help/latest/command/option.html)]

So new the behaviour is the following:
- if `PULL_INSTALL_MOAB` is not defined, value is `OFF`, and `cmake` try to find `MOAB`
- if `PULL_INSTALL_MOAB` is defined as `ON`, then `MOAB_VERSION` is set to the default "5.5.1"
- if `PULL_INSTALL_MOAB` is defined as "x.y.z" then `MOAB_VERSION` is set to "x.y.z"

shimwell:
> if PULL_INSTALL_MOAB is defined as ON, then MOAB_VERSION is set to the default "5.5.1"

I might be tempted not to allow "ON" as an option. What do others think?
Reasoning for this suggestion is that:
- It makes the command more opaque, users won't know what they are getting
- It adds a maintenance burden as we should keep this tag up to date with latest MOAB release

I might be tempted to use 'latest' or 'develop' as options by that is another topic.

cmake/MOAB_PullAndMake.cmake
```diff
@@ -0,0 +1,86 @@
...
+      -DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR>
+      -DCMAKE_BUILD_TYPE=Release
```
pshriwise: I think it makes sense for us to pass the same value for the build type that was provided for DAGMC. If one is looking to debug a problem with DAGMC, they'd likely want MOAB to be build in debug mode as well.

```suggestion
      -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
```

This raises perhaps another question though: What will happen if DAGMC is recompiled with a new value of `CMAKE_BUILD_TYPE` will the MOAB build reflect that change?

cmake/MOAB_PullAndMake.cmake
```diff
@@ -0,0 +1,86 @@
...
+      -DENABLE_HDF5:BOOL=ON
+      -DHDF5_ROOT:PATH=${HDF5_ROOT}
+      -DCMAKE_INSTALL_RPATH=${HDF5_ROOT}/lib:${MOAB_INSTALL_PREFIX}/lib
+      -DENABLE_BLASLAPACK:BOOL=OFF
+      -DBUILD_SHARED_LIBS:BOOL=ON
```
pshriwise, line 37: There's an option to build both static and shared libraries with DAGMC. To build with this option enabled tn the past, I've had to build MOAB twice -- once with this value set to `ON` and again with it set to `OFF`.

I don't think this is a common use case these days though and I'll suggest that, for now, if `PULL_INSTALL_MOAB` is on that we check if static libraries are requested and error with a notification that these settings (`PULL_INSTALL_MOAB=ON` and `BUILD_STATIC_LIBS=ON`) aren't compatible.

bam241: there should already be a safeguard for this case l9-10 of this file.

pshriwise: Ah, right you are. Thanks!

cmake/MOAB_PullAndMake.cmake
pshriwise: This was marked as resolved but I don't see an update to the line above. Am I missing something?

pshriwise: PR was out of date (though no `Refresh` was apparent on the page). Mea culpa. Disregard.

cmake/MOAB_PullAndMake.cmake
```diff
@@ -0,0 +1,86 @@
...
+    MESSAGE(FATAL_ERROR "PULL_INSTALL_MOAB is ONLY compatible with shared libraries.")
+  endIF()
+    
+  # Ensure we have the right variables to build MOAB
+  moab_autobuild_check_deps()
+   
+    
+  INCLUDE(ExternalProject)
+  MESSAGE("HDF5_ROOT: ${HDF5_ROOT}")
+  SET(MOAB_INSTALL_PREFIX "${CMAKE_INSTALL_PREFIX}/moab")
```
ahnaf-tahmid-chowdhury, line 18: It seems MOAB is going to be installed in a custom path, which means only DAGMC will have access to it, as it will use the rpath. This is a good approach, and we can apply the same strategy for HDF5 as well. However, I think we can use the same `CMAKE_INSTALL_PREFIX` for both DAGMC and MOAB instead of `CMAKE_INSTALL_PREFIX}/moab`, as there are no conflicts having the same file name.

bam241: I did put the MOAB installation there, because I though about scenarii where users would already have MOAB installed on it computer but would try an installation of DAGMC against a different version of MOAB.
To avoid conflict, I rather have MOAB installed in the DAGMC install dir, as it is installed through DAGMC.
Also if one would come to uninstall DAGMC by removing the install dir, it would make sense to remove MOAB at the same time.

rherrero-pf:
Really like the idea behind this feature so I though I should give it a try. Happy to simplify the install process

Running this and I think the autodownloading of moab assumes that moab is available in the path ``` /DAGMC/moab``` which in my case is not true.

```
rm -rf ~/code_development/DAGMC
mkdir ~/code_development/DAGMC
cd ~/code_development/DAGMC
git clone --single-branch --branch auto_moab --depth 1 https://github.com/bam241/DAGMC.git
mkdir build
cd build
cmake ../DAGMC -DBUILD_TALLY=ON \
                -DBUILD_STATIC_EXE=OFF \
                -DBUILD_STATIC_LIBS=OFF \
                -DCMAKE_INSTALL_PREFIX=/DAGMC/ \
                -DBUILD_UWUW=OFF \
                -DPULL_INSTALL_MOAB=ON \
                -DHDF5_ROOT=/usr/lib/x86_64-linux-gnu/hdf5
make -j install
```

```
~/code_development$ bash i.sh 
Cloning into 'DAGMC'...
remote: Enumerating objects: 453, done.
remote: Counting objects: 100% (453/453), done.
remote: Compressing objects: 100% (422/422), done.
remote: Total 453 (delta 42), reused 177 (delta 19), pack-reused 0 (from 0)
Receiving objects: 100% (453/453), 18.58 MiB | 21.03 MiB/s, done.
Resolving deltas: 100% (42/42), done.
-- The C compiler identification is GNU 11.4.0
-- The CXX compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Git: /usr/bin/git (found version "2.34.1") 

-- CMAKE_BUILD_TYPE not specified, defaulting to Release
-- CMAKE_BUILD_TYPE: Release

NO MOAB_VERSION specified. Defaulting to 5.5.1
MOAB_VERSION to be installed: 5.5.1
-- MOAB will be downloaded and built
HDF5_ROOT: /usr/lib/x86_64-linux-gnu/hdf5
MOAB_LIBRARY_DIRS: /DAGMC/moab/lib
CMAKE_SHARED_LIBRARY_SUFFIX: .so
CMake Error at /usr/share/cmake-3.22/Modules/ExternalProject.cmake:1754 (file):
  file problem creating directory: /DAGMC/moab
  /usr/share/cmake-3.22/Modules/ExternalProject.cmake:3668 (_ep_set_directories)
  cmake/MOAB_PullAndMake.cmake:25 (ExternalProject_Add)
  CMakeLists.txt:58 (moab_pull_make)


CMake Error at /usr/share/cmake-3.22/Modules/ExternalProject.cmake:1756 (message):
  dir '/DAGMC/moab' does not exist after file(MAKE_DIRECTORY)
  /usr/share/cmake-3.22/Modules/ExternalProject.cmake:3668 (_ep_set_directories)
  cmake/MOAB_PullAndMake.cmake:25 (ExternalProject_Add)
  CMakeLists.txt:58 (moab_pull_make)


See also "/home/rherrero/code_development/DAGMC/build/CMakeFiles/CMakeOutput.log".
make: *** No rule to make target 'install'.  Stop.
```

bam241:
Hello @rherrero-pf thanks for testing my working
I am wondering if the problem doesn't come from `-DCMAKE_INSTALL_PREFIX=/DAGMC/ `
are you sure you have the right to write in the `/DAGMC` folder ? 
folder at `/` are usually protected.

the error you get is: `file problem creating directory: /DAGMC/moab`

shimwell:
Thanks @bam241 

confirming this slightly adapted version works for me
```
sudo apt-get install libhdf5-serial-dev
sudo apt install libeigen3-dev
rm -rf ~/code_development/DAGMC
mkdir -p ~/code_development/DAGMC
cd ~/code_development/DAGMC
git clone --single-branch --branch develop --depth 1 https://github.com/svalinn/DAGMC.git
mkdir build
cd build
cmake ../DAGMC -DBUILD_TALLY=ON \
              -DBUILD_STATIC_EXE=OFF \
              -DBUILD_STATIC_LIBS=OFF \
              -DCMAKE_INSTALL_PREFIX=$HOME/code_development/DAGMC/ \
              -DBUILD_UWUW=OFF \
              -DPULL_INSTALL_MOAB=5.5.1 \
              -DHDF5_ROOT=/usr/lib/x86_64-linux-gnu/hdf5/serial \
              -DEIGEN3_INCLUDE_DIR=/usr/include/eigen3
# in my case I need to help it find Eigen3 with the line above, hopefully not everyone has to do that?
make -j
make -j install
```

then when installing openmc one needs to pass the DAGMC path to openmc like this
```
cd ~/code_development
git clone --single-branch --branch develop --depth 1 https://github.com/openmc-dev/openmc.git
cd openmc
mkdir build
cd build
cmake -DOPENMC_USE_DAGMC=ON -DDAGMC_ROOT=$HOME/code_development/DAGMC -DHDF5_PREFER_PARALLEL=OFF ..
make -j
make -j install
cd ..
python -m pip install .
```

*updated now that the PR is merged to point to svalinn repo and also includes full openmc instructions

CMakeLists.txt
```diff
@@ -44,7 +44,22 @@ if(BUILD_MCNP5 OR BUILD_MCNP6)
...
+  set(MOAB_VERSION ${PULL_INSTALL_MOAB})  
+
+  # Ensure MOAB_VERSION
+  if(${MOAB_VERSION})
```
gonuke: It's not clear to me what conditions re true for this?  It seems like it would evaluate to true if I *DID* set the version? but maybe I don't know how this CMake evaluates `if(5.5.3)`, for example?

cmake/FindMOAB.cmake
```diff
@@ -1,3 +1,5 @@
+# This leverage HDF5_macro.cmake in order to find HDF5 libraries.
```
gonuke:
```suggestion
# This leverages HDF5_macro.cmake in order to find HDF5 libraries.
```

doc/CHANGELOG.rst
```diff
@@ -10,6 +10,9 @@ Next version
 **Changed:**
   * Update formating of few files to comply with clang-format (#937)
 
+**Added:**
+  * Allow download from cmake and compilation at build time of MOAB (#969)
```
gonuke:
```suggestion
  * Allow download & build of MOAB from cmake at build time (#969)
```

gonuke (changes requested):
looks pretty good - a few final clarifications

CMakeLists.txt
bam241:
```
if(<variable>)
True if given a variable that is defined to a value that is not a false constant. False otherwise, including if the variable is undefined. Note that macro arguments are not variables.
``` from https://cmake.org/cmake/help/latest/command/if.html

Comment (gonuke) [reply CMakeLists.txt]: Is a version string evaluated as a "False" constant then? I would have assumed that a version string would evaluate as `True` and cause this message and default to happen??

Comment (bam241) [reply CMakeLists.txt]: you are right I miss understood this. I'll fix it!

Comment (bam241) [reply CMakeLists.txt]: @gonuke this should take care of it !

File: CMakeLists.txt [hunk 26]
```diff
@@ -44,7 +44,22 @@ if(BUILD_MCNP5 OR BUILD_MCNP6)
...
-find_package(MOAB REQUIRED)
+# Dealing with MOAB
+if (PULL_INSTALL_MOAB)
+  set(MOAB_VERSION ${PULL_INSTALL_MOAB})  
+
+  # Ensure MOAB_VERSION
+  if(NOT ${MOAB_VERSION} MATCHES "^[0-9]+\\.[0-9]+\\.[0-9]")
+    MESSAGE("MOAB_VERSION: ${MOAB_VERSION} not recognized. Defaulting to 5.5.1")
+    set(MOAB_VERSION "5.5.1")
+  endif()
```
bam241: This should enforce MOAB_VERSION to be shaped as X.Y.Z with X, Y, Z as numbers.

shimwell: Does this work even if we have a number that takes 2 chars, e.g.  ```5.5.10```

bam241: maybe not I'll try and fix it if not

CMakeLists.txt
```diff
@@ -44,7 +44,22 @@ if(BUILD_MCNP5 OR BUILD_MCNP6)
   enable_language(Fortran)
 endif()
 
-find_package(MOAB REQUIRED)
+# Dealing with MOAB
+if (PULL_INSTALL_MOAB)
+  set(MOAB_VERSION ${PULL_INSTALL_MOAB})  
+
+  # Ensure MOAB_VERSION
+  if(NOT ${MOAB_VERSION} MATCHES "^[0-9]+\\.[0-9]+\\.[0-9]")
```
gonuke: This should address @shimwell's latest question
```suggestion
  if(NOT ${MOAB_VERSION} MATCHES "^[0-9]+\\.[0-9]+\\.[0-9]+")
```

gonuke (changes requested):
Suggested fix for number regex

CMakeLists.txt
bam241: It is now broken I don't understand why...
I'll add this suggestion to the branch as soon I understand why it is not working anymore

cmake/DAGMC_macros.cmake
```diff
@@ -253,6 +255,11 @@ macro (dagmc_install_library lib_name)
             EXPORT DAGMCTargets
             LIBRARY DESTINATION ${INSTALL_LIB_DIR}
             PUBLIC_HEADER DESTINATION ${INSTALL_INCLUDE_DIR})
+    # Required to ensure that MOAB is built before DAGMC and to properly link against MOAB
+    if(PULL_INSTALL_MOAB)
+    target_link_libraries(${lib_name}-shared INTERFACE ${MOAB_LIBRARY_DIRS}/libMOAB${CMAKE_SHARED_LIBRARY_SUFFIX})
+    add_dependencies(${lib_name}-shared MOAB)
+     endif()
```
gonuke: Can you fix this indentation, please?

gonuke (changes requested):
Sorry, but you seem to have messed up some indentation... just want that fixed, please.

gonuke (approved):
LGTM - I'd be happy to have one other person take a last look... @pshriwise ?

pshriwise (approved):

pshriwise:
Gave this another look today. Thanks @bam241!

gonuke:
This caused a test to fail on merge.  We have a test that checks to see if we are compatible with the `develop` branch of HDF5.  I am not sure what changed to make this fail - whether it was the HDF5 code changing or something in this build?

bam241:
I don't see the failing test.
could you provide a link to it ?
I'll keep looking as well

bam241:
found it

gonuke:
It doesn't really matter now, I don't think, since tests pass with the specific version of HDF5, but it may be a sign for the future?

bam241:
I think I am the one who broke it:

https://github.com/bam241/DAGMC/pull/4

bam241:
I revert the change I did on the findMOAB.cmake (responsible to deal with HDF5) and it works

Issue #805, opened 2022-04-20
False positive for assignment of vacuum


pshriwise:
@shimwell recently sent me an input with material assignments like "lower_vaccum_vessel", which DAGMC interpreted as a vaccuum assignment. This is caused by the following lines:

https://github.com/svalinn/DAGMC/blob/50caabf16b12179f92e71d4dcfb5bf6090b30cb6/src/dagmc/dagmcmetadata.cpp#L209-L212

Prior to https://github.com/svalinn/DAGMC/commit/cc505bb4e4c70c343542de7b0d2eaa211e031d53, this would not have been a problem since only the presence of "Vacuum" (as opposed to "vacuum") in the material assignment would have triggered this behavior. This is still what I would call a bug, the corner case was just rarer.

I think we either need to document that "V/vacuum" should not be used in material names or only assign volumes a vacuum condition if the lowercase version of the name matches "mat:vacuum" exactly instead of searching for the presence of "vacuum", which is what happens in the lines linked above.

The word vacuum starts to look very strange after typing it so many times at this time of night.

shimwell:
Oh wow, I thought we squashed this bug a while back but I see how we didn't quite catch all the use cases. Looks like this is the difference between the current DAGMC used in OpenMC implementation and the version Shift implementation you are noticing @jbae11

I remembered not to call tags mat:vacuum. But I didn't recall the fix searched for the string vacuum in the entire material tag. Good to know, I can rename my tags.

Plus one vote for searching for a string that is equal to =="mat: vacuum" instead of finding vacuum in the string.

Material tags that would become a vacuum: 😀

mat:vacuum_vessel
mat:lid_of_vacuum_vessel
mat:not_a_vacuum
mat:partial_vacuum
mat:dyson_vacuum_cleaner
mat:area_needs_vacuuming

pshriwise:
I tend to agree, @shimwell  -- I vote we use an exact match for those assignments. 

@makeclean @gonuke any thoughts/feelings here?

makeclean:
Seems like a regression that was fixed long ago, is the bug in dagmc proper or the alternative arrangements in OpenMC?

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

pshriwise:
[...]

PR #971, opened 2024-12-11
Fix vaccum name in materials

bam241:
This aims to follow up on #805.

Right now I just pulled @pshriwise branch and rebased it. 

## Description
This fixes the definition of the Graveyard and the Vacuum, to respectively exactly `mat:Graveyard` or `mat:graveyard` and `mat:Vacuum` or `mat:vacuum"`
## Motivation and Context
the previous implementation would lead to detection of a vacuum where there was not, if the name of the material contained "vacuum" or "graveyard"

## Changelog file
- [x] done ?

_This works is sponsored by [Proxima Fusion](https://www.proximafusion.com/)_

### DIFF
--- a/src/dagmc/dagmcmetadata.cpp
+++ b/src/dagmc/dagmcmetadata.cpp
@@ -210,10 +210,8 @@ void dagmcMetaData::parse_material_data() {
     // set the material value
     volume_material_property_data_eh[eh] = grp_name;
     logger.message("Group name -- " + grp_name);
-    bool is_graveyard =
-      to_lower(grp_name) == to_lower(graveyard_mat_str());
-    bool is_vacuum =
-        to_lower(grp_name) == to_lower(vacuum_mat_str());
+    bool is_graveyard = to_lower(grp_name) == to_lower(graveyard_mat_str());
+    bool is_vacuum = to_lower(grp_name) == to_lower(vacuum_mat_str());
 
     // not graveyard or vacuum or implicit compliment
     if (!is_graveyard && !is_vacuum && !DAG->is_implicit_complement(eh)) {
--- a/src/dagmc/dagmcmetadata.cpp
+++ b/src/dagmc/dagmcmetadata.cpp
@@ -211,9 +211,9 @@ void dagmcMetaData::parse_material_data() {
     volume_material_property_data_eh[eh] = grp_name;
     std::cout << "Group name -- " << grp_name << std::endl;
     bool is_graveyard =
-      to_lower(grp_name) == "mat:" + to_lower(graveyard_str);
+      to_lower(grp_name) == to_lower(graveyard_mat_str);
     bool is_vacuum =
-        to_lower(grp_name) == "mat:" + to_lower(vacuum_str);
+        to_lower(grp_name) == to_lower(vacuum_mat_str);
 
     // not graveyard or vacuum or implicit compliment
     if (!is_graveyard && !is_vacuum && !DAG->is_implicit_complement(eh)) {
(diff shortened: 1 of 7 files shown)


bam241:
@gonuke in the issue you mention a solution that would be backward compatible with most model out there.

Do you have an idea of the backward compatibility we main encounter ? (so I can implement a fix :) )

gonuke:
> @gonuke in the issue you mention a solution that would be backward compatible with most model out there.
> 
> 
> 
> Do you have an idea of the backward compatibility we main encounter ? (so I can implement a fix :) )

Not really... you all are the users 😀 

Maybe it's not reasonable, but I just wanted to make sure it was considered.

src/dagmc/dagmcmetadata.cpp
```diff
@@ -209,32 +209,30 @@ void dagmcMetaData::parse_material_data() {
...
     // set the material value
     volume_material_property_data_eh[eh] = grp_name;
-
-    bool is_graveyard =
-        to_lower(grp_name).find(to_lower(graveyard_str)) != std::string::npos;
-    bool is_vacuum =
-        to_lower(grp_name).find(to_lower(vacuum_str)) != std::string::npos;
+    logger.message("Group name -- " + grp_name);
+    bool is_graveyard = to_lower(grp_name) == to_lower(graveyard_mat_str());
+    bool is_vacuum = to_lower(grp_name) == to_lower(vacuum_mat_str());
```
gonuke: This may be good for long term readability
```suggestion
    bool is_graveyard = (to_lower(grp_name) == to_lower(graveyard_mat_str()));
    bool is_vacuum = (to_lower(grp_name) == to_lower(vacuum_mat_str()));
```

src/dagmc/dagmcmetadata.hpp
```diff
@@ -351,11 +360,13 @@ class dagmcMetaData {
...
-  const std::string reflecting_str{"Reflecting"};
-  const std::string white_str{"White"};
-  const std::string periodic_str{"Periodic"};
+  const std::string graveyard_str_{"Graveyard"};
+  const std::string vacuum_str_{"Vacuum"};
+  const std::string vacuum_mat_str_{"mat:Vacuum"};
+  const std::string graveyard_mat_str_{"mat:Graveyard"};
+  const std::string reflecting_str_{"Reflecting"};
+  const std::string white_str_{"White"};
+  const std::string periodic_str_{"Periodic"};
```
gonuke: We frequently - but not always - convert these to lower case.  Is there a reason to not just define them as lower case?

gonuke (changes requested):
Thanks for this cleanup @bam241 .  I have a couple of minor questions, but otherwise it looks good.

If you want to consider this for the next release, you can update the base branch as `v3.2.4-rc1`.  (It may require a rebase on that branch and reconciling conflicts in the ChangeLog.)

src/dagmc/dagmcmetadata.hpp
bam241: I think they are only converted to lower case when comparing with a group name read from a h5 geometry.
My guess is the default if not lower case, but allowing some flexibility for the users.

I don't have preferences there, it is really up to you and Pat :)

@pshriwise as you are the initiator of the branch do you have a take on this ?

bam241:
@pshriwise @gonuke would it be interesting to add a setter for the `Graveyard` and `Vacuum` string.

Allowing users to change them on the DAGMC call instead of having to change the whole geometry definition?

That way we could ensure a little of backward compatibility.

we could keep the default as they are now, but allow some flexibility...

bam241:
> @pshriwise @gonuke would it be interesting to add a setter for the `Graveyard` and `Vacuum` string.
> 
> Allowing users to change them on the DAGMC call instead of having to change the whole geometry definition?
> 
> That way we could ensure a little of backward compatibility.
> 
> we could keep the default as they are now, but allow some flexibility...

maybe @makeclean has also an opinion on my last question :)

bam241:
@gonuke I rebased against `svalinn:v3.2.4-rc1` and changed the target branch.

about your comment, I guess I am not the one with the most expertise about this.
I keep as they were.

bam241:
(btw unit test didn't run after my rebase -- Do they only run on PR vs develop ?.)

gonuke (approved):

PR #975, opened 2025-02-13
bumping hdf5 version and fix naming convention for hdf5 version in th…

bam241:
Bumping the hdf5 version we are testing against to 1.14.6.

Also as discussed with @gonuke fixing some inconsistency between the naming convention of the docker container we build for CI and what is actually used.

We made the choice to form the name of the branch of HDF5 we are pulling from hdf5 repo.

I also search for all hdf5_version mention and try to apply the same logic.

I didn't add the changeling yet so we can take the time to review this.

### DIFF
--- a/.github/workflows/linux_build_test.yml
+++ b/.github/workflows/linux_build_test.yml
@@ -44,7 +44,7 @@ jobs:
           clang,
         ]
         hdf5_version : [
-          1_14_6,
+          1.14.3,
         ]
         moab_version : [
           5.4.1,
@@ -64,7 +64,7 @@ jobs:
         include:
           - ubuntu_version: 22.04
             compiler: gcc
-            hdf5_version: 1_14_6
+            hdf5_version: 1.14.3
             moab_version: 5.4.1
             pull_install_moab: on
             geant4_version: off
--- a/.github/workflows/linux_build_test.yml
+++ b/.github/workflows/linux_build_test.yml
@@ -44,7 +44,7 @@ jobs:
           clang,
         ]
         hdf5_version : [
-          1.14.3,
+          1_14_6,
         ]
         moab_version : [
           5.4.1,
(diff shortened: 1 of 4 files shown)


bam241:
Should I have this test on my own fork ?

.github/workflows/linux_build_test.yml
```diff
@@ -64,7 +64,7 @@ jobs:
         include:
           - ubuntu_version: 22.04
             compiler: gcc
-            hdf5_version: 1.14.3
+            hdf5_version: 1_14_6
```
gonuke: I'd rather use the version with dots than underscores, if possible

gonuke (changes requested):
I made a comment about doing this in two steps... but I'm also confused about why we don't test the docker_publish changes in a PR??

bam241:
> I made a comment about doing this in two steps... but I'm also confused about why we don't test the docker_publish changes in a PR??

Sorry I didn't catch it.

you mean first changing the docker_build + docker, then the tests ?

gonuke:
This may need to be done in 2 steps:
1. update the docker_publish workflow (and Dockerfile) from a branch on the upstream repo
2. update the build/test workflow after that is merged

> > I made a comment about doing this in two steps... but I'm also confused about why we don't test the docker_publish changes in a PR??
> 
> Sorry I didn't catch it.
> 
> you mean first changing the docker_build + docker, then the tests ?

(I thought I had commented, but maybe never pressed the button)

I think we need two PR's:
1. update docker_publish and Dockerfile: this may need to be a branch on the upstream repo - I'm not sure.
2. after this is merged: update the build/test

bam241:
closing reopening with upstream branch

bam241:
this has move to #976 building new docker container need to be run from upstream

PR #976, opened 2025-02-13
Docker hdf5 1.14.6

bam241:
same as #975 but from upstream

PR #978, opened 2025-02-14
propagating hdf5 version update to CI

bam241:
This follows #976.
now updating the version of HDF5 used in the CI.

### DIFF
--- a/.github/workflows/linux_build_test.yml
+++ b/.github/workflows/linux_build_test.yml
@@ -54,8 +54,8 @@ jobs:
           off,
         ]
         geant4_version : [
-          10.7.4,
-          11.1.2
+          v10.7.4,
+          v11.1.2
         ]
         double_down_version : [
           off,
--- a/.github/workflows/linux_build_test.yml
+++ b/.github/workflows/linux_build_test.yml
@@ -97,10 +97,10 @@ jobs:
           mkdir -p build
           cd build
           cmake ../ \
-              -DMOAB_DIR=${moab_install_dir} \
+              -DMOAB_DIR=${MOAB_INSTALL_DIR} \
               -DBUILD_GEANT4=$([ "${{ matrix.geant4_version }}" != "off" ] && echo "ON" || echo "OFF") \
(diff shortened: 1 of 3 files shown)


bam241:
I don't understand how this was working before

bam241:
btw it appears that we build more Docker container than we use in the `build_test.yml` (build matrix are not the same...)

bam241:
I think it would be nice to have the matrix existing in an other file, and pull it, but I didn't find a easy/clean way to do it :)

gonuke:
> I think it would be nice to have the matrix existing in an other file, and pull it, but I didn't find a easy/clean way to do it :)

I agree that this would be great!

.github/workflows/linux_build_test.yml
```diff
@@ -83,7 +83,7 @@ jobs:
                 matrix.geant4_version
               }}-double_down_${{
                 matrix.double_down_version
-              }}/moab:latest
+              }}:latest
```
gonuke: I'm a little concerned about this, because it means building DAGMC in a container that already has DAGMC installed....  The `moab` descriptor should be from a version of the build that stopped at MOAB

gonuke (changes requested):
I want to make sure we use the right container for testing so that we don't get false positive results.

bam241:
there is no `/moab` container:

image

gonuke:
> there is no `/moab` container:
> 

hmmm... Maybe that's another inconsistency that has creeped in....

gonuke:
Let's remove the `moab` again and review the tests carefully to see that they are using the right build for testing.

I think we should maybe go through and delete all the old containers from the GHCR, carefully document the CI process, and start fresh.

bam241:
from one I see in the Dockerfile we need to target `external_deps`

bam241:
@gonuke `external_deps` descriptor seems to be ok, and it is the stage before `DAGMC` installation in the `Dockerfile`.

gonuke (approved):
Thanks for the follow up here @bam241

PR #981, opened 2025-04-23
Fix G4 tracking

makeclean:
## Description
This PR updates the DagSolid implementation so that it (finally) works as intended for geometries that are imprinted and merged. Through a number of factors, the original author only ever intended for it to work with single isolated dagmc volumes, not entire assemblies as we typically do. 

## Motivation and Context
This makes it so that DAGMC geometies can be properly, and correctly be transported on using either the MOAB or DoubleDown backend. 

## Changes
Bug fixes

## Behavior
Fixes previously broken implementation. 

## Changelog file
Changelog will be updated when review completed.


doc/CHANGELOG.rst
```diff
@@ -9,6 +9,7 @@ Next version
 
 **Changed:**
 
+  * Fixed Geant4 tracking when using non-trivial geometries
```
gonuke:
```suggestion
  * Fixed Geant4 tracking when using non-trivial geometries (#981)
```

src/geant4/DagSolid.cc
```diff
@@ -237,60 +241,90 @@ EInside DagSolid::Inside(const G4ThreeVector& p) const {
 G4ThreeVector DagSolid::SurfaceNormal(const G4ThreeVector& p) const {
   G4double ang[3] = {0, 0, 1};
   G4double position[3] = {p.x() / cm, p.y() / cm, p.z() / cm};  // convert to cm
+  G4double distance;
+  EntityHandle surface;
+  fdagmc->closest_to_location(fvolEntity, position, distance, &surface);
 
-  fdagmc->get_angle(My_sulf_hit, position, ang);
```
gonuke, line 241: I don't think we use `My_sulf_hit` anymore? Can we drop it?

src/geant4/DagSolid.cc
```diff
@@ -237,60 +241,90 @@ EInside DagSolid::Inside(const G4ThreeVector& p) const {
...
   G4double minDist = kInfinity;
   G4double position[3] = {p.x() / cm, p.y() / cm, p.z() / cm};  // convert to cm
-  G4double dir[3] = {v.x(), v.y(), v.z()};
-  EntityHandle next_surf;
+  G4ThreeVector vec = v.unit();
+  G4double dir[3] = {vec.x(), vec.y(), vec.z()};
+  EntityHandle next_surf = 0;
   G4double distance;
 
-  DagMC::RayHistory history;
```
gonuke, line 260: Do we really not need a `RayHistory` anymore?  Where do you get robustness for some of the pathologies for which it exists?

src/geant4/DagSolid.cc
```diff
@@ -190,41 +198,37 @@ DagSolid::~DagSolid() {}
...
+  // are we inside
+  fdagmc->point_in_volume(fvolEntity, point, inside);
+  // also need to know how far from the surface
+  fdagmc->closest_to_location(fvolEntity, point, minDist);
 
   // if on surface
-  if (minDist <= 0.5 * kCarTolerance) {
-    return kSurface;
+  EInside result;
+  if (minDist * cm <= kCarToleranceHalf) {
```
gonuke, line 125: Was it just an oversight that we didn't do this unit version (I assume that's what it is) before?

src/geant4/DagSolid.cc
```diff
@@ -237,60 +241,90 @@ EInside DagSolid::Inside(const G4ThreeVector& p) const {
...
+//
+/////////////////////////////////////////////////////////////////////////
 G4double DagSolid::DistanceToIn(const G4ThreeVector& p) const {
   G4double minDist = kInfinity;
   G4double point[3] = {p.x() / cm, p.y() / cm,
                        p.z() / cm};  // convert position to cm
 
   fdagmc->closest_to_location(fvolEntity, point, minDist);
-  minDist *= cm;  // convert back to mm
-  if (minDist <= kCarTolerance * 0.5)
```
gonuke, line 290: You do this tolerance check elsewhere, why no longer here?

src/geant4/DagSolid.cc
```diff
@@ -311,61 +345,92 @@ G4double DagSolid::DistanceToIn(const G4ThreeVector& p) const {
 //       exiting surface.
 // If calcNorm is false, then validNorm and n are unused.
 
+// distqance to out - when inside - if the distaance is within halfkcartolerance
+// then return 0, if aasked to calculate the normal then
```
gonuke, line 250:
```suggestion
// distance to out - when inside - if the distance is within halfkcartolerance
// then return 0, if asked to calculate the normal then
```

src/geant4/DagSolid.cc
```diff
@@ -311,61 +345,92 @@ G4double DagSolid::DistanceToIn(const G4ThreeVector& p) const {
...
 }
 
 ///////////////////////////////////////////////////////////////////////////////
 //
 // G4double DistanceToOut(const G4ThreeVector& p)
 //
 // Calculate distance to nearest surface of shape from an inside point.
-
+// The distance can be an underestimate.
 G4double DagSolid::DistanceToOut(const G4ThreeVector& p) const {
```
gonuke, line 322: Other than the name of the function in the debug string, this is identical to `DistanceToIn()` - should we have one function that does this and wrap it in `DistanceToIn()` and `DistanceToOut()` to ensure consistency while preserving the API and semantics?

src/geant4/DagSolid.cc
```diff
@@ -484,18 +549,31 @@ G4double DagSolid::GetMinZExtent() const { return zMinExtent / cm; }
...
 
-///////////////////////////////////////////////////////////////////////////////
-//
+/*
+ * Return the surface area of the volume -
+ * note DAGMC is always in the units of cm
+ */
 G4double DagSolid::GetSurfaceArea() {
   G4double result;
-  fdagmc->measure_area(fvolEntity, result);
```
gonuke, line 499: Was this not working?  Should we fix it so that it does?

src/geant4/app/src/ExN01RunAction.cc
```diff
@@ -24,8 +24,8 @@ void ExN01RunAction::BeginOfRunAction(const G4Run* /*run*/) {
   G4AnalysisManager* analysisManager = G4AnalysisManager::Instance();
 
   // Open an output file  - DagGeant.root
-  G4String fileName = "DagGeant";
-  analysisManager->OpenFile(fileName);
+  // G4String fileName = "DagGeant";
+  // analysisManager->OpenFile(fileName);
```
gonuke, line 28: Should we just delete these?

src/geant4/tests/dagsolid_unit_tests.cpp
```diff
@@ -118,6 +118,83 @@ TEST_F(DagSolidTest, point_outside_volume) {
...
+
+  // set the vector to be the lhs
+  position.setX(0.);
+  position.setY(-50.);
+
+  normal = vol_1->SurfaceNormal(position);
+
+  EXPECT_EQ(normal[0], 0.0);
+  EXPECT_EQ(normal[1], -1.0);
+  EXPECT_EQ(normal[2], 0.0);
```
gonuke: This looks like a repeat

src/geant4/tests/dagsolid_unit_tests.cpp
```diff
@@ -118,6 +118,83 @@ TEST_F(DagSolidTest, point_outside_volume) {
...
+  // set the vector to be the lhs
+  position.setX(-50.);
+
+  normal = vol_1->SurfaceNormal(position);
+
+  EXPECT_EQ(normal[0], -1.0);
+  EXPECT_EQ(normal[1], 0.0);
+  EXPECT_EQ(normal[2], 0.0);
+
+  // set the vector to be the lhs
```
gonuke: Update comment?
```suggestion
  // set the vector to be in front
```
(Or maybe on top??)

src/geant4/tests/dagsolid_unit_tests.cpp
```diff
@@ -118,6 +118,83 @@ TEST_F(DagSolidTest, point_outside_volume) {
...
+  position.setX(0.);
+  position.setY(-50.);
+
+  normal = vol_1->SurfaceNormal(position);
+
+  EXPECT_EQ(normal[0], 0.0);
+  EXPECT_EQ(normal[1], -1.0);
+  EXPECT_EQ(normal[2], 0.0);
+
+  // set the vector to be the lhs
```
gonuke: Update this comment consistent with choice for previous comment

src/geant4/tests/dagsolid_unit_tests.cpp
```diff
@@ -118,6 +118,83 @@ TEST_F(DagSolidTest, point_outside_volume) {
...
+  position.setY(0.);
+  position.setZ(50.);
+
+  normal = vol_1->SurfaceNormal(position);
+
+  EXPECT_EQ(normal[0], 0.0);
+  EXPECT_EQ(normal[1], 0.0);
+  EXPECT_EQ(normal[2], 1.0);
+
+  // set the vector to be the lhs
```
gonuke: Update comment consistent with previous choices

src/geant4/tests/dagsolid_unit_tests.cpp
```diff
@@ -118,6 +118,83 @@ TEST_F(DagSolidTest, point_outside_volume) {
...
+  position.setX(0.);
+  position.setY(50.);
+
+  normal = vol_1->SurfaceNormal(position);
+
+  EXPECT_EQ(normal[0], 0.0);
+  EXPECT_EQ(normal[1], 1.0);
+  EXPECT_EQ(normal[2], 0.0);
+
+  // set the vector to be the lhs
```
gonuke: Update comment consistent with previous choices

gonuke (changes requested):
Thanks @makeclean - here are a few minor comments

src/geant4/DagSolid.cc
makeclean: Yep, removed I done a big tidy up as well.

src/geant4/DagSolid.cc
makeclean: Yep don't need it, Geant4 has a tolerance kCarTolerance which is set to, I think 5e-8 mm so when you get to a surface, once you are within that distance you are considered on it, and when finding the next volume you are removed from its candidacy of being a hit, which does the same sort of thing but not by exclusion of hits of particular triangles.

src/geant4/DagSolid.cc
makeclean: yeah an oversight, I think ultimately the right way to do this is to have the DagSolid own its own MOAB instance, copy the vertices and triangles used by the solid into a new DAGMC instance and scale them to be in native units of Geant4, then we can do away with the cm conversions all together.

src/geant4/DagSolid.cc
pshriwise: How does that work if the last surface crossed is also the next surface hit in the new volume? For example, a sinusoid surface or the like?

src/geant4/DagSolid.cc
makeclean: DistanceToIn(p) and DistanceToOut(p) are used as safety(s) and other implementations in G4 do not use the tolerance check for other solids, so I was following their style.

src/geant4/DagSolid.cc
makeclean: It would approach the crossing surface, be on the crossing surface, leave volume 1 crossing the surface, G4 would fire against all volumes on that level of the hierarchy ignoring the possiblity of going back into the current volume, it would hit the surface at 0 or close to 0 distance given the surface is the same with the opposite normal, and enter into that volume

pshriwise: Ah okay so it isn't masking that surface, just some distance within the previous hit. Is that right?

makeclean: It masks the entire current volume in Geant4 terms you are not allowed to hit it again, however entering the next volume with a shared surface would be allowed since you would be entering and would have the sign flipped

src/geant4/tests/dagsolid_unit_tests.cpp
makeclean: Done

src/geant4/tests/dagsolid_unit_tests.cpp
makeclean: Done

src/geant4/tests/dagsolid_unit_tests.cpp
makeclean: Done

src/geant4/tests/dagsolid_unit_tests.cpp
makeclean: Done

src/geant4/tests/dagsolid_unit_tests.cpp
makeclean: Removed

doc/CHANGELOG.rst
makeclean: Done

makeclean:
Now, what I can't figure out is why the G4Polyhedron.hh header isn't being found in CI, when it is present in the version of Geant4 in CI, and is present on my machine, where it builds fine.

image

makeclean:
Aha, Geant 10.7 is too old to support the most recent changes, I'll put in some support for older versions

makeclean:
Green ticks - :partying_face: - just waiting on Mac

gonuke:
> Aha, Geant 10.7 is too old to support the most recent changes, I'll put in some support for older versions

Should we be moving on from older versions? I don't have my finger on that pulse...

makeclean:
> > Aha, Geant 10.7 is too old to support the most recent changes, I'll put in some support for older versions
> 
> Should we be moving on from older versions? I don't have my finger on that pulse...

Yeah I would stick to the newest version and maybe upto 2 minor versions previous.

